PR #44, opened 2026-04-17
ticket 0001: L1 file-lock concurrency guard for /verify

MinhHaDuong:
## Summary

- Rescopes ticket 0001 from three coordination layers to L1 only (same-machine file lock). L2 (GitHub label) and L3 (monotonic counter) deferred — speculative for a single-user harness; multi-machine sync is STATE.md backlog; /verify-gate already caps rounds by comment scan.
- Adds L1 file lock to `skills/verify/SKILL.md` mirroring the existing `.git/ticket-wip/` pattern. Path: `.git/verify-wip/<pr>.wip`. Single-line content with ISO timestamp, session id, worktree path, round. 30-min staleness.
- Lock acquired at Phase 1 step 1a (before `gh pr checkout`). Released on every terminal path: APPROVED, ESCALATE, round-2 REROLL upgraded to ESCALATE, telemetry-threshold escalation, `--force-approve`. Round 2 re-entry holds the lock and rewrites the round marker.

## Test plan

- [ ] Terminal A: `/verify <pr>` — creates `.git/verify-wip/<pr>.wip`, proceeds.
- [ ] Terminal B (same machine, while A runs): `/verify <pr>` — aborts, prints lock path + content, no PR comment.
- [ ] After A completes (APPROVED or ESCALATE): lock file deleted.
- [ ] Manually age a lock file by 31 min: next `/verify` reclaims it and records an informational line on the verdict comment.
- [ ] `--force-approve` path: acquires, posts force-approved comment, releases.

Closes ticket 0001.

### DIFF
--- a/skills/verify/SKILL.md
+++ b/skills/verify/SKILL.md
@@ -26,18 +26,53 @@ Merge is always the human's or the orchestrator's call.
 - The fix loop between rounds makes commits on the PR branch; no changes to other branches.
 - `--force-approve` is supported for explicit human override; it is logged loudly in the
   PR comments and the skill transcript.
+- Only one `/verify` may run per PR per machine at a time. Enforced by an L1 file lock
+  (see `## Claim file (L1)`). Cross-machine coordination is out of scope.
+
+## Claim file (L1)
+
+A same-machine concurrency guard that mirrors the `.git/ticket-wip/` pattern used for
+ticket claims. Prevents the common collision: two chats / two terminals running
+`/verify <same-pr>` simultaneously on the same workstation.
+
+- **Path:** `.git/verify-wip/<pr>.wip` (inside `.git/`, never committed, shared across
+  worktrees via `git-common-dir`).
+- **Content:** one line — `{ISO-8601-timestamp} {session_id} {worktree-path} round={n}`.
+- **Acquire (Phase 1, step 1a):**
+  1. If the file does not exist → create it with the current timestamp, session id,
+     worktree path, and `round=1`. Proceed.
+  2. If the file exists and its timestamp is < 30 minutes old → abort. Print the
+     lock file path and its single-line content so the user can find the owning
+     session. Do not post a PR comment (the other session is still running and
+     will post its own).
+  3. If the file exists and its timestamp is ≥ 30 minutes old → treat as stale.
+     Delete, recreate with current metadata, and record `L1: stale lock reclaimed
+     (previous: <content>)` as an informational line in the final verdict comment.
+- **Round 2 re-entry:** the lock is held across rounds. Rewrite the file with
+  `round=2` when entering round 2 (same session, updated round marker). Do not
+  re-acquire.
+- **Release:** delete the file on every terminal path — APPROVED, ESCALATE (including
+  round-2 REROLL upgraded to ESCALATE), telemetry-threshold escalation, and
+  `--force-approve`. Release must happen **after** the verdict comment is posted,
+  so an observer sees the lock tied to the posted outcome.
+- **Cross-machine:** not covered. If multi-machine `/verify` collisions become real,
+  add a GitHub-label layer (L2) and a monotonic round counter (L3) as a follow-up.
+  See ticket 0001 body for the deferred design.
 
 ## Phases
 
 ### 1. Setup
 
+- **1a. Acquire L1 lock** (see `## Claim file (L1)`). If held by another live session,
+  abort before doing any work. If stale, reclaim and record for the verdict comment.
 - `gh pr checkout $ARGUMENTS` into an isolated worktree. Abort if the PR is not mergeable
-  or if there are open merge conflicts.
+  or if there are open merge conflicts. On abort, release L1.
 - Collect:
   - The ticket file referenced in the PR title or body (`tickets/*.erg`).
   - PR body, full diff, all existing review comments, all inline comments, all commit
     messages on the branch.
-- If any of these cannot be located, ESCALATE with a clear message. Do not proceed.
+- If any of these cannot be located, ESCALATE with a clear message and release L1.
+  Do not proceed.
 
 ### 2–4. Read-only review fan-out (parallel)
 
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
## /review-pr — multi-perspective review

**Verdict: comment** (no blockers; approve pending test-plan walkthrough)

PR is scope-appropriate: doc-only change to `skills/verify/SKILL.md` (46/10) and the already-closed ticket 0001 (24/2). Rescope from three layers to L1-only is clearly documented in the ticket body; deferred L2/L3 design preserved for history. Exit criteria for the L1 scope are covered by the contract.

### Correctness

- **verifiable:** `skills/verify/SKILL.md:34` — L1 contract mirrors `.git/ticket-wip/` pattern at `tickets/FORMAT.md:141-148`. Acquire/release/stale semantics align with the existing ticket claim protocol. Path, `git-common-dir` note, and "never committed" invariant all match.
- **verifiable:** `skills/verify/SKILL.md:54-57` — Release list enumerates: APPROVED, ESCALATE (incl. round-2 REROLL upgrade), telemetry-threshold escalation, `--force-approve`. The setup-abort paths on lines 73-75 also explicitly release. Full coverage of named terminal paths in the verdict/circuit-breaker list.
- **consider:** `session_id` appears in the content spec (`skills/verify/SKILL.md:40`) but is never defined in the harness. Only hit elsewhere is `skills/end-session/SKILL.md:15` (also undefined there). Fine for a human-readable diagnostic field, but an author implementing this has no convention to follow. Consider a one-line note: "session_id is any stable token distinguishing the invocation — e.g., `$$` or a uuid".
- **consider:** Circuit breakers (`skills/verify/SKILL.md:135-140`) list ESCALATE paths for "phases 2-5 error/timeout", "fix agent timeout", "gate disagrees" that are not individually annotated with "release L1" (unlike the verdict block at 110-119 where every branch spells it out). The Release bullet's generic "ESCALATE" term does cover them, but an implementer reading the circuit-breaker list in isolation wouldn't see the release reminder there.

### Consistency

- **verifiable:** Pattern `.git/verify-wip/<pr>.wip` mirrors `.git/ticket-wip/{ID}.wip`. Path convention, sharing-via-git-common-dir note, and local-only semantics are identical.
- **consider:** Content format drifts from the ticket-wip precedent. `tickets/FORMAT.md:144` specifies `{timestamp} {actor} {worktree-path}`; this PR uses `{ISO-8601-timestamp} {session_id} {worktree-path} round={n}`. The `round={n}` suffix is justified by verify's two-round structure. The `session_id` vs `actor` substitution is unexplained — both identify who holds the lock, but harness-wide consistency would favor `actor` unless there's a reason (multiple concurrent sessions by the same actor?) to distinguish.
- **verifiable:** Ticket 0001 log verbs (`note`, `claimed`, `status`) conform to `tickets/FORMAT.md:103` closed verb set. Status transition `open` → `closed` with log entry logged. Well-formed `%erg v1`.

### Scope

- **verifiable:** Diff touches exactly 2 files. No speculative L2 (label) or L3 (counter) code. The `## Rescope` section in the ticket body explicitly marks both as deferred with rationale. Ticket's original three-layer design preserved for historical context. No scope creep.
- **nofollow:** The preserved historical sections in the ticket body (Threat model, Design: three-layer, Rejected alternatives, original Test plan, original Exit criteria) are now partially obsolete against the Rescope section. Leaving them as-is is the stated intent.

### Red team

- **consider:** Stale-reclaim is not atomic (`skills/verify/SKILL.md:46-49`: "Delete, recreate with current metadata"). Two sessions both seeing a ≥30-min-old lock could race — both delete, both recreate, both proceed. Low-probability (requires two invocations to cross the 30-min boundary within the same few ms), but the ticket's original design explicitly called for `O_EXCL`. The SKILL contract dropped that. Consider adding: "Use atomic create (e.g., `set -o noclobber` redirect or `O_EXCL`) so that concurrent reclaims resolve to a single winner."
- **consider:** Crash-then-restart within 30 min: a session that crashed after acquire but before release sees its own abandoned lock as `< 30 minutes old → abort`. Recovery requires either waiting 30 min or manual `rm`. Documenting the manual-override path (or a `--force-unlock` flag) would help; not strictly required.
- **consider:** Verdict-post failure strands the lock. Release happens "after the verdict comment is posted" (line 54-55). If `gh` API fails on the final post, release never runs. Stale reclaim handles it after 30 min, so the damage is bounded. Acceptable.

### Tests

- Automated: `make check` not present in this repo; the change is doc-only, no executable code paths to exercise.
- Manual: PR body's five-item test plan covers the contract adequately (A/B collision, release-on-complete, stale reclaim, `--force-approve`). Recommend executing items 1-3 at least once before declaring the feature trusted.

### Summary

No blockers. Two `consider:` items worth a follow-up if you want to tighten:
1. Define `session_id` (or switch to `actor` for consistency with `tickets/FORMAT.md`).
2. Specify atomic-create semantics for the acquire step (matches the ticket's original intent).

Everything else is stylistic or preserved-for-history.

MinhHaDuong:
## Closing: cost-benefit reversed on reflection

The reimagine scoped this down from three layers to L1 only on the argument that the three-layer design was speculative. That judgment was right. But on further reflection, L1 itself is speculative and worse: it buys a rare, intent-driven prevention at the cost of guaranteed operational friction.

**The asymmetry that kills it.**

The ticket-wip pattern is a good fit for ticket claims because ticket work lives for hours to days — a stale `.wip` file is low-friction to notice and clear. `/verify` is a ~5-minute operation. A 30-minute stale-lock TTL outlives the operation it protects by 6×.

Expected frequencies per month of normal use:
- Intentional double-run prevented: 0–1 (requires the user to start A, notice it's running, open a second terminal, and start B before A finishes)
- Stale lock from a crashed / Ctrl-C'd / hook-blocked run: 3–10

The thing being prevented is rarer than the breakage it introduces, and the breakage is a 30-minute wait or a manual `rm` on a path the user doesn't remember.

**What's already there that I missed.**

`/verify-gate` already scans PR comments for `round=` markers. If two `/verify` runs on the same PR ever really do race, the second gate verdict sees the first's round=1 comment and either reconciles or escalates. The doubling risk lives at the *comment* layer, not the *correctness* layer. Doubled comments are cheap to clean up.

**Decision.** Close PR, drop the skill changes, mark ticket 0001 "won't build." The original concern (confusion from parallel verifies) was real, but the proposed fix has worse expected cost than the problem. If the collision ever materialises in practice, revisit with real data.

**Lesson worth keeping:** stale-lock TTL has to fit the protected operation's lifetime. Copying a pattern across lifetime scales without adjustment is a smell. Added to the ticket body as a post-mortem note for future speculative-architecture reviews.

PR #56, opened 2026-04-24
feat(0018): add set -euo pipefail to all hook scripts + CI guard

MinhHaDuong:
## Summary

- Add `set -euo pipefail` as line 2 to all 6 hook scripts (`guard-destructive-bash.sh`, `guard-commit-on-main.sh`, `block-pr-merge-in-worktree.sh`, `on-start.sh`, `lint-on-edit.sh`, `warn-stale-rules.sh`); `shell-init.sh` is sourced and stays exempt
- Add `|| true` where commands legitimately return non-zero: `git branch --show-current`, `grep` pipeline in `persist_env`, all 4 ruff invocations (`check` + `format`, both `uv` and direct paths), and `grep -oP` in `warn-stale-rules.sh`
- Switch `$CLAUDE_PROJECT_DIR` and `$CLAUDE_ENV_FILE` expansions to `${VAR:-}` form to be safe under `set -u` when variables may be unset
- Add `pipefail-guard` CI job that checks all `scripts/*.sh` (except `shell-init.sh`) have the guard, with a self-test that verifies the guard catches a script without it

## Test plan

- [ ] CI `pipefail-guard` job passes on this branch
- [ ] `grep -L 'set -euo pipefail' scripts/*.sh | grep -v 'shell-init\.sh'` returns empty
- [ ] `bash scripts/check-project-leak.sh` exits 0 (no regressions)
- [ ] Existing CI jobs (`validate-tickets`, `skill-lint`, `leak-guard`) still pass

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -65,3 +65,28 @@ jobs:
             exit 1
           fi
           echo "OK: escape hatch suppressed leak"
+
+  pipefail-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: All hook scripts must have set -euo pipefail
+        run: |
+          missing=$(grep -L 'set -euo pipefail' scripts/*.sh | grep -v 'shell-init\.sh')
+          if [ -n "$missing" ]; then
+            echo "FAIL: missing set -euo pipefail in:"
+            echo "$missing"
+            exit 1
+          fi
+          echo "OK: all hook scripts have set -euo pipefail"
+      - name: Self-test — guard must catch a script without pipefail
+        run: |
+          tmp=$(mktemp --suffix=.sh)
+          echo '#!/bin/bash' > "$tmp"
+          echo 'echo hello' >> "$tmp"
+          missing=$(grep -L 'set -euo pipefail' "$tmp")
+          if [ -z "$missing" ]; then
+            echo "FAIL: guard did not catch missing pipefail"
+            exit 1
+          fi
+          echo "OK: guard catches missing pipefail"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -72,7 +72,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: All hook scripts must have set -euo pipefail
         run: |
-          missing=$(grep -L 'set -euo pipefail' scripts/*.sh | grep -v 'shell-init\.sh')
+          missing=$(grep -L 'set -euo pipefail' scripts/*.sh | grep -v 'shell-init\.sh' || true)
           if [ -n "$missing" ]; then
             echo "FAIL: missing set -euo pipefail in:"
             echo "$missing"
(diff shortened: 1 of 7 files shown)


MinhHaDuong (commented):
## Multi-perspective review — PR #56 (0018: set -euo pipefail)

**Risk level**: Standard + scripts. Agents run: Correctness, Consistency, Doc propagation.

**CI status**: All 4 jobs pass (validate-tickets, skill-lint, leak-guard, pipefail-guard). ✓

---

### Correctness

All `|| true` placements are correct and complete:

- `guard-commit-on-main.sh`: `git branch --show-current || true` — needed; git returns non-zero outside a repo.
- `on-start.sh` (persist_env): `grep … | sed … || true` — needed; grep exits 1 on all-comment/blank .env file.
- `lint-on-edit.sh`: both `ruff check … || true` and `ruff format … || true` — needed; ruff exits 1 on unfixable violations. Output is still captured correctly (`|| true` is inside `$(...)` so stdout is captured before the exit code is swallowed).
- `warn-stale-rules.sh`: `grep -oP … || true` — needed; grep exits 1 when no `last-reviewed:` line found.

Scripts that add `set -euo pipefail` without `|| true`:
- `guard-destructive-bash.sh`: all greps are inside `if` conditions — safe under `set -e`. ✓
- `block-pr-merge-in-worktree.sh`: grep is inside `if … && grep -q …` — safe under `set -e`. ✓

**on-start.sh guard logic** (advisor concern): `[ -n "${CLAUDE_PROJECT_DIR:-}" ] && cd "$CLAUDE_PROJECT_DIR" || exit 0` correctly handles all three cases (unset, set+cd-ok, set+cd-fail). The `${CLAUDE_PROJECT_DIR:-}` on line 42 is redundant (the guard on line 39 ensures we only reach line 42 when the variable is set) but harmless.

consider: `on-start.sh` line 48 (`git config core.hooksPath hooks`) is a bare statement now activated by `set -euo pipefail` — if `CLAUDE_PROJECT_DIR` points to a non-git directory, `git config` would fail and abort the hook. Pre-existing code, low real-world risk (harness always runs in git repos), and SessionStart hooks ignore exit codes, but worth noting. No `|| true` strictly required.

**Coverage**: All 7 non-exempt scripts have the guard. `check-project-leak.sh` had it already (added in PR #53). `shell-init.sh` is correctly exempted (sourced, not executed). CI guard accounts for the exemption.

Verdict: **approve**

---

### Consistency

The fix commit (ebb5218) correctly diagnosed and resolved the CI happy-path bug (grep pipeline exits 1 in command substitution when no files are missing, which fails under the runner's implicit `set -e`). The three CI self-test steps now mirror the established `leak-guard` pattern.

consider: The CI guard matches the exact string `set -euo pipefail`. A future script written as `set -eu -o pipefail` or `set -e -u -o pipefail` would silently pass the filesystem check but fail the CI guard — or vice versa depending on spelling. The convention is enforced by the guard's own string, so this is self-sealing as long as all scripts follow the pattern established here. No action required.

nofollow: Checked whether `block-pr-merge-in-worktree.sh`'s `gh api` example text in the heredoc should be updated for forge-agnostic wording — pre-existing content, out of scope for this ticket.

Verdict: **approve**

---

### Doc propagation

The ticket body and log accurately describe the final implementation (6 scripts, shell-init.sh exempt, 3 specific `|| true` locations). No rules/*.md files reference pipefail convention. No README or user-facing docs affected.

nofollow: Ticket exit criterion 1 says "All scripts/*.sh have set -euo pipefail" without mentioning the shell-init.sh exemption — the CI guard is the authoritative specification; the loose wording is not a doc error.

Verdict: **approve**

---

### Synthesis

No blockers. One `consider:` finding (git config bare statement in on-start.sh) is pre-existing, low risk, and out of scope. All exit criteria are satisfied:
- ✓ All `scripts/*.sh` have `set -euo pipefail` (shell-init.sh correctly exempt)
- ✓ CI job enforces it with two self-tests
- ✓ All four CI jobs green

**Overall verdict: approve — ready to merge.**

PR #74, opened 2026-04-29
fix(0060): erg sweep-skip slice-aliasing — fix + repair 8 tickets + validator rule

MinhHaDuong:
## Summary

- **Fix**: `cmdSweepSkip` corrupted ticket bodies via slice aliasing (`os.ReadFile` returns `cap > len`; `beforeBody := raw[:idx]` then `append` mutated `raw[idx:]` in place). Build the result in a fresh buffer.
- **Repair**: 8 tickets carried the corruption (0013, 0028, 0034, 0042, 0043, 0047, 0049, 0051) — bodies restored from each ticket's last clean commit.
- **Validator**: rule 11 extended to reject duplicate `--- log ---` / `--- body ---` separators (the corruption fingerprint), so any future regression is caught at commit time.

`appendToTicket` was audited (per ticket 0060 action 1) and a round-trip test confirmed it is already safe; left alone.

## Test plan

- [x] Go: `TestSweepSkipPreservesBody` (red without the fix)
- [x] Go: `TestAppendToTicketPreservesBody` (locks in current safety)
- [x] Go: `TestValidateRejectsDuplicateBodySeparator` + `…LogSeparator`
- [x] All 18 Go tests pass; all 85 Python tests pass
- [x] All 9 previously corrupt tickets now show one separator and a byte-identical pre-corruption body (verified against each ticket's `LAST_CLEAN` commit)

## Note for the operator

The system `erg` binary at `~/.local/bin/erg` is stale (Apr 13). After merge, rebuild and reinstall:

```sh
cd tickets/tools/go && go build -o ~/.local/bin/erg .
```

Closes ticket 0060 (closed in `a17ed5b`).

### DIFF
--- a/STATE.md
+++ b/STATE.md
@@ -41,7 +41,6 @@ Level 4 (Hooks) + raid + `/verify` loop + git-erg tickets + bibliography pipelin
 - 0051 — beat should try another project when current one is idle or frozen
 - 0057 — route .erg mutations through erg binary (blocked by erg binary exposing mutation commands)
 - 0058 — rewrite README with Imperial Dragon voice (remove GSD, new opener)
-- 0060 — fix erg sweep-skip slice-aliasing bug + repair 8 corrupted tickets
 - 0054 — [discussion] restore Five-Claws phase announcement at session start
 - 0055 — [discussion] milestone/epic layer above tickets
 - 0056 — [discussion] mid-session pause/resume checkpoints
--- a/tickets/0013-bib-to-zotero-skill.erg
+++ b/tickets/0013-bib-to-zotero-skill.erg
@@ -12,10 +12,12 @@ Author: user
 2026-04-29T02:05Z claude note sweep-skip: scope-too-large hash:eb7beb6f93fa expires:2026-04-30T02:05Z
 2026-04-29T09:18Z claude note sweep-skip: scope-too-large hash:f6250604695b expires:2026-04-30T09:18Z
 2026-04-29T09:25Z claude note sweep-skip: scope-too-large hash:5e3f6732fbef expires:2026-04-30T09:21Z
+2026-04-29T18:55Z claude note repaired body corruption from erg sweep-skip bug (ticket 0060)
 --- body ---
-:25Z claude note sweep-skip: scope-too-large hash:5e3f6732fbef expires:2026-04-30T09:21Z
---- body ---
-ib is staging. The canonical library is Zotero. Today the last step
+## Context
+
+The bibliography pipeline is: related-work-note → bib-merge → local refs.bib.
+Local .bib is staging. The canonical library is Zotero. Today the last step
 (refs.bib → Zotero) is manual (File → Import). This ticket automates it.
 
 The user has a Zotero PAT and Zotero v9+ (fast release cycle, API stable).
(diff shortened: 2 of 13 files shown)


MinhHaDuong (commented):
## Multi-perspective review — looks good to merge

**Verdict (advisory, since I cannot approve my own PR):** approve. The bug fix is correct, the eight ticket repairs are byte-identical to pre-corruption historic versions on `origin/main`, the validator rule is consistent across `FORMAT.md` and the Go code, and scope is tight (no unrelated changes).

### Verified claims
- `cmdSweepSkip` now builds the result in a fresh buffer (`make([]byte, 0, ...)`) — the original aliasing read of `raw[:idx]` is gone.
- `TestSweepSkipPreservesBody`, `TestAppendToTicketPreservesBody`, `TestValidateRejectsDuplicateBodySeparator`, `TestValidateRejectsDuplicateLogSeparator` all pass; 18 top-level Go tests in total (matches the PR description).
- All 8 repaired ticket bodies (0013, 0028, 0034, 0042, 0043, 0047, 0049, 0051) match a pre-corruption clean version on `origin/main` byte-for-byte (verified per ticket against specific historic SHAs).
- No file in `tickets/` or `tickets/archive/` has more than one `--- body ---` or `--- log ---` separator after this PR.
- `validate` against `tickets/` reports only the pre-existing `0059 -> git-erg/0008` cross-repo Blocked-by warning, which is untouched by this PR.

### Minor observations

**consider: `appendToTicket` is empirically safe but the load-bearing invariant is unstated.** A naive read suggests the same aliasing risk: `beforeBody := raw[:idx]` then `append(beforeBody, '\n')` could mutate `raw[idx]`. I probed this with a test that supplies a ticket where the byte preceding `--- body ---` is *not* a newline (forcing the `HasSuffix` branch). The function is safe — but only because `result = append(result, sep...)` re-emits a fresh separator after `beforeBody`, masking the single-byte aliasing write at `raw[idx]`, and `bodyContent` is sliced from `raw[idx+len(sep):]` whose bytes are not touched. A short comment in `appendToTicket` naming this invariant ("re-emitting fresh `sep` masks any aliased write at `raw[idx]`") would prevent a future "simplification" from removing the masking and reintroducing the corruption.

**consider: the strict "exactly once" rule is a useful regression guard but is mildly opinionated.** A future ticket whose body legitimately contains literal `--- log ---` or `--- body ---` lines (e.g., a ticket documenting the format itself, line-anchored without code-fence escape) will now fail validation. Today no ticket does this — `tickets/FORMAT.md` is `.md` not `.erg`, and ticket 0060 has no body-internal separators. Not a blocker; flagging only because the constraint is stricter than the v1 spec text suggests.

**nofollow: `make check` is mentioned in the review skill but there is no `Makefile` in the repo.** `go test ./...` from `tickets/tools/go/` is the actual gate. Not in scope for this PR.

### Perspectives covered

| Perspective | Result |
|---|---|
| Correctness | bug fix verified; repairs match pre-corruption historic SHAs; tests would be red without fix |
| Consistency | FORMAT.md rule 11 text matches code; error messages name the rule; comments explain the bug |
| Scope | only the bug fix, validator strengthen, repairs, 0060 close, FORMAT update, STATE prune |
| Red team | probed `appendToTicket` aliasing — safe in practice (see `consider:` above) |
| Doc propagation | FORMAT.md and `Erg` struct comments updated in lockstep with code |

PR #82, opened 2026-05-02
refactor(ticket-new): delegate ID computation to erg next-id

MinhHaDuong:
Closes ticket 0074.

## Summary
- Replace the 5-line inline bash pipeline in `ticket-new/SKILL.md` step 1 with `erg next-id tickets/`
- The `erg` binary already provides this command; the inline script was redundant

## Test plan
- [ ] Read `skills/ticket-new/SKILL.md` — step 1 shows `erg next-id tickets/`, no bash pipeline
- [ ] `erg next-id` is available in the harness (`tickets/tools/go/erg`)

### DIFF
--- a/skills/ticket-new/SKILL.md
+++ b/skills/ticket-new/SKILL.md
@@ -15,7 +15,8 @@ from a conversation. Extract the intent and normalize to `%erg v1`.
 
 1. Determine the next ID:
    ```bash
-   erg next-id tickets/
+   ERG=${ERG:-tickets/tools/go/erg}
+   $ERG next-id tickets/
    ```
 
 2. Choose a slug: lowercase kebab-case, ASCII only (`[a-z0-9-]`), derived from the title.
--- a/skills/ticket-new/SKILL.md
+++ b/skills/ticket-new/SKILL.md
@@ -13,17 +13,10 @@ from a conversation. Extract the intent and normalize to `%erg v1`.
 
 ## Steps
 
-1. Determine the next ID — take the maximum across local files **and** the cached
-   `origin/main` ref, then increment. No network call; uses whatever the last
-   `git fetch`/`pull`/`push` left in the local ref store.
+1. Determine the next ID:
    ```bash
-   LOCAL=$(ls tickets/*.erg tickets/archive/*.erg 2>/dev/null \
-     | sed 's|.*/||; s|-.*||' | sort -n | tail -1)
-   REMOTE=$(git ls-tree --name-only origin/main tickets/ 2>/dev/null \
-     | grep '\.erg$' | sed 's|-.*||' | sort -n | tail -1)
-   MAX=$(printf '%s\n%s\n' "$LOCAL" "$REMOTE" | grep -v '^$' | sort -n | tail -1)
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
## Multi-perspective review — PR #82

**Verdict: approve** (with one consistency comment)

All 96 pytest tests pass. Exit criteria met: `skills/ticket-new/SKILL.md` step 1 reads `erg next-id tickets/` with no inline shell pipeline.

---

### Correctness — approve (high confidence)

Exit criteria satisfied. The inline 5-line bash pipeline is fully removed. `erg next-id tickets/` appears correctly in step 1.

**Behavioral delta**: the old pipeline checked both local filesystem AND `origin/main` via `git ls-tree` (no network call, but used the cached remote ref). `erg next-id` checks local `tickets/` and `tickets/archive/` only. In a multi-machine or concurrent-worktree scenario where a ticket was created on another machine and pushed but not yet fetched locally, `erg next-id` could produce a duplicate ID.

This delta is present in the `erg` binary itself (not introduced by this PR), and the old code's comment already framed it as "uses whatever the last git fetch/pull/push left in the local ref store" — meaning the intent was always local-only. The deletion is correct with respect to ticket 0074's scope.

nofollow: duplicate-ID risk from missing `git fetch` is a pre-existing concern in `erg next-id` (out of scope for this PR). Ticket 0059 (delegate to `erg pick`) is the right place to revisit if this becomes a problem.

---

### Consistency — comment (medium confidence)

`STATE.md` line 21 states: *"Skills resolve erg binary as `${ERG:-erg}` for PATH portability."*

The updated `ticket-new/SKILL.md` uses bare `erg next-id tickets/` without the `${ERG:-env}` resolution step. Compare `pick-ticket/SKILL.md` step 0:
```bash
ERG=$(command -v erg 2>/dev/null || echo "tickets/tools/go/erg")
```
`ticket-ready/SKILL.md` hard-codes `tickets/tools/go/erg` and provides a manual fallback section.

consider: `ticket-new` should either add the binary-resolution step (mirroring `pick-ticket`) or document why bare `erg` is acceptable here (e.g., the skill is interactive-only and not used in nightbeat sweeps where PATH may not include `erg`). As-is, it deviates from the stated convention and will fail silently on machines where `erg` is not on PATH.

This is not a regression from before this PR — the old version used a pure bash pipeline that didn't need the binary at all. The PR introduces `erg` as a dependency without adding the resolution guard. Not a blocker for an interactive skill, but worth noting.

---

### Scope — approve (high confidence)

Change is tightly scoped: one skill file (`skills/ticket-new/SKILL.md`) and one ticket file (`tickets/0074-ticket-new-use-erg-next-id.erg`). No unrelated changes. Ticket 0074 added correctly.

The ticket is still `Status: open` — close it on merge.

---

### Doc propagation — approve (high confidence)

`STATE.md` does not reference `ticket-new` step 1 directly, so no doc update needed there. `tickets/FORMAT.md` and `tickets/tools/go/main.go` are unaffected. No downstream doc drift detected.

---

*Reviewed by: Correctness · Consistency · Scope · Doc propagation agents*

PR #172, opened 2026-05-13
fix(beat): guard git checkout in housekeeping post-cleanup path

MinhHaDuong:
Closes ticket 0137. Mirrors the dirty-tree guard from _raid() (added in 0120) to the housekeeping cleanup path at ~line 906.

## Summary
- Adds a `git status --porcelain` check before `_git("checkout", default_branch)` in `_housekeeping_phase()`
- Returns `"aborted-dirty-tree"` if the working tree is dirty after the skill run, preventing a silent checkout failure
- Adds `"aborted-dirty-tree": "fail"` to `_HK_OUTCOME_MAP` so the caller maps it correctly
- Adds a test that patches `_git` to return non-empty stdout for `status --porcelain` and asserts the function returns `"aborted-dirty-tree"` without reaching checkout

## Test plan
- [x] `python3 -m pytest tests/test_beat.py -x -q` passes (153 tests)
- [x] New test `test_aborted_dirty_tree_when_post_skill_tree_dirty` explicitly asserts no checkout is called when tree is dirty

### DIFF
--- a/tests/test_beat.py
+++ b/tests/test_beat.py
@@ -1757,6 +1757,10 @@ class TestProjectConfigFields:
         cfg = beat.ProjectConfig(path=Path("/tmp/p"), max_turns_housekeeping=9999)
         assert cfg.max_turns_housekeeping == 2 * beat.MAX_TURNS_HOUSEKEEPING
 
+    def test_max_turns_pick_ticket_cap(self):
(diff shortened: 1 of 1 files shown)


MinhHaDuong:
Closing: branch contamination during creation. Replacing with a clean PR from t0137v3.

PR #149, opened 2026-05-12
feat: offload housekeeping git phase to shell script (#34)

MinhHaDuong:
## Summary

- Extracts the git sync, beat-skip expiry, and erg DAG check from the housekeeping skill into `scripts/housekeeping-git.sh`
- Updates `beat.py` to run the script as a subprocess before invoking the skill, logging `housekeeping-git: done` or `rc=N`
- Updates `skills/housekeeping/SKILL.md` step 1 to delegate to the script on interactive runs, and skip on beat.py runs (since beat.py already ran it)

## Test plan

- [ ] `bash -n scripts/housekeeping-git.sh` passes (syntax check)
- [ ] `scripts/housekeeping-git.sh` is executable
- [ ] `skills/housekeeping/SKILL.md` no longer contains old step 1a/1b prose (beat-skip json, erg check)
- [ ] `beat.py` logs `housekeeping-git: done` before the skill invocation
- [ ] Beat mode section in SKILL.md still makes sense

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Ticket: tickets/0034-housekeeping-split-git-and-ticket-phases.erg

PR #177, opened 2026-05-13
test(beat): add test_max_turns_pick_ticket_cap

MinhHaDuong:
Ticket: tickets/0148-test-max-turns-pick-ticket-cap.erg

Closes ticket 0148. Missing regression test for max_turns_pick_ticket cap added in PR #169.

PR #169, opened 2026-05-13
fix(beat): cap max_turns_pick_ticket in __post_init__

MinhHaDuong:
## Summary
- `ProjectConfig.__post_init__` already caps `max_turns_housekeeping` at `2 × MAX_TURNS_HOUSEKEEPING`
- `max_turns_pick_ticket` had no analogous cap — a bad `beat.json` overlay could cause runaway agent costs
- This PR adds the same enforcement for parity

Ticket: tickets/0143-beat-max-turns-pick-ticket-missing-cap-i.erg

Scope overflow: tickets/0148

## Test plan
- [ ] Existing `ProjectConfig` tests in `tests/test_beat.py` pass
- [ ] `python3 -m pytest tests/test_beat.py -x -q`

PR #146, opened 2026-05-12
feat(0132): raid merges APPROVED PRs after verify-gate + scope check

MinhHaDuong:
**Ticket:** tickets/0132-raid-merge-celebrate-phases.erg

## Summary

- Fold scope audit (old Phase 7) into verify-gate as a scope-containment check — the gate detects overflow and reports `suggested_disposition` (TICKETED/ESCALATE); the caller handles ticketing and PR annotation
- Add Phase 7 (Merge): sequential per-wave merge of APPROVED PRs, with `gh pr checkout` before each `/merge` call
- Add Phase 8 (Celebrate): runs `/celebrate` from updated main after all merges
- Add drift guard after Phase 2 (Imagine): exit criteria changes → ESCALATE, implementation approach changes → allowed
- Remove interactive/autonomous mode split — raid always merges now
- One REROLL allowed per PR (existing Phase 6 behavior, unchanged)
- Ticket 0131 (token economy review) rides along as a follow-up

## Test plan

- [ ] verify-gate verdict shape includes `scope_overflow` section
- [ ] verify-gate decision rules handle TICKETED vs ESCALATE scope overflow
- [ ] raid Phase 7 merge flow: checkout → mergeable check → `/merge` → error handling
- [ ] drift guard criteria after Imagine phase
- [ ] No stale "Phase 7 = scope audit" references remain
- [ ] All `gh` calls have harness-extension-point annotation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #182, opened 2026-05-13
fix(healthcheck,housekeeping): probe squash-merges by PR number

MinhHaDuong:
## Summary
- Replace broken regex probe `(feat|fix|chore|ticket)\(NNNN\)` with PR-number extraction from `Closed:` line
- Match `(#NNN)` in main's commit subjects — the actual squash-merge format
- Conservative fallback: when no PR# recorded, skip deletion (not false-positive delete)

## Test plan
- [x] Probe returns PASS for all 6 confirmed-merged branches (t0119, t0121, t0123, t0137, t0145, t0148)
- [x] Probe returns false for branches without PR# in Closed: (no false-positive deletion)

Closes #0149

### DIFF
--- a/skills/healthcheck/SKILL.md
+++ b/skills/healthcheck/SKILL.md
@@ -32,10 +32,10 @@ Parse the JSON output. Use its fields to populate checks 1–9 below without re-
    - **Closed-ticket branches** (name matches `t<NNNN>-*` and ticket NNNN has a `Closed:` line) where the squash-merge probe succeeds: classify as **fix-now** — safe to delete. Run this probe per branch:
      ```bash
      merge_base=$(git merge-base main <branch>)
-     ticket_id=$(echo <branch> | grep -oP 't\K\d+')
-     git log $merge_base..main --format="%s" | grep -qP "(feat|fix|chore|ticket)\($ticket_id\)"
+     pr_num=$(grep -oP '#\K[0-9]+' tickets/closed/$(ls tickets/closed/ | grep -P "^$(echo <branch> | grep -oP 't\K\d+')-" | head -1) 2>/dev/null | head -1 || true)
+     [ -n "$pr_num" ] && git log $merge_base..main --format="%s" | grep -qE "\(#${pr_num}\)"
      ```
-     If the grep exits 0, the branch was squash-merged into main. Emit one fix-now bullet per matching branch: `Delete local branch \`<branch>\` (closed ticket <NNNN>, squash-merged into main)`.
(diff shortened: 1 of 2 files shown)


MinhHaDuong:
## /verify actions

round: 1
adherence: skipped (escalated at setup)
review-pr: skipped (size-gate: 10 lines, 2 files)
simplify: skipped (escalated at setup)

## /verify-gate verdict

verdict: ESCALATE

/verify stopped: setup-phase circuit breaker — three issues require human judgment before review can proceed.

### 1. Missing ticket file (blocking)

The PR references `Closes #0149` and the branch is named `t0149-fix-squash-merge-probe`, but ticket `0149` does not exist in either `tickets/` (open) or `tickets/closed/`. The `/verify` spec requires the ticket file be locatable at setup; without it the exit criteria cannot be evaluated.

**Action needed:** Create `tickets/0149-fix-squash-merge-probe.erg` (or confirm the ticket was inadvertently omitted and create it before re-verification).

### 2. PR-number anchor is not anchored to Closed: line

The new probe:
```bash
pr_num=$(grep -oP '#\K[0-9]+' tickets/closed/<ticket-file> 2>/dev/null | head -1 || true)
```
grabs the **first** `#NNN` anywhere in the ticket file, not specifically from the `Closed:` header. A ticket whose body references an earlier PR number (e.g., "see PR #169") would yield the wrong `pr_num`, causing the probe to search for the wrong commit subject and either false-positive (wrong PR happens to be in main's range) or false-negative (skip valid deletion).

Actual example: `tickets/closed/0148-test-max-turns-pick-ticket-cap.erg` has `#177` on line 5 (`Closed:`) and `#169`, `#143`, `#146`, `#169` later in the body. `head -1` returns `177` only because `Closed:` is early. If the body had mentioned an earlier PR first, the probe would misbehave.

**Suggested fix:** Anchor the grep to the `Closed:` line:
```bash
pr_num=$(grep -oP '^Closed:.*#\K[0-9]+' tickets/closed/<ticket-file> 2>/dev/null | head -1 || true)
```

### 3. Self-referential gap

The new probe relies on a ticket file existing in `tickets/closed/` for the branch being checked. This PR itself has no ticket file, so after merge the probe will find nothing and conservatively skip deletion of `t0149-fix-squash-merge-probe`. This is not a blocker but is worth noting — a ticket file should be created anyway (see issue 1).

---

telemetry: wall=180s agents=0 tokens=na cost~=na

MinhHaDuong:
Closing: original branch had wrong commits (execute agent wrote to wrong branch). Replaced by clean t0149-v2 branch.

PR #202, opened 2026-05-14
fix(beat): lazy-load PROJECTS to unblock pytest-guard

MinhHaDuong:
## Summary
Module-level `PROJECTS = load_projects(...)` exits at import time when projects.json is missing. The pytest-guard CI job fails before any test can run. Solution: lazy-load via cached accessor _get_projects() deferred to first call.

## Changes
- Replace module-level PROJECTS constant with _PROJECTS_CACHE + _get_projects()
- Update call sites in _pick_project() and main() fallback logic
- Update 3 test patches to mock _PROJECTS_CACHE instead of PROJECTS

## Test Plan
- [x] All 159 test_beat.py tests pass locally
- [x] Verified no lingering PROJECTS references except comments
- [x] import beat now succeeds even with missing projects.json
- [x] CI pytest-guard job can now run

**Ticket:** tickets/0161-ci-pytest-guard-fails-load-projects-exit

🤖 Generated with Claude Code

### DIFF
--- a/scripts/beat.py
+++ b/scripts/beat.py
@@ -90,7 +90,9 @@ DRY_RUN: bool = os.environ.get("BEAT_DRY_RUN") == "1"
 class _State:
     beat_start: float = 0.0
     project: Path | None = None
-    current_proc: subprocess.Popen | asyncio.subprocess.Process | None = field(default=None, repr=False)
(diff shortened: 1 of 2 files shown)


MinhHaDuong:
Contaminated with both 0160 and 0161 changes. Closing to re-execute 0160 and 0161 separately.

PR #223, opened 2026-05-29
feat(0178): housekeeping step 2.6 — archive closed tickets and remove stale live duplicates

MinhHaDuong:
## Summary

- Adds step 2.6 to `/housekeeping` between the exit-criteria audit (2.5) and fix-now (3)
- Runs `erg archive tickets/` to move tickets with a `Closed:` header from `tickets/` to `tickets/closed/`
- Removes stale live duplicates: any `tickets/*.erg` whose basename already exists in `tickets/closed/`
- Uses `shopt nullglob` so `git rm` does not fail when no `.erg` files remain
- Fixes are staged and folded into the existing `chore: housekeeping fixes (sweep)` commit — no separate commit

## Test plan

- [ ] `make check` passes (pre-existing failure in ticket 0179 is unrelated)
- [ ] Step 2.6 appears in correct position in `skills/housekeeping/SKILL.md` between steps 2.5 and 3
- [ ] Manual: run `/housekeeping` on a repo with an un-archived closed ticket; confirm it is moved to `tickets/closed/`
- [ ] Manual: run `/housekeeping` on a repo with a duplicate live ticket; confirm `git rm` removes the live copy

**Ticket:** tickets/0178-housekeeping-ticket-corpus-hygiene.erg

### DIFF
--- a/skills/housekeeping/SKILL.md
+++ b/skills/housekeeping/SKILL.md
@@ -71,6 +71,7 @@ Run full repo housekeeping and act on every finding.
 
    ```bash
    tickets/erg archive tickets/
+   git add tickets/closed/
    ```
 
    This moves any ticket with a non-empty `Closed:` header from `tickets/`
@@ -83,7 +84,7 @@ Run full repo housekeeping and act on every finding.
    for f in tickets/*.erg; do
      base=$(basename "$f")
      if [ -f "tickets/closed/$base" ]; then
-       git rm -f "$f"
+       git rm "$f"
      fi
    done
    shopt -u nullglob
--- a/skills/housekeeping/SKILL.md
+++ b/skills/housekeeping/SKILL.md
@@ -67,6 +67,32 @@ Run full repo housekeeping and act on every finding.
    Process all open tickets unconditionally (not just candidates).
    Batch-closing is allowed here: close every qualifying ticket in one pass.
 
+2.6. **Archive closed tickets.** Skip if `tickets/` is absent.
+
+   ```bash
+   tickets/erg archive tickets/
+   ```
+
+   This moves any ticket with a non-empty `Closed:` header from `tickets/`
+   into `tickets/closed/`. After archiving, remove any file from `tickets/`
+   that already exists in `tickets/closed/` (duplicate committed under both
+   paths):
+
+   ```bash
+   shopt -s nullglob
+   for f in tickets/*.erg; do
+     base=$(basename "$f")
+     if [ -f "tickets/closed/$base" ]; then
+       git rm -f "$f"
+     fi
+   done
+   shopt -u nullglob
+   ```
+
+   If any files were moved or removed, stage and include them in the step 3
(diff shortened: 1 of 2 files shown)


MinhHaDuong (commented):
## Multi-perspective review

**Risk level:** Standard. Single file, 26 lines added to a skill doc. Agents: Correctness + Consistency + Scope.

---

### CI failure (agnostic-guard)

The one failing check is in `tickets/0179-refresh-state-ignores-path-arg.erg`, a pre-existing issue unrelated to this PR. The diff introduces no new agnostic violations.

---

### Correctness

**`erg archive` staging behavior (blocker candidate — investigated)**

`erg archive` does _not_ use `git mv`. It:
1. Removes the source file from the git index (staged deletion).
2. Creates the destination file as an **untracked** file in `tickets/closed/`.

Verified by running `erg archive` in a test repo: `git status --porcelain` shows `D tickets/0001-test.erg` + `?? tickets/closed/`. The destination is left untracked until an explicit `git add tickets/closed/` is run.

The step instruction says _"stage and include them in the step 3 commit"_ — this is sufficient direction for an LLM agent to infer that both the deletion and the new files need staging. However, the parallel duplicate-removal section spells out exact `git rm -f` commands, while the archive section relies on the agent knowing to `git add tickets/closed/` after `erg archive`. An agent that only follows the literal bash snippet would commit only the deletion, silently losing the archived ticket from history.

consider: Add explicit staging instructions after the `erg archive` code block, parallel to how `git rm -f` is explicit in the duplicate loop. For example:

```bash
tickets/erg archive tickets/
git add tickets/closed/   # stage newly archived files
```

**Guard completeness**

The ticket's EC4 requires guarding against both `tickets/` and `tickets/closed/` being absent. The diff guards only against `tickets/` being absent. Analysis shows this is functionally correct:
- If `tickets/closed/` does not exist, `erg archive` creates it on demand.
- The duplicate loop's `test -f tickets/closed/$base` returns false when the directory is absent, so no incorrect `git rm` fires.

nofollow: EC4 wording is stricter than necessary; the implementation is safe without the second guard.

**Step position**

Step 2.6 correctly appears between 2.5 (exit-criteria audit) and 3 (fix-now). Verified on PR branch.

---

### Consistency

The step follows the document's conventions: numbered float (2.6), bold heading, bash fenced blocks, and prose describing semantics. Consistent with steps 1.5, 2.5. The commit message reference (`chore: housekeeping fixes (sweep)`) matches step 3's existing convention.

---

### Scope

Only `skills/housekeeping/SKILL.md` is modified. 26 lines added, 0 deleted. No out-of-scope changes.

---

### Summary

The change is correct and well-scoped. The one actionable finding is the ambiguous staging instruction for `erg archive`: the archived files land untracked and need an explicit `git add tickets/closed/`, which the current text implies but does not state. I'm posting this as `consider:` rather than blocking because a competent agent will infer the intent — but the inconsistency with the explicitly-coded `git rm -f` in the duplicate loop is worth tightening.

**Verdict: comment** (one consider-level finding; no blockers)

PR #230, opened 2026-05-29
fix(0181): beat-skip.json jq filter — map(select) replaces .entries

MinhHaDuong:
`housekeeping-git.sh` was silently dropping all beat-skip expiry logic
because `.entries` errors on a top-level array. Replace with
`map(select(type == "object" and ...))` which handles the canonical
object-array schema and tolerates legacy string arrays without erroring.

Adds `tests/test_housekeeping_git.py` covering all three input schemas
(object array, legacy string array, empty) plus the no-file case.

**Ticket:** tickets/0181-fix-beat-skip-json-schema-mismatch-in-ho.erg

### DIFF
--- a/scripts/housekeeping-git.sh
+++ b/scripts/housekeeping-git.sh
@@ -9,7 +9,7 @@ SKIP_FILE=".git/beat-skip.json"
 if [[ -f "$SKIP_FILE" ]]; then
     NOW=$(date -u +%Y-%m-%dT%H:%M:%SZ)
     TMP=$(mktemp)
-    if jq --arg now "$NOW" '.entries |= [.[] | select(.until == null or .until > $now)]' "$SKIP_FILE" > "$TMP"; then
+    if jq --arg now "$NOW" 'map(select(type == "object" and (.until == null or .until > $now)))' "$SKIP_FILE" > "$TMP"; then
         mv "$TMP" "$SKIP_FILE"
     else
         rm -f "$TMP"
--- a/tests/test_housekeeping_git.py
+++ b/tests/test_housekeeping_git.py
@@ -1,0 +1,79 @@
+"""Tests for the beat-skip expiry logic in scripts/housekeeping-git.sh."""
+
+import json
+import subprocess
+from pathlib import Path
+
+SCRIPT = Path(__file__).parent.parent / "scripts" / "housekeeping-git.sh"
+
+
+def _fake_repo(tmp_path):
+    """Minimal fake git repo layout the script expects."""
+    subprocess.run(["git", "init", "-q"], cwd=str(tmp_path), check=True)
+    (tmp_path / "tickets").mkdir()
+    return tmp_path
+
+
+def _run(cwd):
+    return subprocess.run(
+        ["bash", str(SCRIPT)],
+        cwd=str(cwd),
+        capture_output=True,
+        text=True,
+    )
+
+
+def test_object_array_prunes_expired(tmp_path):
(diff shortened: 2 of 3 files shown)


MinhHaDuong:
## /verify actions

round: 1
adherence: FAIL — 1 blocking, 1 non-blocking
review-pr: skipped — see below for manual review findings
simplify: skipped (adherence blocking)
fix agent: not spawned — see verdict

## /verify-gate verdict

verdict: REROLL

**Exit criteria (ticket 0181):**
- `housekeeping-git.sh` expiry filter uses `map(select(type == "object" ...))`, not `.entries`: ADDRESSED — line 12 of `scripts/housekeeping-git.sh`
- Legacy string-array entries tolerated without error: ADDRESSED — `type == "object"` guard filters them silently
- No jq stderr against all three schemas (object array, string array, empty): ADDRESSED — verified manually and by tests
- Expired entries pruned, indefinite preserved: ADDRESSED — tested by `test_object_array_prunes_expired`

**Blocking adherence violations:**

1. **feat(0180) delivers an untracked hook — incomplete.** The commit message says "Add a skills-catalog-guard block to the pre-commit hook", but:
   - `.git/hooks/pre-commit` is untracked; it is not in version control and never will be via a normal diff
   - The repo has no `hooks/` directory (checked-in source that `on-start.sh` would install from)
   - `git config core.hooksPath` = `.git/hooks` — the default untracked location
   - `tests/test_guard_skills_catalog.sh` exercises a *synthetic* hook built inside `$TMPDIR`, not the real repo hook
   - Conclusion: merging this PR does not ship the skills-catalog guard to any other machine or fresh clone. The test passes locally but validates a phantom.

2. **Two tickets on one branch (scope violation).** Per `rules/workflow.md` "One PR closes exactly one ticket." This branch carries both ticket 0180 work (`feat(0180): skills-catalog-guard`) and ticket 0181 work (`fix(0181): beat-skip.json jq filter`). The `**Ticket:** tickets/0181-...` line means `erg-pr-merge` closes 0181 only; 0180 stays open indefinitely with its work already merged and its test unverifiable.

**Non-blocking adherence violation:**

3. **Missing `@pytest.mark.integration` markers.** `tests/test_housekeeping_git.py` calls `subprocess.run()` in `_fake_repo()` and `_run()` but has no `@pytest.mark.integration` markers. Per `rules/coding-python.md`: "Tests using `subprocess.run()`: mark `@pytest.mark.integration`." This is mechanical — CI runs all tests regardless — but it breaks the contract that `make check-fast` excludes subprocess tests.

**Unresolved review comments:** none prior

**Rationale:**

The 0181 jq fix is correct, well-tested, and closes cleanly against all exit criteria. The problem is what surrounds it. The feat(0180) commit is structurally incomplete: its intent (prevent README.md drift at commit time) is sound, but the delivery mechanism is broken — the hook lives only on the author's machine, the test proves nothing about the real repo, and `erg-pr-merge` won't know to close 0180. The workflow rule on single-ticket PRs exists precisely to prevent this kind of "shadow work" situation.

**Required fixes for round 2:**
1. Either: add `hooks/pre-commit` (tracked, installed by `on-start.sh` once the directory exists) containing the skills-catalog-guard block, and update the test to reference the real installed hook path; OR split 0180 off into a separate PR that can be properly reviewed and closed.
2. Add `@pytest.mark.integration` to all four test functions in `tests/test_housekeeping_git.py`.

telemetry: wall=~8min agents=1 tokens=na cost~=$na

PR #241, opened 2026-06-03
ticket(0189): PreToolUse guard — block cd-to-primary mutations in worktree sessions

MinhHaDuong:
## Summary

Adds a PreToolUse `Bash` guard (`scripts/guard-cd-primary-repo.sh`) that blocks
`cd <primary-repo-root> && <mutating git/erg>` during a worktree session, with a
committed shell test suite, hook registration in `settings.json`, and a rule in
`rules/workflow.md`.

In an `EnterWorktree` session the shell already sits in the worktree and resets
there after every command. Prefixing `cd ~/<repo> &&` silently cd's into the
PRIMARY checkout (on `main`), so the `git`/`erg` mutation lands on the wrong tree.

## Motivation — three incidents in one autonomous session (2026-06-03, aedist-technical-report)

- `erg close` + a stray `git add -A` landed a **493-file commit on primary `main`** (recovered via `git reset --mixed HEAD~1`).
- `git switch -c t0357` created the branch in the **PRIMARY repo** while the worktree stayed on a stale branch; worktree-path edits were stranded on the wrong branch (recovered via stash + re-branch).
- `uv run pytest` ran against the primary repo's **stale files**, masking RED tests.

A same-session memory did not prevent recurrence — the failure is a reflex
(prefix `cd <repo> &&` for "safety"), so a note is too weak. A mechanical guard
is the right altitude, mirroring the existing worktree-path Edit/Write guard.

## Design

- Pure-string worktree detection via the payload's `.cwd` (marker `/.claude/worktrees/`); no git calls in the hot path.
- Fast path: command without a `cd` word → exit 0.
- Extracts the first `cd <target>`, strips quotes/trailing slash, expands leading `~` and `$HOME`. Only when target equals the primary root AND the command carries a mutating verb (`git commit|add|switch|checkout|reset|merge|rebase|push|mv|rm|stash` or `erg close|archive`) does it block (exit 2) with the `git -C <path>` alternative.
- Read-only inspection (`git status`, `git log`) passes **silently**.
- Fails **open** when `jq` is missing — narrow advisory guard, not a security boundary (contrast `guard-destructive-bash.sh`, which fails closed on real data loss).

### Deliberate scope (YAGNI)

- Only `cd && <git/erg>` is handled — not `sed -i`, `rm`, or `>>` redirects.
- The `cd` must be the leading token; `echo x && cd ~/repo && git commit` is not caught. None of the observed incidents or required test cases exercise this, and broadening would enlarge the hot path for little gain.

## Tests

`tests/test_guard_cd_primary_repo.sh` drives the guard purely through the JSON
payload (`.cwd` + pinned `HOME=/home/testuser`), so it is environment-independent
and needs no test-only backdoor in the production script. 16 cases: blocks for
each mutating verb (incl. tilde form), allows non-worktree / bare-command /
in-worktree-subtree / unrelated-dir / empty / no-`.cwd` (fail-safe) / no-jq
(fail-open), and asserts read-only cases are silent (empty stderr).

## CI note

`python3 -m pytest tests/test_bash_suites.py` is **green in a fresh checkout**
(verified via a clean `git clone` where `.git` is a directory — identical to CI).
When run *inside a git worktree*, the pre-existing `test_pretooluse_worktree_path_guard.sh`
case 4 fails because `.git` is a *file* there, tripping that sibling guard's
`_in_worktree()` heuristic. This is unrelated to this change (the diff does not
touch that script) and is a candidate for a follow-up brittleness ticket.

**Ticket:** tickets/0189-worktree-cd-primary-repo-guard.erg

Related follow-up: tickets/0192-worktree-path-guard-test-brittle-in-worktree.erg — pre-existing in-worktree brittleness of the sibling guard test, surfaced by this wave (ticket bundled on this branch, stays open on merge).

### DIFF
--- a/rules/workflow.md
+++ b/rules/workflow.md
@@ -41,6 +41,8 @@ During an `EnterWorktree` session, `Edit`/`Write`/`Read` tools accept any absolu
 
 For everything else (source code, manuscript prose, data files): if `git branch --show-current` is `main`, stop and switch to a branch first.
 
+The same trap exists on the Bash surface: prefixing a command with `cd <primary-repo-root> &&` silently lands `git`/`erg` mutations on the primary checkout (on main) instead of the worktree branch. A PreToolUse guard blocks `cd <primary-repo-root> && <mutating git/erg>` during worktree sessions; read-only inspection (`git status`, `git log`) is not penalised, and an intentional primary-repo mutation should use `git -C <path>` rather than a `cd`.
+
 # Escalation Protocol
 
 When stuck, escalate progressively:
--- a/scripts/guard-cd-primary-repo.sh
+++ b/scripts/guard-cd-primary-repo.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+set -euo pipefail
+# PreToolUse hook: block `cd <primary-repo-root> && <mutating git/erg>` during a
+# worktree session. In an EnterWorktree session the shell already sits in the
+# worktree and resets there after every command; prefixing `cd ~/<repo> &&`
+# silently cd's into the PRIMARY checkout (on main), so the mutation lands on
+# the wrong tree. Read-only inspection (status/log/diff) is not penalised.
+# See ticket 0189. Sibling of guard-destructive-bash.sh / the worktree-path guard.
+#
+# Exit 0 = allow, exit 2 = block with a stderr message.
+#
+# Fail-OPEN (not fail-closed) when jq is missing: this is a narrow advisory
+# guard against one reflex, not a security boundary — a missed warning is far
+# cheaper than blocking every Bash call on a box without jq. (Contrast
+# guard-destructive-bash.sh, which fails closed because it gates real data loss.)
+#
+# Deliberate YAGNI: only `cd && <git/erg>` is handled. We do NOT inspect
+# `sed -i`, `rm`, or `>>` redirects targeting the primary repo — those are not
+# the observed reflex and would broaden the hot path for little gain.
+
+input=$(cat)
+
+command -v jq &>/dev/null || exit 0
+
+cmd=$(echo "$input" | jq -r '.tool_input.command // empty' 2>/dev/null || true)
+[ -z "$cmd" ] && exit 0
+
+# Fast path: no `cd` word → nothing to guard. One grep, before the second jq.
+echo "$cmd" | grep -qwP 'cd' || exit 0
+
+cwd=$(echo "$input" | jq -r '.cwd // empty' 2>/dev/null || true)
+[ -z "$cwd" ] && exit 0  # fail-safe: cannot tell if this is a worktree session
+
+# Worktree detection (pure string, no git calls): cwd under .claude/worktrees/.
+case "$cwd" in
+    */.claude/worktrees/*) ;;
+    *) exit 0 ;;
+esac
+# primary_root = the prefix before the worktree marker.
+primary_root="${cwd%%/.claude/worktrees/*}"
+
+# Extract the first `cd <target>` token (before &&, ;, |, or whitespace end).
+# Take the text after the first `cd ` and lop off at the first delimiter.
+rest="${cmd#*cd }"
+target="${rest%%[ ;&|]*}"
+[ -z "$target" ] && exit 0
+
+# Strip surrounding quotes and a trailing slash.
+target="${target#\"}"; target="${target%\"}"
+target="${target#\'}"; target="${target%\'}"
+target="${target%/}"
+
+# Expand leading `~` and a literal `$HOME` using $HOME.
+target="${target/#\~/$HOME}"
+target="${target//\$HOME/$HOME}"
+
+# Only the primary repo root itself is the offence; subtrees (incl. the
+# worktree path, which lives under primary_root) are fine.
+[ "$target" = "$primary_root" ] || exit 0
+
+# Mutating git verbs or erg verbs after the cd → block.
+if echo "$cmd" | grep -qP '\bgit\s+(commit|add|switch|checkout|reset|merge|rebase|push|mv|rm|stash)\b' \
+   || echo "$cmd" | grep -qP '\berg\s+(close|archive)\b'; then
+    echo "BLOCKED: 'cd $target && ...' targets the PRIMARY repo (on main), but this is a worktree session (cwd=$cwd)." >&2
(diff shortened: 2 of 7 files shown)


MinhHaDuong:
## review-pr — multi-perspective (correctness · consistency · scope · red-team · doc-propagation)

Verdict: **request-changes** (one blocker). The guard is well-shaped, well-documented, and the new suite is green (16 cases, clean checkout and in a linked worktree). `make check` passes (RC=0, CI-equivalent). One correctness bug contradicts the ticket's own invariant.

---

### BLOCKER — false block on `<worktree-mutation> && cd <primary>` (correctness + red-team, high confidence)

The mutating-verb regex scans `$cmd` (the *whole* command) instead of the segment *after* the `cd`. So a command that mutates the worktree first and then cds to primary is wrongly blocked, even though the mutation correctly hit the worktree and nothing mutating follows the `cd`.

`verifiable:` reproduced against `origin/t0189-cd-guard:scripts/guard-cd-primary-repo.sh`:

```
payload: {"command":"git add . && cd /home/testuser/repo","cwd":".../.claude/worktrees/t001"}
         (HOME=/home/testuser)
result:  exit 2  (BLOCKED)
expected: exit 0  — `git add .` ran in the worktree (correct); the trailing `cd` has no mutation after it
```

This directly violates the ticket's stated invariant: *"Fail-safe over fail-closed: a false block is more annoying than a missed warning."* The PR's Design section argues blocking on "command carries a mutating verb" is deliberate — but intending to gate `cd → mutate` is not the same as gating `mutate → cd`. The latter is the safe, common case (stage in worktree, then hop to primary for a read).

Root cause: `script:line` — the block applies `grep` to `$cmd`, but the `cd` target was parsed from `rest="${cmd#*cd }"`. Fix: scan the post-`cd` segment (`$rest`) for the mutating verb, not the entire `$cmd`. Add a test case for `git add . && cd $PRIMARY` (expect allow) — the current 16-case suite has no `mutate && cd` ordering, which is why it stays green over the bug.

---

### consider: PR's "CI note" about in-worktree red is not reproducible (doc-propagation, medium confidence)

The PR body and the bundled ticket 0192 claim `test_pretooluse_worktree_path_guard.sh` case 4 fails when run from inside a linked worktree (`.git` is a file). I could not reproduce it: in a freshly added linked worktree (confirmed `.git` is a FILE), `bash tests/test_pretooluse_worktree_path_guard.sh` exits 0 — all 7 cases pass — and `pytest tests/test_bash_suites.py` reports 9 passed. The new suite also passes in-worktree. Not load-bearing for this PR (0192 is bundled to stay open on merge), but the stated justification for 0192 may be stale or environment-specific. Author should re-confirm 0192's premise before working it.

---

### nofollow: second ticket bundled on the branch (scope)

The branch adds `tickets/0192-...erg` (62 lines) alongside the 0189 work. The PR body's `**Ticket:** tickets/0189-...` line means `erg-pr-merge` closes only 0189; 0192 stays open. Consistent with the stated intent and the "one PR closes one ticket" rule. No action.

---

### Confirmed good (no action)

- consistency: matches sibling guards (`guard-destructive-bash.sh`, `guard-commit-on-main.sh`) — same stdin/jq/exit-2 shape, PCRE `\b`/`\s` usage, hook registration block in `settings.json`.
- The deliberate fail-OPEN-on-missing-jq (vs siblings' fail-closed) is justified in-script: this is an advisory reflex-guard, not a data-loss boundary. Sound.
- `git add .` / read-only-at-primary (`git status`, `git log`) pass silently — verified.
- doc: the new `rules/workflow.md` paragraph accurately describes the guard's intended behavior.

The blocker is a one-segment fix plus one test case. Everything else is clean.

MinhHaDuong:
## /verify actions

round: 2
adherence: PASS — 0 blocking
review-pr: 1 comment posted (correctness blocker — false-block of `mutate && cd <primary>`)
simplify: 0 fixes applied (code already clean)
fix agent: 1 commit (f1b6067) — scan post-`cd` text (`$rest`) not whole command (`$cmd`) for mutating verbs; +2 RED→GREEN test cases

## /verify-gate verdict

verdict: APPROVED

Exit criteria (tickets/0189-worktree-cd-primary-repo-guard.erg):
- PreToolUse Bash hook detects cd-to-primary in worktree session: ADDRESSED — `settings.json` registers the `Bash` matcher → `$HOME/.claude/scripts/guard-cd-primary-repo.sh` (timeout 5); pure-string `.cwd` detection of `/.claude/worktrees/`.
- Mutating git/erg blocked, read-only not penalised: ADDRESSED — blocks `cd <primary> && <git commit|add|switch|checkout|reset|merge|rebase|push|mv|rm|stash | erg close|archive>`; `git status`/`git log` pass silently. Round-1 blocker (whole-command verb scan false-blocked `git add . && cd <primary>`) fixed in f1b6067 by scanning post-`cd` text.
- Predicate has unit test covering fire/no-fire cases: ADDRESSED — `tests/test_guard_cd_primary_repo.sh`, 18/18 green (all mutating verbs incl. tilde; non-worktree, bare-command, subtree, unrelated-dir, empty, no-`.cwd` fail-safe, no-`jq` fail-open; + the two new mutate-before-`cd` ordering cases).
- Documented next to the Worktree-paths rule: ADDRESSED — `rules/workflow.md` adds the Bash-surface trap + `git -C <path>` guidance.
- `make check` green: ADDRESSED — skills-catalog in sync, agnostic check clean.

Unresolved simplify: none
Adherence: PASS (0)

Rationale:
The guard is at the correct altitude — a mechanical PreToolUse sibling of the existing worktree-path Edit/Write guard, motivated by three same-session incidents that a memory note failed to prevent. The `/verify` review fan-out found one real correctness blocker (empirically reproduced): the mutating-verb regex scanned the whole command, false-blocking the safe `git add . && cd <primary>` pattern and contradicting the ticket's own fail-safe invariant. The round-1 fix scans the post-`cd` remainder instead; this keeps every block case (including the tilde form) firing while clearing the false positive, with two regression tests that were RED before the change. All exit criteria are now met, all 7 CI checks were green pre-fix and the fix only touches the guard + its auto-discovered test. The one pre-existing in-worktree pytest failure (`test_pretooluse_worktree_path_guard.sh` case 4) is unrelated to this diff, environment-dependent, and already ticketed as 0192 (bundled on this branch, stays open on merge — only `**Ticket:** 0189` auto-closes).

telemetry: wall=na agents=3 tokens=na cost~=na

PR #240, opened 2026-06-03
ticket(0187): /release skill — pre-release audits, GPG pause, download-URL update

MinhHaDuong:
## Summary

Adds the `/release` skill encoding the release checklist agreed after git-erg 0181-0184: sign tags at release cadence (not CI cadence) and pin the download URL to the latest signed tag so CI rebuilds never invalidate attested downloads.

- `skills/release/SKILL.md` — pre-flight gate (`make check` + `needs-human` ticket scan, both HALT), two parallel background audit agents (security + UX dry-run, one ticket per finding), HIGH-finding stop, download-URL update, **human-only GPG signing pause** (the skill never signs), `git verify-tag` on resume, deferred blog-post stub, summary. Forge-agnostic prose, no avian analogies.
- `skills/release/rewrite-download-url` — executable bash helper. Rewrites the tag-like segment (date `YYYY-MM-DD` or semver) inside `/raw/<tag>/` and `/download/<tag>/` URL positions, forge-generic (matches path shape, not host). Exit code keys off "a tag-bearing URL was found", not "the file changed" — so re-running with the same tag is idempotent (exits 0, no-op) while a file with no such URL exits non-zero.
- `tests/test_release_rewrite_url.sh` — TDD-first, four cases: single URL, two URLs in one file, no-match (non-zero exit), idempotent re-run. Auto-discovered by `test_bash_suites.py`.
- README catalog regenerated (`make skills-catalog`); `make check-skills-drift` and `check-agnostic.sh` pass.

## Tests

- New suite green (4/4 cases).
- `make check` passes (skills-drift + agnostic-tickets).
- `python3 -m pytest tests/test_bash_suites.py -q`: 8 passed including the new suite. One pre-existing unrelated failure (`test_pretooluse_worktree_path_guard.sh`) confirmed failing on `origin/main` too — out of scope.

**Ticket:** tickets/0187-release-skill-automate-pre-release-audits-sign-tag.erg

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #243, opened 2026-06-03
ticket(0189): PreToolUse guard — block cd-to-primary mutations in worktree sessions

MinhHaDuong:
## Summary

PreToolUse Bash guard that blocks `cd ~/<repo> && <mutating git/erg verb>`
inside an EnterWorktree session, where the shell already sits in the
worktree. Prefixing `cd ~/<repo> &&` silently cd's into the PRIMARY
checkout (on main), so the mutation lands on the wrong tree. This bit one
autonomous session three times on 2026-06-03 (a 493-file commit on primary
main, a branch created in the wrong repo, pytest run against stale primary
files). A same-session memory did not prevent recurrence — the failure is a
reflex, so a mechanical guard is the right altitude, mirroring the existing
worktree-path Edit/Write guard.

The guard does pure-string worktree detection via the payload `.cwd` (no git
calls in the hot path), extracts the first `cd` target, expands `~`/`$HOME`,
and blocks (exit 2) only when the target equals the primary repo root AND the
command carries a mutating git/erg verb. Read-only inspection (status/log)
passes silently. Fails OPEN when jq is missing — a narrow advisory guard, not
a security boundary.

Also bundles the follow-up ticket 0192 (file only, no fix) per the
bundle-follow-up convention: the worktree-path-guard test brittleness was
surfaced independently by this raid wave's execute agents.

## Tests

- `tests/test_guard_cd_primary_repo.sh`: 17 cases, all green (drives the guard
  purely via JSON payload `.cwd` + pinned HOME — environment-independent, no
  test-only backdoor).

## Note for reviewers

Running `pytest tests/test_bash_suites.py` from inside a linked worktree shows
one pre-existing red: `test_pretooluse_worktree_path_guard.sh` case 4. It is
**not** introduced by this branch (the test file is byte-identical to
origin/main) and is CI-green on a fresh checkout where `.git` is a directory.
Tracked by ticket 0192 — does not block this PR.

**Ticket:** tickets/0189-worktree-cd-primary-repo-guard.erg

### DIFF
--- a/rules/workflow.md
+++ b/rules/workflow.md
@@ -41,6 +41,8 @@ During an `EnterWorktree` session, `Edit`/`Write`/`Read` tools accept any absolu
 
 For everything else (source code, manuscript prose, data files): if `git branch --show-current` is `main`, stop and switch to a branch first.
 
+The same trap exists on the Bash surface: prefixing a command with `cd <primary-repo-root> &&` silently lands `git`/`erg` mutations on the primary checkout (on main) instead of the worktree branch. A PreToolUse guard blocks `cd <primary-repo-root> && <mutating git/erg>` during worktree sessions; read-only inspection (`git status`, `git log`) is not penalised, and an intentional primary-repo mutation should use `git -C <path>` rather than a `cd`.
+
 # Escalation Protocol
 
 When stuck, escalate progressively:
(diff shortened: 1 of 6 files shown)


MinhHaDuong:
Closing: rogue artifact. A /verify sub-agent pushed the orchestrator's local working branch (t0189-followup) and opened this PR at 20:50Z during the #240/#241 verification window. All content here is already on main via PR #241 (merged). Incident logged on ticket 0193 (verify sub-skill reliability).

PR #259, opened 2026-06-04
feat(0198): erg-pr-merge queues auto-merge, drops the CI-watch loop

MinhHaDuong:
## Summary

`erg-pr-merge` no longer babysits CI. It now:
- Pushes the ticket-close commit **unconditionally** (drops the erg-only fast path that skipped the push and stranded the close commit — git-erg#256).
- Queues the merge with `gh pr merge --merge --auto`, letting GitHub's `main-required-checks` ruleset gate the landing. The `--auto` call is the condition of an `if` so the exit code drives a fallback.
- **Fallback** for repos without `allow_auto_merge`: the legacy `timeout 600 gh pr checks --watch` then plain `gh pr merge`.
- **Step 4 gating**: the return-to-base fast-forward now runs only on the fallback path. With `--auto` the merge is deferred, so fast-forwarding the local base immediately would be wrong; instead it prints a sync-after-it-lands note.
- Prunes the two CI-race bullets (mergeability `UNKNOWN`, "N checks still running") from the `rules/git.md` merge-bounce recovery section; keeps the idempotency and detached-HEAD bullets. Stamp bumped to 2026-06-04.
- Adds a one-line note to `skills/merge/SKILL.md`.

## Test plan

`bash tests/test_erg_pr_merge.sh` — 3 pre-existing cases unchanged, 3 new cases added (RED-verified against the unmodified script, GREEN after the edits):

1. **Auto-merge happy path** — asserts the merge log contains `--auto` and the checks log has NO `--watch` line (no CI poll).
2. **Fallback** — `STUB_AUTO_FAILS=1`: asserts a `--auto` attempt was made, the checks log was watched (`--watch`), and a plain `--merge` (without `--auto`) followed (`grep -v -- '--auto' | grep -q -- '--merge'`, not a tautological grep).
3. **Push-before-merge regression** (git-erg#256 strand) — erg-only branch: asserts `origin/$BRANCH` exists and `$BRANCH` is an ancestor of it, proving the close commit was pushed before the merge.

Fixture deviation (justified): added `git config receive.denyCurrentBranch ignore` in `seed_repo` because the close commit is now pushed unconditionally to the self-origin whose `$BRANCH` is checked out. Without it the real push would be refused and abort under `set -e`. This is also what makes case 3 assertable (the push is real).

`make check` passes (skills-drift + agnostic-tickets). The one failing entry in `test_bash_suites.py` (`test_pretooluse_worktree_path_guard.sh`) is a pre-existing, environment-specific failure byte-identical to `origin/main`, unrelated to this change.

**Ticket:** tickets/0198-erg-pr-merge-use-auto-merge.erg

### DIFF
--- a/rules/git.md
+++ b/rules/git.md
@@ -1,4 +1,4 @@
-<!-- last-reviewed: 2026-06-03 -->
+<!-- last-reviewed: 2026-06-04 -->
 # Git Discipline
 
 - **Always work on a branch.** Main is read-only — no exceptions. Everything (code, docs, tickets, STATE, memory, config) lands via branch + PR. See `rules/workflow.md` § Worktree paths.
@@ -11,8 +11,6 @@
 - **Rebase before merge.** Before merging any PR, rebase onto current `origin/main`, push `--force-with-lease`, and wait for CI to pass. This prevents "Base branch was modified" failures mid-merge and keeps history linear.
 - **After an APPROVED `/verify`, sync before merging.** `/verify` may commit and push fixes from its own review worktree, leaving your local branch behind `origin`. Run `git fetch origin && git merge --ff-only origin/<branch>` before you rebase/merge — otherwise a rebase + `--force-with-lease` silently drops the verify fix.
 - **Merge-bounce recovery.** `erg-pr-merge` legitimately bounces in sequence; each bounce has a specific retry — do not blanket-fall back to `gh pr merge`:
-  - *Mergeability `UNKNOWN` / "not mergeable" right after a `--force-with-lease` push or the ticket-close commit* — GitHub is recomputing mergeability/checks. Wait for the live check (`gh run watch <run-id>`), then re-invoke `erg-pr-merge`.
-  - *"CI has N check(s) still running"* — `gh run watch` the pending run to completion, then re-invoke.
   - *"close: no ticket found" on a retry* — the FIRST run already `erg close`d + archived + pushed the close commit (the script is **not** idempotent past that step). Do NOT re-run it and do NOT hand-close the ticket; the close commit is already on the branch, so finish with `gh pr merge <N> --merge` directly once CI is green.
   - *"must run from PR branch" after a fast-forward* — HEAD detached after `merge --ff-only`; `git checkout <branch>`, then retry.
 - **Delete branches after merge.** All repos use `deleteBranchOnMerge: true` on GitHub, so the remote branch disappears automatically when a PR merges. Clean up stale local branches with:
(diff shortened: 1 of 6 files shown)


MinhHaDuong:
## /verify actions

round: 2
adherence: PASS — 0 blocking (round-1 agnostic-guard failure fixed)
review-pr: skipped (focused 110-line diff, fully read inline)
simplify: skipped (adherence blocking in round 1)
fix agent: 1 commit (8ab0db4 — restored two dropped harness-extension-point hatches)

## /verify-gate verdict

verdict: APPROVED

Round 1 found a blocking adherence violation: the auto-merge rewrite reworded the
step-3 comment block and reshuffled the fallback CI-watch block, dropping/misplacing the
two `harness-extension-point` escape-hatch annotations. `agnostic-guard` (a required
check) went red on this branch while green on `origin/main`. Notable irony: a PR that
makes merge depend on the required-checks ruleset failed a required check — and with
`--auto` queued it would have sat unmergeable forever. The fix commit reworded the
step-3 comment to drop the literal `gh ` token and moved the fallback hatch onto the
preceding line so the guard's same-or-preceding-line rule matches.

Exit criteria:
- erg-pr-merge exits immediately after queueing auto-merge: ADDRESSED — `if gh pr merge ... --auto` sets AUTO_QUEUED=1, prints "Merge queued", Step 4 leaves base untouched.
- Fallback to watch-then-merge when auto-merge unavailable: ADDRESSED — `else` branch runs `timeout 600 gh pr checks --watch` then plain `--merge`; covered by test case 5 (STUB_AUTO_FAILS=1).
- tests cover both paths: ADDRESSED — cases 4 (happy), 5 (fallback), 6 (push-before-merge / git-erg#256 strand) added; suite 6/6 PASS.
- rules/git.md merge-bounce section pruned: ADDRESSED — the two CI-race bullets (UNKNOWN, "N checks running") removed; idempotency + detached-HEAD bullets kept.

Unresolved simplify: none
Adherence: PASS (all 7 required checks green on 8ab0db4)

Rationale:
The implementation is sound: AUTO_QUEUED is set/consumed consistently, the Step-4
three-way branch (worktree / auto-queued / fallback-ff) is correct, and test case 6
genuinely asserts the unconditional push via an origin-ancestor check. The lone blocking
defect was the dropped escape hatches, now fixed in round 2 with all required checks
green. The PR-body caveat about `test_pretooluse_worktree_path_guard.sh` is a
pre-existing, environment-specific failure byte-identical to main and outside the
required-checks set — not blocking.

telemetry: wall=~900s agents=1 (advisor) tokens=na cost~=na

PR #277, opened 2026-06-04
fix(0204): repo-relative script paths in skills; lint guard

MinhHaDuong:
## Summary

- **Fix 3 repo-relative script references** in celebrate, end-session, and housekeeping SKILL.md files — prefixed with `~/.claude/` so they resolve when skills run against non-harness projects.
- **Document is-ancestor discard-safety rule** in celebrate step 9b — ExitWorktree's "N commits would be discarded" warning is a false alarm from a stale local main when the merge pre-check already passed.
- **Renumber celebrate steps** 9.5/10/11 to 10/11/12 — the fractional step was a lazy insert, not a genuine subdivision.
- **Add lint guard** to `check-agnostic.sh` catching bare `scripts/*.{sh,py}` references in SKILL.md files (BRE pattern in SKILL_PATTERNS array).
- **Escape-hatch** 3 nightbeat-supervisor lines where `scripts/beat.py` legitimately appears in `$HARNESS_DIR`-scoped git-add commands.

## Test plan

- [x] `bash scripts/check-agnostic.sh skills` passes with zero violations
- [x] Fixture test: bare `scripts/my-tool.sh` in a skills/ dir is caught by the new pattern
- [x] `make check` passes
- [x] Exit criteria grep returns only `$HARNESS_DIR`-scoped or escape-hatched lines
- [x] `grep -n '^[0-9]' skills/celebrate/SKILL.md` shows steps 1-12 with no fractional steps

**Ticket:** tickets/0204-skills-reference-harness-scripts-repo-re.erg

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Scope overflow: tickets/0213 (regression-pin fixture for the new lint pattern — bundled on this branch)

PR #280, opened 2026-06-04
skills: fix worktree-exit-preflight.sh path (celebrate + end-session)

MinhHaDuong:
Both skills invoked the exit preflight as `scripts/worktree-exit-preflight.sh` — project-relative, so it 404s in every project repo (exit 127 observed during AEDIST /celebrate, 2026-06-04). The script lives in `~/.claude/scripts/`; siblings `worktree-gc.sh` and `worktree-salvage.sh` were already referenced with the harness-global path.

Swept `skills/` and `rules/` for the pattern — these two lines were the only offenders.

Ticket: none

### DIFF
--- a/skills/celebrate/SKILL.md
+++ b/skills/celebrate/SKILL.md
@@ -44,7 +44,7 @@ If the current branch is not merged into origin/main, stop and tell the user. Do
 9. **Exit worktree** (if in one):
     a. Preflight from inside the worktree:
        ```bash
-       scripts/worktree-exit-preflight.sh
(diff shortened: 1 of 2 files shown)


MinhHaDuong:
Superseded by #277 (ticket 0204), which fixes the same two preflight lines plus housekeeping/nightbeat-supervisor and adds the lint guard + regression fixture (0213). Closing — independent confirmation that the bug bites in the field: exit 127 observed during an AEDIST /celebrate on 2026-06-04.

PR #315, opened 2026-06-06
fix(0224): dream v2 — decay-confirmation path and provenance write race

MinhHaDuong:
Closes the two non-blocking residuals from the Dream v2 verify gate (PR #302, ticket 0165).

## Defect 1 — decay-confirmation loop half-wired
Once an entry is promoted, its project-level copy is tombstoned, so the step-7 `record` call that would refresh `last_confirmed` never fires again — every promoted harness entry decay-flags at 90 days regardless of continued relevance.

Fix: new `confirm <slug>` verb in `provenance.py` refreshes `last_confirmed` on a promoted entry (no project-list mutation; errors on unknown/unpromoted slug like `promote`). Wired into `SKILL.md` as step 7b — a consolidation that still finds a promoted lesson supported by the project's content confirms it; unsupported entries are left to decay-flag (the intended human-review signal).

## Defect 2 — lost-update race on `.provenance.json`
`record()`/`promote()` were unlocked read-modify-write; the cron fires all projects at 02:00, so concurrent consolidations could clobber provenance and suppress a promotion. Serialized the full RMW under `fcntl.flock` on a sidecar `.provenance.lock` (auto-releases on exit, protects interactive runs a staggered cron would not, and avoids the truncation-vs-lock conflict of locking the json itself).

## Tests
- Confirm path: refresh makes decay skip a >90d entry; rejects unknown/unpromoted slugs.
- Race path is a genuine teeth-test — a `DREAM_PROVENANCE_TEST_DELAY` hook forces the two critical sections to overlap, so the test FAILS without the lock (verified by disabling it) and passes with it. Marked `integration`.
- Added `pytest.ini` registering the `integration`/`slow` markers (none existed) to retire the unknown-marker warning.

All 430 repo tests pass; `make check` and `ruff` clean.

**Ticket:** tickets/0224-dream-v2-debt-decay-confirmation-loop-an.erg
Related follow-up: tickets/0225-dream-v2-atomic-provenance-write-to-prev.erg (torn-read under concurrency — distinct failure class, surfaced by /gaze)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #318, opened 2026-06-06
fix(dream): atomic provenance write to prevent torn reads (0225)

MinhHaDuong:
## Summary

Closes the torn-read gap left open by PR #315 (ticket 0224). That PR serialized provenance WRITERS under an `fcntl.flock` but kept READERS lock-free; `_save_provenance` still used `Path.write_text`, which truncates then writes. A reader (`candidates`/`decay`/`show`/`_load_provenance`) landing between truncate and content-flush sees a partial file and raises `JSONDecodeError`. Under the 02:00 cron fan-out this is a real, if narrow, window.

## Fix

Make the write atomic rather than locking the read path: stage the full document in a temp file in the same directory (`tempfile.mkstemp`), then `os.replace()` onto the live path. The rename is atomic, so a reader always observes the complete old-or-new document — no tear — without serializing readers against writers. This is the cheaper of the two options the ticket weighed; the read-lock alternative would extend the critical section for no added safety. No `fsync`: tear-prevention comes from the atomic rename, not durability.

## Test (teeth verified)

The existing `DREAM_PROVENANCE_TEST_DELAY` hook fires *between* load and save — before the write begins, when the on-disk file is still the intact old document — so it cannot create a torn-read window. Added a sibling `DREAM_PROVENANCE_WRITE_DELAY` hook firing *inside* `_save_provenance` at the safe point (temp staged, live file intact, before `os.replace`). `test_provenance_read_during_write_never_torn` interleaves a reader into that window and asserts the reader loads a valid document with the pre-existing entry intact.

Teeth check (exit criterion 3): reverting `_save_provenance` to an instrumented non-atomic `open(...,"w")` form turns the test RED — the reader subprocess exits nonzero on `json.loads("")` (`JSONDecodeError: Expecting value, char 0`). The atomic form is GREEN. 445 tests pass; `make check` passes.

## Exit criteria

- [x] Provenance writes are atomic (temp + `os.replace`)
- [x] A reader interleaved with a writer never observes a torn file
- [x] Test covers the path and fails against non-atomic write; `make check` passes

**Ticket:** tickets/0225-dream-v2-atomic-provenance-write-to-prev.erg

### DIFF
--- a/skills/dream/provenance.py
+++ b/skills/dream/provenance.py
@@ -14,6 +14,7 @@ import fcntl
 import json
 import os
 import sys
+import tempfile
 import time
 from contextlib import contextmanager
 from datetime import datetime, timezone
@@ -26,9 +27,17 @@ PROJECTS_BASE = Path.home() / ".claude" / "projects"
 DECAY_DAYS = 90
 
 # Test-only hook: seconds to sleep between read and write inside a locked
-# mutation, used to force critical-section overlap in the race test. Unset in
-# production. See tests/test_dream.py::test_provenance_concurrent_record_*.
+# mutation, used to force critical-section overlap in the lost-write race test.
+# Unset in production. See tests/test_dream.py::test_provenance_concurrent_record_*.
 _TEST_DELAY_ENV = "DREAM_PROVENANCE_TEST_DELAY"
+
+# Test-only hook: seconds to sleep mid-write — after the full content is staged
+# in the temp file but before os.replace publishes it. Lets the torn-read test
+# land a reader inside the write window. In the atomic implementation the live
+# file is still the intact old document here, so a reader sees old-or-new but
+# never a tear. Unset in production.
+# See tests/test_dream.py::test_provenance_read_during_write_never_torn.
+_WRITE_DELAY_ENV = "DREAM_PROVENANCE_WRITE_DELAY"
 
 
 @contextmanager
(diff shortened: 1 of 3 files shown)


MinhHaDuong (commented):
## /review-pr synthesis — approve-tier

Focused, well-documented concurrency fix. Reviewed correctness, consistency, scope, and red-team perspectives directly (contained 116-line single-function change; harness rules favor not fanning out parallel agents on a fix this small and already teeth-verified).

### Correctness — verified
- **Atomicity is sound.** Stage in a same-directory temp via `tempfile.mkstemp` → `os.replace`. `os.replace` is atomic within a filesystem, and the temp lives in `PROVENANCE_PATH.parent`, so the rename never crosses a filesystem boundary. A lock-free reader sees the complete old-or-new document. Correctly matches the cheaper of the two options the ticket weighed; no read-path lock added (right call).
- **Teeth independently reproduced.** Reverted `_save_provenance` to the instrumented non-atomic `open(...,"w")` form from the test docstring → `test_provenance_read_during_write_never_torn` goes RED with exactly `JSONDecodeError: Expecting value: line 1 column 1 (char 0)` (reader subprocess exit 1). Atomic form GREEN. The test catches the torn read, not merely an implementation swap. `make check` green (22/22 dream tests, full suite exit 0).
- The implementation note correctly diagnoses why the existing `DREAM_PROVENANCE_TEST_DELAY` hook can't create a torn-read window (it fires pre-write, live file still intact) and adds the correctly-sited `DREAM_PROVENANCE_WRITE_DELAY`.

### Consistency
- New `_WRITE_DELAY_ENV` / `_write_delay()` mirror the existing `_TEST_DELAY_ENV` / `_test_delay()` pair in naming, docstring style, and placement. Comment on `_TEST_DELAY_ENV` was sharpened to "lost-write race" to disambiguate from the new torn-read hook — good.

### Findings (non-blocking)

- **consider:** First atomic write silently changes the file mode. `mkstemp` creates the temp at `0o600` and `os.replace` preserves the source inode's permissions, so `.provenance.json` flips from `664` → `600` on first write. Verified empirically (`stat -c %a` before/after a `record` on a `write_text`-seeded file, umask `0002`). For a private single-user file under `~/.claude/memory` this is fine-or-better; flagging only because it's the one behavior change in the diff not covered by a test.
- **consider:** A SIGKILL between `mkstemp` and `os.replace` leaves an orphan `.provenance.*.tmp` (the `except BaseException` only covers in-process failures). Harmless — readers use the fixed live path, so orphans are never read — but it's litter under repeated crashes. No action needed.

### Scope
Changes are confined to `_save_provenance` + its test hook + one test; ticket exit criteria all met. No scope creep.

**Verdict: approve.** No blockers. Both notes are author's-discretion.

MinhHaDuong:
Exit criteria: 3/3  Review: 0 unresolved  Simplify: 0 unresolved
Adherence: 0 blocking  Scope overflow: 0 (0/0)
Minors: verifiable:0 consider:2 nofollow:0 malformed:0

Rationale: All three exit criteria addressed with first-hand evidence. Atomic write (tempfile.mkstemp + os.replace) confirmed in skills/dream/provenance.py:83,90 (commit 057440f). Test teeth independently re-verified: test_provenance_read_during_write_never_torn is GREEN on the atomic form and RED on the reverted non-atomic open(...,"w") form (reader subprocess exits 1 on JSONDecodeError: Expecting value, char 0) — it catches the torn read, not merely an implementation swap. Full tests/test_dream.py module re-run first-hand: 22/22 pass (shared DREAM_PROVENANCE_TEST_DELAY-hook siblings unregressed). ruff clean. Scope fully contained: only _save_provenance, its sibling write-delay hook, and one new test. The sole /review-pr review is approve-tier with no blockers; both findings (664→600 mode flip, orphan tmp on SIGKILL) are correctly tagged consider: — informational, author's discretion.

MinhHaDuong:
## /gaze actions

round: 1
adherence: PASS — 0 blocking
review-pr: approve, 2 consider-tier notes (non-blocking)
simplify: 0 fixes applied (code already clean)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Provenance writes are atomic (temp + os.replace): ADDRESSED — `_save_provenance` stages via `tempfile.mkstemp` in the same dir, then `os.replace`
- Reader interleaved with writer never observes a torn file: ADDRESSED — `test_provenance_read_during_write_never_torn` lands a reader inside the `DREAM_PROVENANCE_WRITE_DELAY` window; asserts valid doc + pre-existing entry intact
- Test fails against non-atomic write; make check passes: ADDRESSED — teeth re-verified first-hand (RED on reverted `open(...,"w")` with `JSONDecodeError` char 0; GREEN on atomic), 22/22 dream tests pass, ruff clean

Unresolved simplify: none
Adherence: PASS

Rationale:
The teeth claim is independently reproduced: reverting `_save_provenance` to the instrumented non-atomic form turns the torn-read test RED with the exact JSONDecodeError a real tear produces, GREEN on the atomic form. Atomicity is sound — same-directory temp keeps `os.replace` within one filesystem, so a lock-free reader always sees a complete old-or-new document. Scope is confined to the write primitive plus its sibling test hook and one test. Both review notes (664→600 mode flip on first atomic write; orphan .tmp on SIGKILL) are correctly informational. The /gaze caller decides on merge.

telemetry: agents=3 tokens=na cost~=na

PR #366, opened 2026-06-10
make check runs the full pytest suite (ticket 0238)

MinhHaDuong:
## Summary
`rules/coding-python.md` defines `make check` as the full suite (integration + slow included), but the target ran only the static drift/agnostic checks — zero pytest. Sessions using it as the pre-PR gate silently skipped ~500 tests.

- Add a `check-tests` prerequisite running bare `python3 -m pytest tests/`; `check` keeps its three static prerequisites.
- Add `tests/test_makefile_gates.py` (source inspection, unmarked → runs in `check-fast` too) so a future regression of either gate definition turns the suite red. Written red-first against the old Makefile.

## Verification
- `make check`: static checks + 504 passed.
- `make check-fast`: 485 passed, 19 deselected.
- `/verify-adherence t238-make-check-pytest`: PASS (one nit on the test heuristic, fixed in the second commit; gate re-run green).

**Ticket:** tickets/0238-make-check-runs-no-tests-contradicts-cod.erg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: skills-catalog check-skills-drift check-agnostic-tickets check-agnostic-skills check check-fast
+.PHONY: skills-catalog check-skills-drift check-agnostic-tickets check-agnostic-skills check check-fast check-tests
 
 skills-catalog:
 	./scripts/update-skills-catalog.py
@@ -18,4 +18,8 @@ check-agnostic-tickets:
 check-agnostic-skills:
 	./scripts/check-agnostic.sh skills
 
-check: check-skills-drift check-agnostic-tickets check-agnostic-skills
+# Full gate (coding-python.md): the whole suite, integration + slow included.
+check-tests:
+	python3 -m pytest tests/
+
+check: check-skills-drift check-agnostic-tickets check-agnostic-skills check-tests
--- a/tests/test_makefile_gates.py
+++ b/tests/test_makefile_gates.py
@@ -1,0 +1,45 @@
+"""Hygiene: Makefile gate targets match the coding-python.md definitions.
+
+`check` is the pre-PR gate: full pytest suite (integration + slow included)
+plus the static checks. `check-fast` is the development gate: unit tests only.
+Source inspection, not subprocess (ticket 0238).
+"""
+
+import re
+from pathlib import Path
+
+MAKEFILE = Path(__file__).parent.parent / "Makefile"
+
+
+def target_recipe(name: str) -> str:
+    """Return the full rule (deps line + recipe) for a Makefile target."""
+    text = MAKEFILE.read_text()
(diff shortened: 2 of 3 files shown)


MinhHaDuong:
## /review-pr — multi-perspective review

**Verdict: approve** (with comments)

### Correctness — high confidence
- Exit criteria met: `check` now depends on `check-tests` (`python3 -m pytest tests/`, no marker filter) and keeps all three static prerequisites; full suite passes (504 passed, 9.5s) under a sane env.
- The hygiene test was mutation-verified during review: dropping `check-tests` from the `check` deps → RED; adding `-m "not slow"` to the recipe → RED; restored state → GREEN (3 passed). The ratchet has teeth.

### Red team — high confidence
- verifiable: on this machine `make check` (and pre-existing `make check-fast`) abort with `Erreur 4`: `__main__.py: error: unrecognized arguments: --cache-dir=/data/cache/pytest`. Cause: `/etc/profile.d/dev-cache.sh` exports an invalid `PYTEST_ADDOPTS` (known issue, recorded in memory). Pre-existing and machine-level, not introduced here — the PR merely makes it loud where the old no-pytest `check` masked it. Fail-closed is the right behavior; no change requested. Reproduce: `make check` in any login-shell-derived env; clears with `PYTEST_ADDOPTS= make check`.
- consider: ticket 0238's body carries the reminder to file the `dev-cache.sh` fix in its home repo; merging this PR closes 0238 and orphans that reminder. Worth filing the follow-up ticket at the destination before or at merge.

### Consistency — high confidence
- `python3 -m pytest` matches the existing `check-fast` recipe style; `.PHONY` updated; docstring cites the rule file and ticket. Fits the codebase.

### Scope — high confidence
- Diff touches exactly the two files the ticket names. No drift.

### Doc propagation — high confidence
- `rules/coding-python.md` gate definitions now match the Makefile (that was the point). No other doc references `make check` semantics in this repo.

### Minor
- consider: `test_check_runs_full_pytest_suite` hard-codes the `check-tests` name for the dep redirect; renaming that target (behavior-preserving) would false-fail the test. Dismissible — the failure message points straight at the rule.

MinhHaDuong:
Review comments addressed:

- **Orphaned dev-cache.sh reminder**: filed at the destination — padme ticket 0043, landing via MinhHaDuong/padme#67. Closing IDH 0238 no longer loses the follow-up. This also covers the red-team note: `make check` errors loudly on this host until that env fix lands (fail-closed, by design).
- **Hard-coded `check-tests` name (consider, dismissible)**: intentional. If the target is renamed, the test falls back to inspecting `check` directly, finds no bare pytest recipe, and fails — which forces the renamer to update the gate test consciously. For a ratchet test, a false-fail on rename is the desired behavior; tolerating arbitrary delegation names would weaken exactly the regression it guards.

PR #475, opened 2026-07-11
eager local-main sync: rule + safe mechanism (0276)

MinhHaDuong:
## What

Promotes the sync-main lesson to a rule and makes local-main sync eager and automatic, per the author's request ("tired of unsynced local main").

1. **`scripts/sync-local-main.sh`** — safe sync callable from any checkout or worktree: moves only the default branch, ff-only; by ref (`fetch main:main`) when nothing has it checked out, ff-merge in the one worktree that does; reports instead of touching dirty-overlap or diverged state; always exits 0 (hook-safe).
2. **`scripts/on-start.sh`** — calls it backgrounded (timeout 60) in the silent section, so every session start refreshes main and an offline fetch can't stall startup.
3. **`skills/merge/SKILL.md`** — after the merge is confirmed landed, run the sync script; forbids the hand-rolled `merge --ff-only` on the primary checkout.
4. **`rules/git.md`** — new bullet "Local main syncs eagerly, by ref — never by assumption", naming the 2026-07-11 failure mode (the primary checkout was on another session's branch; the sync advanced it).
5. Memory entry updated to point at the rule as authoritative.

## Verification

- [x] `tests/test_sync_local_main.sh` — 8 cases green, covering the untouched-guarantees (dirty overlap preserved, diverged main not moved, feature branch untouched) as strictly as the moves; auto-discovered by test_bash_suites
- [x] `make check-fast` — 616 passed
- [x] `make check-skills-drift` OK; `tests/test_erg_pr_merge.sh` green after the SKILL.md addition
- [x] `erg validate` PASS on 0276

**Ticket:** tickets/0276-eager-local-main-sync-rule-mechanism.erg


MinhHaDuong:
## Multi-perspective review — PR #475 (eager local-main sync)

Ran the four requested perspectives against `scripts/sync-local-main.sh`, `scripts/on-start.sh`, `skills/merge/SKILL.md`, `rules/git.md`, and `tests/test_sync_local_main.sh`. Test suite is green (8/8 cases). No blockers found. Summary below; findings tagged `verifiable:` (reproducible evidence attached), `consider:` (hypothesis, no repro), `nofollow:` (checked, disproven or out of scope).

### 1. CORRECTNESS — approve (high confidence)

Deep-read the script plus targeted empirical stress tests:
- Raced `git fetch origin main:main` (ref-update path) against a concurrent `git worktree add -b feat main` 20×: no corruption, clean atomic serialization each time.
- Confirmed git's own worktree-checkout protection refuses to fetch-update a branch checked out in *any* worktree, not just the invoking one — a defense-in-depth backstop beyond the script's own `co_path` detection.
- Verified default-branch resolution against a repo whose default is `trunk` (not main/master) and against a clone with no `origin/HEAD` set — both resolve correctly.
- Mutation-tested the test suite: an off-by-one in the `co_path` awk parser is caught (goes RED); a dropped `--ff-only` is masked by the earlier `merge-base --is-ancestor` gate (see minor below — not a real safety gap, just redundant layering).

Minors:
- `verifiable:` Two sessions syncing the same shared primary checkout concurrently can race the `git -C "$co_path" merge --ff-only` step; one hits `.git/index.lock` contention and gets misreported as "conflicting local state — left untouched" (`scripts/sync-local-main.sh` ~L59-63). Repro: run two `git merge --ff-only --quiet origin/main` in parallel against the same checkout → second exits 1 with `Unable to create '.git/index.lock'`. No data loss (self-heals next sync), just a misleading diagnostic under this specific race.
- `verifiable:` `tests/test_sync_local_main.sh` case 6 ("silent when in sync") only captures stdout via `out=$(...)`; dropping `--quiet` from the primary `git fetch origin "$default"` call (L32) produces stderr noise that this assertion does not catch. Reproduced via mutation. Not a bug in the shipped script (it does have `--quiet`), but a gap in what the test can discriminate — worth a stderr-inclusive assertion (`out=$(... 2>&1)`) if regression coverage on quietness matters.
- `consider:` `local_sha=$(git rev-parse "refs/heads/$default")` (and `remote_sha=...`) is unguarded under `set -e`; a TOCTOU where the ref vanishes between the preceding `show-ref --verify` and this line would exit the script non-zero, contradicting the documented "always exit 0" hook-safety guarantee. Narrow window, not reproduced.
- `nofollow:` Suspected `timeout 60` might not kill the `git fetch` grandchild on expiry, leaving a lingering network fetch. Empirically disproven — `timeout` cleans up the full descendant chain in this environment, confirmed with a sleep-based repro.

### 2. CONSISTENCY — approve (high confidence)

No contradiction with the existing "sync before starting work" (`rules/workflow.md`, operates on feature-branch-vs-origin diffing, informational only) or "After an APPROVED `/gaze`, sync before merging" (operates on the *feature* branch ref, not main) bullets — different scope, no overlap. `skills/merge/SKILL.md`'s new "After the merge lands" step is additive and doesn't conflict with `erg-pr-merge`.

- `verifiable:` `projects/-home-haduong--claude/memory/feedback_primary_checkout_staleness_gates_skills.md` (still linked from the injected `MEMORY.md` index, line 47) still instructs `git -C ~/.claude merge --ff-only origin/main` as the fix for stale skills, in both its body and "How to apply" section — exactly the hand-rolled idiom the new `rules/git.md` bullet calls out as unsafe and supersedes. This is the closest sibling doc to the incident this PR fixes and wasn't updated.

### 3. RED-TEAM — approve (medium-high confidence)

Attacked the stated invariants directly (see CORRECTNESS stress tests above: 20-iteration concurrent race, cross-worktree fetch-refusal check). Could not make the script move a non-default branch, lose work, or corrupt state under concurrency. The lock-contention finding above is the only adversarial win, and it degrades to a mislabeled no-op, not data loss. A worktree mid-rebase/mid-merge causes the `merge --ff-only` to fail safely (git itself refuses), landing in the same "left untouched" branch.

### 4. DOC-PROPAGATION — comment (medium confidence)

`grep -rn "merge --ff-only origin/main"` across the repo returns only `rules/git.md`'s own (intentional) description of the incident — no other hand-rolled idiom survives in rules/skills text, except the memory entry above.

- `consider:` `README.md`'s `scripts/` directory tree (~L33-40) doesn't list the new `scripts/sync-local-main.sh`, and the `on-start.sh` one-line description isn't updated to mention the new sync call. Low priority: that tree is already stale for several pre-existing files (e.g. `rules/` subdirectories), so this isn't a regression introduced by the PR, just an opportunity.

### Verdict

No untagged blockers from any perspective. Recommend merge; the two `verifiable:` minors (lock-contention message accuracy, stderr-blind silence test) and the stale memory-entry doc gap are worth a fast-follow, not a hold.

MinhHaDuong:
## /gaze actions

round: 2
adherence: PASS — 0 blocking (nits: unguarded worktree-list pipe; on-start wiring untested directly)
review-pr: panel comment posted — 0 blockers, 3 verifiable / 2 consider / 1 nofollow minors
simplify: 2 fixes applied (commit 0661cae — local ref-fetch instead of second network round-trip; test case 5 now covers the cross-worktree ff path); 7 skips with rationale (test-helper refactor contradicts repo convention; beat.py `_sync_origin_main` duplication and 3 altitude findings are follow-up material outside this diff; 2 micro-efficiency items negligible)
fix agent: 1 commit (ed3bdf4 — round-1 reroll: exit-0 cd guard + test case 9, honest ff-failure wording, stderr-aware case 6, retired stale idiom in the memory entry)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Rule bullet in rules/git.md; script + call sites merged: ADDRESSED — rules/git.md:31 (d92161f); scripts/on-start.sh:54 backgrounded call; skills/merge/SKILL.md:44-50 post-merge step
- Local main refreshed at session start and post-merge without manual intervention: ADDRESSED — both call sites wired; script behavior proven by tests/test_sync_local_main.sh 9/9 (run live by the gate)
- Test suite green: ADDRESSED — suite 9/9; `make check-fast` 616 passed; guards green in CI

Round-1 items (all verified fixed in ed3bdf4):
- B1 unguarded `cd` vs exit-0 contract: FIXED — live run on a nonexistent path → exit 0 + "skipped"; new test case 9
- C1 ff-failure misdiagnosis wording: FIXED — states observation, keeps "left untouched"
- C2 stderr-blind silent-case test: FIXED — case 6 captures 2>&1 (mutation-proven RED/GREEN)
- C3 memory entry still prescribing `merge --ff-only origin/main`: FIXED — How-to-apply now points at scripts/sync-local-main.sh; remaining mentions are historical narrative only

Unresolved simplify: none
Adherence: PASS (non-blocking nits listed above; B2 nonstandard-refspec edge and B3 ahead-vs-diverged wording dispositioned informational)
Scope overflow: none — every touched file traces to a ticket action

Rationale:
Round 1 bounced on a CONFIRMED contract violation (script promised exit-0, exited 1 on a bad path) plus three verifiable minors; the fix commit addressed all four with RED/GREEN evidence, and round 2 re-verified each directly rather than trusting the claim. Follow-up material (not blocking, recommend a ticket): wire the sync mechanically into erg-pr-merge step 4 (its AUTO_QUEUED path currently just prints "sync after the PR lands"), consider a branch argument for non-default merge bases, and surface the script's report from the session-start call site (currently discarded past `exec >/dev/null`); beat.py `_sync_origin_main` duplicates this logic and could delegate to the script.

/gaze: slow run — telemetry over warn thresholds (wall > 15 min, tokens > 500k).
telemetry: wall~=1300s agents=10 tokens=717525 cost~=na

PR #378, opened 2026-06-10
trace-doctor phase 4: counterfactual accounting — bounds settled, routing fixed

MinhHaDuong:
Executes ticket 0244 (phase 4 — counterfactual accounting), with the H10-sharpened spec and all six skeptical-review constraints applied.

**Implementation** (2 commits): disjoint per-turn cache_read buckets in trace-stats.py (tail-after-LAST-marker > post-2nd-verify/gaze > nav/idle-excluding-final > core) so the addressable total is an exact sum; compute_refined() in trace-hypotheses.py (relocation premium vs 15K baseline, per-family pricing, reflexivity sensitivity, H2 segmentation); trace-pr-join.py forge join with committed offline cache. 19 new tests, red-first. make check exit 0.

**Headlines**: the author-flagged H10 first-marker confound was real (~4x: $968 → $247 tail, $57/wk premium — fresh-session relocation REJECTED); micro-turns settle at $268/wk exact (adopt prompt-rule); verification re-entry $259/wk and NOT explained by diff size (re-entry median diff 465 vs 597 — A/B); H12 dissolves ($0.7/wk); H5 shrinks to $32/wk (rejected); dedup addressable total $2,356 = 34.8% of spend; 112/112 joined PRs merged (5 unresolved); REROLL 33 / ESCALATE 18 are the operative guardrail. Phase-5 shortlist: model right-sizing + verification convergence.

**Constraints honored**: shadow-dollar header, exact dedup, reflexivity sensitivity (2 study sessions, negligible this window), H2 segmented (no Simpson artifact), H13 difficulty-conditioned, zero open-coding prose cited.

**Ticket:** tickets/0244-trace-doctor-phase-4-counterfactual-acco.erg
Ticket-ref: tickets/0236-trace-doctor-session-trace-economics-aud.erg

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #494, opened 2026-07-11
fix(0246): forge-join blank-row retry, mandated-regex anchor, report nit

MinhHaDuong:
Post-merge review of PR #378 (ticket 0244, trace-doctor phase 4) found four
defects; action 4 already landed via PR #380. This closes out actions 1-3.

## Defect 1 — `build_join` never re-fetched blank cached rows
`scripts/trace-pr-join.py` skipped every pair already keyed in the cache, so a
row left blank by a prior failed fetch (`merged` in `None`/`""`) stayed blank
forever. Now only *resolved* rows are skipped; blank rows are re-fetched and
overwritten. **The 5 currently-stuck blank CSV rows (`-claude-projects` 67/366,
`git-erg` 309, `padme` 321/322) are left as-is: they resolve on the author's
next networked join run, which is the intended design** — this PR does not
re-run the networked join and does not touch `docs/trace-pr-join-2026-06.csv`.

## Defect 2 — mandated-regex anchor was asymmetric
`MANDATED_BASH_RE` bound the `^\s*` anchor only to the first alternative, so
`cd x && git worktree prune` did NOT count as mandated while `cd x && erg close`
DID. Replaced the line anchor with word boundaries on both alternatives;
compound commands now count as mandated.

## Defect 3 — stale joined-session count in the report
`docs/trace-counterfactuals-2026-06.md`: the final "what accounting could not
settle" line still said `34`; every other section says `33`. Commit c2b9f0f
fixed two of three occurrences and missed this one. Corrected to `33` (the
recomputed post-fix value from trace-hypotheses H4r).

## Tests (RED-first)
- `test_build_join_refetches_blank_cached_rows` — cache with one resolved + one
  blank pair, stubbed fetcher; asserts only the blank pair is re-fetched and its
  row overwritten, resolved row untouched.
- `test_mandated_regex_matches_compound_commands` — `cd x && git worktree prune`
  is now True (was False), plus regression on `erg close`, leading
  git-branch/fetch, and a non-mandated command.

Both were confirmed failing before the fix; `make check` passes.

**Ticket:** tickets/0246-review-378-follow-up-forge-join-blank-ro.erg

### DIFF
--- a/docs/trace-counterfactuals-2026-06.md
+++ b/docs/trace-counterfactuals-2026-06.md
@@ -81,5 +81,5 @@ the operative guardrail metrics (merged-rate is saturated at 100%).
   needs the A/B, not traces.
 - Whether Sonnet mains deliver gaze-APPROVED PRs at Opus rates — the
   census never observes the counterfactual model.
-- H4 beyond medians — 34 joined sessions is too thin for a regression;
+- H4 beyond medians — 33 joined sessions is too thin for a regression;
   the join cache will accumulate across monthly runs.
--- a/scripts/trace-pr-join.py
+++ b/scripts/trace-pr-join.py
@@ -87,11 +87,15 @@ def gh_fetch_pr(owner: str, repo: str, n: int) -> dict:
 
 def build_join(pairs: list[tuple[str, int]], cache: dict, fetcher) -> dict:
     """Resolve every pair from cache, then fetcher; fetch failures yield a
-    blank row (merged='') so the join never silently drops a PR."""
+    blank row (merged='') so the join never silently drops a PR.
+
+    A cached row left blank by a prior failed fetch (merged in (None, ''))
+    is re-fetched on this run and overwritten; a resolved row is kept as is."""
     out = dict(cache)
     for repo, n in pairs:
-        if (repo, n) in out:
-            continue
+        cached = out.get((repo, n))
+        if cached is not None and cached.get("merged") not in (None, ""):
+            continue  # resolved row — keep, never re-fetch
         try:
             out[(repo, n)] = fetcher(repo, n)
         except Exception as e:  # noqa: BLE001 — any fetch failure is non-fatal
--- a/scripts/trace-stats.py
+++ b/scripts/trace-stats.py
@@ -109,7 +109,9 @@ PR_NUMBER_RE = re.compile(r"(?:PR\s*#|pulls/)(\d+)")
 # Ticket 0244 (H10'): harness-mandated wrap-up work — memory/STATE writes,
 # branch hygiene, wrap-up skills. Used only to classify tail turns.
 MANDATED_PATH_RE = re.compile(r"/memory/|MEMORY\.md|STATE\.md")
-MANDATED_BASH_RE = re.compile(r"\bgit\s+(?:branch|worktree|fetch\s+--prune)|\berg\s+close\b")
+MANDATED_BASH_RE = re.compile(
+    r"(?:^|&&|;|\|)\s*(?:git\s+(?:branch|worktree|fetch\s+--prune)|erg\s+close)\b"
+)
 MANDATED_SKILLS = {"roar", "lair", "dream", "molt", "memory", "housekeeping"}
 PATH_TOOLS = {"Read", "Edit", "Write", "NotebookEdit"}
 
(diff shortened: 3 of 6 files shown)


MinhHaDuong:
## /gaze phase 2-4 review — PR 494

**Scope:** three independent, ticket-scoped fixes, each with a RED-first test:
1. `trace-pr-join.py::build_join` — re-fetches cache rows left blank by a prior failed fetch instead of skipping them forever (resolved rows still skipped).
2. `trace-stats.py::MANDATED_BASH_RE` — widened from a start-anchored alternation to a symmetric `\b`-bounded one, so compound commands (`cd x && git worktree prune`) count as mandated like `erg close` already did.
3. `docs/trace-counterfactuals-2026-06.md` — corrected a stale "34 joined sessions" to "33", matching the other two occurrences in the file.

**Verification:** both new tests (`test_build_join_refetches_blank_cached_rows`, `test_mandated_regex_matches_compound_commands`) pass; full fast tier 651 passed; full suite 683 passed; `build_join`'s blank/resolved discriminator (`merged not in (None, "")`) is sound against how `gh_fetch_pr`/CSV round-tripping stores `merged` (always a real bool, never coincides with the blank sentinel).

**Finding — consider, not blocking:** the regex fix does not just rebalance the anchor between alternatives, it drops command-position anchoring altogether (`^\s*` -> bare `\b`). This is one of the two remediations the ticket itself authorized ("make the anchor symmetric **or drop it**"), but as a side effect `MANDATED_BASH_RE` now matches the keywords anywhere in a Bash command string, not just when the command actually invokes them:

    MANDATED_BASH_RE.search('echo "see: run git branch -D to clean up"')
    # -> matches now; did NOT match under the old anchored regex

(The `erg close` alternative already had this false-positive-on-mention behavior pre-fix -- that half is a pre-existing gap, not a regression. The `git branch|worktree|fetch --prune` alternative gains it new with this change.) Net effect: `tail_mandated_turns`/`mandated_share` may count sessions that merely *mention* running these commands (in an echo, a commit message, a code comment) as if they ran them. Given this script feeds harness-accounting metrics, worth a follow-up ticket to require the match sit at true command position -- not blocking this PR, which correctly implements the ticket's explicit "or drop it" option with tests proving the compound-command case works as specified.

**Other:** `check-agnostic-scripts` currently fails on 2 pre-existing absolute-path violations (`scripts/seat-runner-prototype.sh:89`, `scripts/beat-settings.json:25`) -- untouched by this PR, pre-existing.

No blockers. Recommend APPROVE, with the regex-scope-widening noted above as a `consider:`-tagged follow-up.

MinhHaDuong:
## /gaze actions

round: 1
adherence: PASS — 0 blocking
review-pr: skipped (network-degraded sandbox during Agent C's run — synthesized manually by the orchestrator using Agent C's findings, verified independently)
simplify: 1 fix applied (MANDATED_BASH_RE re-anchored to true command position, commit 7e8212c)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Blank-row retry + symmetric mandated-regex landed with red-first tests: ADDRESSED — c1d5a0f (build_join re-fetches blank cached rows; test_build_join_refetches_blank_cached_rows PASS) + 95cca04 (MANDATED_BASH_RE symmetric; test_mandated_regex_matches_compound_commands PASS)
- Report and 0236 nits fixed: ADDRESSED — 708d290 (33 vs 34 joined-sessions fixed, doc now consistent); 0236's stale placeholder was already removed by prior PR #380 per the ticket's own log
- make check passes: ADDRESSED — check-skills-drift, check-agnostic-tickets, check-agnostic-skills, check-tests all exit 0; 684 passed full suite, 652 passed fast tier

Unresolved simplify findings: none — the one altitude-pass finding (MANDATED_BASH_RE over-widened to match keyword mentions inside quoted/echoed strings, not just true invocations, e.g. `echo "run git branch -D"`) was fixed in commit 7e8212c and independently verified resolved by the gate, with a new regression test (test_mandated_regex_does_not_match_mentions_in_prose) confirming both the negative cases and that the original compound-command fix (test_mandated_regex_matches_compound_commands) is not regressed.
Adherence: PASS (0 blocking) — 1 pre-existing, unrelated nit noted (check-agnostic-scripts fails on 2 violations untouched by this branch)

Rationale:
Three ticket-scoped commits map 1:1 to the three exit criteria with RED-first tests, all independently re-run and confirmed passing. A fourth commit landed during the simplify phase: three independent review passes (this run's built-in /review agent and the simplify altitude pass) converged on the same finding — the mandated-regex fix (95cca04) correctly solved the ticket's stated compound-command bug but, by dropping the `^\s*` anchor entirely rather than re-anchoring after command separators, widened matching to hit mere mentions of the keywords inside quoted or echoed text. This risked silently inflating the tail_mandated_turns/mandated_share metrics this script feeds. The fix was applied within this PR's own scope (same function, same file, ticket-authorized "or drop it" semantics refined to the actually-intended "compound commands that run the keyword" semantics), tested red-first, and re-verified by the gate. No scope overflow, no unresolved comments, no blocking adherence violations. CI green on the final commit (7 checks). PR state OPEN, not draft, mergeStateStatus CLEAN.

telemetry: wall=~35min agents=8 tokens=na cost~=na

PR #496, opened 2026-07-11
CI check to detect ticket-ID collisions across open PRs

MinhHaDuong:
## What

Adds a CI gate against the optimistic-ID trap (git-erg#282, wontfix) *across open PRs*. `erg check` and the `validate-tickets` job run per-branch, so two branches that each add `tickets/0300-*.erg` both pass — the duplicate only surfaces on `main` after the second merges. On 2026-06-18 this produced the three-way 0257/0258 clash and the 0248 double-claim, each renumbered by hand after the fact.

## How

- **`scripts/check-cross-pr-ticket-collision.sh`** — lists the ticket IDs this PR adds (`git diff --diff-filter=A` against the base, top-level `tickets/NNNN-*.erg` only so a close-into-`tickets/closed/` archive is not miscounted as a new claim) and fails if any is also added by another open PR, naming the sibling PR and the next free ID (`./tickets/erg next-id`). Fast path: exits 0 without any forge call when the PR adds no ticket file. Forge coupling is confined to two `# harness-extension-point` `gh` calls (`pr list`, `api .../pulls/N/files`), the isolation idiom from `skills/merge/erg-pr-merge`.
- **`tests/test_check_cross_pr_ticket_collision.sh`** — follows `tests/test_erg_pr_merge.sh`: temp repo, `gh` stub first on PATH dispatching on `$1 $2`, canned `STUB_*` responses. Cases: (a) colliding IDs fail naming the sibling PR, (b) distinct IDs pass, (c) a no-ticket PR exits 0 without invoking `gh` (asserts the stub log is empty). Auto-discovered by `tests/test_bash_suites.py`.
- **`.github/workflows/CI.yml`** — new `pull_request`-only job with `permissions: pull-requests: read` and `GH_TOKEN` (first gh-authenticated job here), `fetch-depth: 0` so `origin/<base>` resolves.
- **`tickets/AGENTS.md`** — documents the gate next to the optimistic-ID note.

## Premise correction

The ticket's action "extend `erg-github verify`" is **void**: `tickets/erg-github` does not exist in this repo (only referenced in CLAUDE.md/AGENTS.md prose). Built as a small standalone script instead. The stale `tickets/erg-github verify` sentence in `tickets/AGENTS.md` is left untouched as pre-existing drift (out of scope — worth a follow-up cleanup).

## Fork-PR token handling

The job requests only `contents: read` + `pull-requests: read` and passes `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}`. It performs no writes, so a fork PR's read-only token is sufficient on public repos; the per-PR `gh api` call is tolerant (`|| true`) so a transient forge error degrades to "no sibling IDs seen" rather than a spurious block — the renumber-on-merge discipline remains the backstop.

## Test evidence

RED (script hidden) → the suite fails; GREEN (restored) → all three cases pass. `make check` passes (exit 0); the new suite is collected under pytest via `test_bash_suites.py`.

**Ticket:** tickets/0262-ci-detect-cross-pr-ticket-id-collision.erg

---

**Fix round (post-review):** the review's blocker — the base-tip collision blind spot (an ID landed via an already-merged claimant is invisible to both the merge-base diff and the open-PR scan) — is closed by checking added IDs against the base tip directly in plain git, same-path hits excluded. Red-first test cases (e)/(f) lock it in.

Related follow-up: tickets/0285-tickets-agents-md-documents-erg-github-v.erg (stale `erg-github verify` sentence in tickets/AGENTS.md).


MinhHaDuong:
## Multi-perspective review — PR #496 (t0262-cross-pr-ticket-collision)

Four parallel perspectives (correctness, consistency, red-team, doc-propagation) reviewed this PR against ticket `tickets/0262-ci-detect-cross-pr-ticket-id-collision.erg`'s exit criteria. `make check` in the review worktree: **682 passed, 1 pre-existing unrelated warning, exit 0** — all three new gh-stub test cases pass.

### Blocker (request-changes)

**[correctness] The script does not actually cover the "already claimed by the base branch" case, and its own inline comment's reasoning for why is wrong.**

`scripts/check-cross-pr-ticket-collision.sh:12-14,62-63` claims: "A ticket added relative to the base branch cannot already exist on the base by construction… so the base-branch collision… is covered for free." This is false. `git diff --diff-filter=A "${BASE_REF}...HEAD"` uses three-dot (merge-base) semantics, so it only reports files added since the branch diverged — it says nothing about whether that ID has since landed on the base tip via an already-merged sibling PR. Reproduced directly in a temp repo: a ticket ID merged onto base by a closed PR, plus an independent open branch adding a *different filename* claiming the same ID, produces `OWN_IDS` that never sees the base's copy, and since the colliding PR is already merged it's absent from `gh pr list --state open` too — the script prints "no open PR claims the same ticket ID(s) — OK" and exits 0 on a genuine base collision.

In practice this specific scenario likely still fails CI overall, because `validate-tickets` (the pre-existing job, `.github/workflows/CI.yml:9-14`) runs `erg check` against `actions/checkout@v4`'s **default** ref for a `pull_request` event — the auto-generated merge commit (`refs/pull/N/merge`), which unifies both files under `tickets/`, and `erg check` does enforce "no duplicate ticket IDs across the corpus" (confirmed via `./tickets/erg --help --all`). But that coverage is: (a) incidental to a *different* job than the one this PR adds, resting on an undocumented default-checkout behavior that silently disappears if anyone ever pins that job's checkout to the head SHA (a common hardening move); and (b) produces a generic `erg check` duplicate-ID failure with no PR name and no next-free-ID suggestion — failing exit criterion (c) ("failure message names the colliding PR and suggests the next free ID") for this specific case. Recommend either fixing the comment to state plainly that base coverage is incidental to `validate-tickets`' default checkout (with a test proving it), or making this script check the base branch's current ticket IDs directly so the actionable message applies uniformly.

### Minor findings

- `consider:` [correctness] `git mv tickets/closed/NNNN-x.erg tickets/NNNN-x.erg` (reopening a closed ticket) is classified `R100` by git, never `A`, so it silently escapes `--diff-filter=A` detection. Reproduced with git 2.53. Narrow edge case, outside the ticket's stated scope.
- `consider:` [consistency] The PR description frames the `# harness-extension-point` usage as "the isolation idiom from `skills/merge/erg-pr-merge`" — true as far as it goes (both `gh` calls are marked), but `erg-pr-merge` applies the marker 16× including flow-control lines, a denser application than this script's 2 marks. A valid subset, not full parity; the framing slightly overstates it.
- `consider:` [consistency] The new script skips `erg-pr-merge`'s `die()` / `"<script>: $*"` prefixed-error convention — `COLLISION:` lines have no script-name prefix, unlike the script's own informational lines (which use `cross-pr-collision:`). Defensible for a shorter script, but a real deviation from the claimed sibling pattern.
- `consider:` [consistency] `tests/test_check_cross_pr_ticket_collision.sh:26` checks `[[ -f "$SCRIPT" ]]`; the sibling `tests/test_erg_pr_merge.sh:19` checks `[[ -x "$SCRIPT" ]]`. Weaker guard than the pattern it claims to mirror — currently harmless since the file is in fact executable.
- `consider:` [red-team] `tickets/erg next-id` (a repo-committed binary) executes from the PR's own fork-modified checkout. Bounded by the job's minimal `contents: read` + `pull-requests: read` permissions and no extra secrets, and mirrors the identical trust pattern the pre-existing `validate-tickets` job already accepts (`./tickets/erg check`) — not a new risk class, worth noting only.
- `consider:` [red-team] At scale (many open PRs), one `gh api --paginate` call per sibling could approach GitHub's per-run token rate limit, degrading toward the fail-open path rather than failing consistently. Low likelihood, not attacker-triggerable by a single contributor.
- `nofollow:` [red-team] The `|| true` fail-open on `gh api` sibling-fetch errors is acceptable given the ticket's own framing (hygiene gate, "renumber-on-merge discipline remains the backstop") and is not exploitable via a single fork PR's own token/rate limit.
- `consider:` [doc-propagation] The pre-existing stale "`tickets/erg-github` verify" sentence in `tickets/AGENTS.md:42` (the helper doesn't exist in this repo) is left as an untracked "worth a follow-up cleanup" note in the PR description rather than a filed ticket. Confirmed no other file (CLAUDE.md, README, other open tickets) references it. Recommend filing a tracked follow-up ticket instead of a bare promise, per this repo's own "loophole found → offer to plug it" norm.

### Confirmed correct / no issue

- Regex `^tickets/[0-9]{4}-.*\.erg$` correctly excludes `tickets/closed/…` and any rename-to-`closed/` (classified `R`, not `A`) — the PR description's claim holds.
- `comm -12` intersection and multi-sibling/multi-ID collision loop are correct; verified via direct test run (all 3 cases PASS, including the gh-stub dispatch on `"$1 $2"` correctly matching both `pr list` and `api repos/.../pulls/N/files` calls).
- No injection risk: attacker-controlled filenames/branch names flow only through quoted `grep`/`sed`/`echo`, never `eval`/`source`/`bash -c`.
- Trigger is `pull_request` (not `pull_request_target`); permissions are read-only (`contents: read`, `pull-requests: read`); no token leakage (no `set -x`).
- Naming (`check-*.sh`), CI.yml job style, test auto-discovery (`tests/test_bash_suites.py` glob), and the `tickets/AGENTS.md` doc placement/tone all match existing repo conventions.

### Verdict: **request-changes**, on the single correctness blocker above. Everything else is minor/informational.

MinhHaDuong:
Exit criteria: 4/4  Review: 0 unresolved  Simplify: 0 unresolved
Minors: verifiable:0 consider:7 nofollow:1 malformed:0
Rationale: The review's single blocker — the base-tip collision blind spot (an ID landed via an already-merged claimant, invisible to both the merge-base diff and the open-PR scan) — is closed by commit 469167c (2026-07-11T22:21Z, after the 21:31Z review): the script now compares added IDs against the base tip directly in plain git (scripts/check-cross-pr-ticket-collision.sh, base-tip block), same-path hits excluded, and the previously wrong inline rationale is rewritten to state the blind spot accurately. Red-first test cases (e)/(f) lock both the detection and the same-path exclusion in; case (a) asserts the failure message names the sibling PR (#77) and suggests the next free ID, covering verification item (c) uniformly across both surfaces. All 8 CI checks pass on the fixed tip. Every minor is tagged (7 consider:, 1 nofollow:, no verifiable:, no untagged); the doc-propagation consider was acted on beyond requirement — ticket 0285 filed (692165f), referenced as Related follow-up (not a close claim), ID verified collision-free against main and open PR #497. Simplify round applied in f47a2fe; no unresolved findings. Scope is contained: every changed file traces to a ticket action or to review-comment disposition.
Nit (non-blocking): branch is 11 commits behind origin/main; mergeStateStatus CLEAN, zero file overlap with upstream — merge as-is is acceptable per git.md when force-push is denied.

PR #497, opened 2026-07-11
ticket(0284): sweep stale pre-reorg home paths outside the nightbeat registry

MinhHaDuong:
Finish the home-reorg stale-path sweep (ticket 0284). The original execute
agent stalled (stream watchdog, 600s); its WIP was salvage-committed, then a
finisher was relaunched to complete the remaining scope.

## What the salvage contained (commit 8bad412)
- `scripts/setup-claude-agent.sh` — PROJECTS array re-pointed to post-reorg locations
- `skills/maw-audit/SKILL.md` — `~/git-erg`→`~/CNRS/code/git-erg` and `~/aedist-technical-report`→`~/CNRS/papiers/actif/AEDIST-technical-report`

## What the finisher added (commit f77435f)
- `STATE.md` — AEDIST maw-audit directive re-pointed to `~/CNRS/papiers/actif/AEDIST-technical-report`
- Ticket log — recorded circuit-breaker relaunch and scry/SKILL.md clean verification

`skills/scry/SKILL.md` verified clean (no pre-reorg paths, no edit needed).
`scripts/check-agnostic.sh` passes; `make check` passes. README.md untouched
(salvage did not regenerate it).

**Ticket:** tickets/0284-home-reorg-stale-paths-sweep.erg

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #493, opened 2026-07-11
nightbeat: survive the 2026-07 home reorg (registry paths + survey guard)

MinhHaDuong:
## Summary

Nightbeat supervisor pre-flight (2026-07-11 run on doudou) found the run's own survey helper dead on arrival:

- **`scripts/projects.json`**: all seven non-harness paths pointed at pre-reorg `~/<name>` locations that no longer exist. Re-pointed each to its new home, identity confirmed via `git remote get-url origin`. The `Climate_finance` label is updated to `climate-finance-het` to match the directory (consumers derive project names from `path.name`, so the label is documentation).
- **`scripts/nightbeat-supervisor-survey.py`**: a missing project path crashed the helper (`FileNotFoundError` from subprocess cwd) instead of degrading. New `_existing_projects()` filter skips absent checkouts with a stderr warning, matching scry's existing behavior. Unit test added.
- **cadens**: the initial survey missed the relocated checkout because its post-reorg directory is capitalized — `~/CNRS/papiers/actif/Cadens`, not the lower-case `cadens` it searched for — and removed the entry. PR #493 verification found the live checkout there (origin `https://github.com/MinhHaDuong/cadens.git`, HEAD 7b55808) and re-registered the entry pointing at it. Ticket 0283 now records the incident.

## Verification

- `make check`: 681 passed, 0 failed.
- End-to-end: the survey helper now runs clean on the corrected registry (exit 0, no warnings) and correctly discovers real open work (climate-finance-het PR #1016).

**Ticket:** tickets/0283-cadens-checkout-relocated-after-home-reorg.erg
Ticket-ref: tickets/0284-home-reorg-stale-paths-sweep.erg

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #500, opened 2026-07-12
setup-claude-agent: read project list from projects.json registry

MinhHaDuong:
## Summary

`scripts/setup-claude-agent.sh` hardcoded a 4-project `PROJECTS` array that
duplicated `scripts/projects.json` — the same 2026-07 home reorg fact had to be
edited in both places (PR #493 and PR #497). This makes the script read the
canonical registry via `jq`, so the next path change lands in one place.

## Semantic decision: full registry minus the harness

The old array listed only 4 of the 8 registry entries. Step 2 of the script
chowns each project dir to `haduong:dev-projects` with setgid so the overnight
`claude-agent` can write to it. Every registry project is such a dir, so all of
them belong — the old 4 were an incomplete snapshot that silently omitted
`git-erg`, `chemin-de-voix`, and `padme`. The one principled exclusion is the
harness (`~/.claude`): steps 4-5 grant it read-only ACL access, and chowning it
to the group with `g+rwX` would contradict that design. `derive_project_dirs`
therefore emits every registry entry except the harness — the original 4 keep
identical treatment, the 3 previously-missing projects are now covered, and a
registry edit propagates without touching the script.

## Robustness

- Tilde expansion mirrors the Python consumers' `expanduser` via `${path/#\~/$HOME}` (leading `~` only).
- Paths flow through `jq @tsv` → `read -r` loop → `mapfile`, so `Fuzzy Corpus` and any future spaced path survive as single elements (no word-splitting).
- New `--list` mode prints the derived dirs and exits before any sudo — a side-effect-free, testable entry point.

## Test (RED → GREEN)

`tests/test_setup_claude_agent_projects.sh` drives `--list` against a fixture
registry and asserts: harness excluded, `~` expanded to `$HOME`, no tilde leaks,
spaced path stays one element, exact count. Auto-discovered by
`test_bash_suites.py`. `make check`: 686 passed.

## Other hardcoded copies

Swept `scripts/` and `skills/`: no other list existed. Remaining grep hits are
incidental single-project-name mentions in skill docs/examples, not a copy of
the project list.

**Ticket:** tickets/0286-setup-claude-agent-sh-projects-array-dup.erg

### DIFF
--- a/scripts/setup-claude-agent.sh
+++ b/scripts/setup-claude-agent.sh
@@ -4,12 +4,33 @@
 set -euo pipefail
 
 HARNESS="$HOME/.claude"
-PROJECTS=(
-    "$HOME/CNRS/papiers/actif/AEDIST-technical-report"
-    "$HOME/CNRS/papiers/actif/Cadens"
-    "$HOME/CNRS/projets/actifs/climate-finance-het"
-    "$HOME/CNRS/papiers/actif/Fuzzy Corpus"
-)
+# Canonical project registry — the single source of truth also read by
+# beat.py, the nightbeat survey, and scry. Override for tests.
+REGISTRY="${PROJECTS_JSON:-$HARNESS/scripts/projects.json}"
+
+# Derive the project directories that get dev-projects group ownership.
+# Reads the registry, expands a leading ~ to $HOME (as the Python consumers do
+# via expanduser), and excludes the harness itself: ~/.claude receives
+# read-only ACL access in steps 4-5, so granting it group write here would
+# contradict that design and expose the config tree. Every other registry
+# entry is a project the overnight agent works on and needs write access to.
+derive_project_dirs() {
+    local name path
+    while IFS=$'\t' read -r name path; do
+        path="${path/#\~/$HOME}"          # expand leading ~ only
+        [ "$path" = "$HARNESS" ] && continue
+        printf '%s\n' "$path"
+    done < <(jq -r '.[] | [.name, .path] | @tsv' "$REGISTRY")
+}
+
+# `--list` prints the derived dirs and exits — the testable, side-effect-free
+# entry point. Everything below it mutates the system and needs sudo.
+if [ "${1:-}" = "--list" ]; then
+    derive_project_dirs
+    exit 0
+fi
+
+mapfile -t PROJECTS < <(derive_project_dirs)
 
 echo "── 1. User + group ──────────────────────────────────────────────────────"
 sudo groupadd -f dev-projects
--- a/scripts/setup-claude-agent.sh
+++ b/scripts/setup-claude-agent.sh
@@ -15,12 +15,25 @@ REGISTRY="${PROJECTS_JSON:-$HARNESS/scripts/projects.json}"
 # contradict that design and expose the config tree. Every other registry
 # entry is a project the overnight agent works on and needs write access to.
 derive_project_dirs() {
-    local name path
+    local name path tsv
+    # Capture jq's output into a variable BEFORE the loop so its exit status is
+    # observable under `set -e`. A `< <(jq …)` process substitution hides jq's
+    # failure — `set -e` only reacts to a pipeline's status under `pipefail`,
+    # and process substitution is neither — so a missing or malformed registry
+    # would silently yield an empty list and exit 0, and the ownership loop
+    # would run zero iterations while the script still printed "Done." A valid
+    # but empty registry (`[]`) yields empty output with jq exit 0, which is a
+    # legitimate degenerate state and must NOT abort.
+    tsv="$(jq -r '.[] | [.name, .path] | @tsv' "$REGISTRY")" \
+        || { echo "derive_project_dirs: failed to read registry $REGISTRY" >&2; exit 1; }
+    # Empty registry (`[]`) → empty $tsv. A here-string of "" still feeds one
+    # blank line, so guard it to avoid emitting a spurious empty project dir.
+    [ -z "$tsv" ] && return 0
     while IFS=$'\t' read -r name path; do
         path="${path/#\~/$HOME}"          # expand leading ~ only
         [ "$path" = "$HARNESS" ] && continue
         printf '%s\n' "$path"
-    done < <(jq -r '.[] | [.name, .path] | @tsv' "$REGISTRY")
+    done <<< "$tsv"
 }
 
 # `--list` prints the derived dirs and exits — the testable, side-effect-free
--- a/scripts/setup-claude-agent.sh
+++ b/scripts/setup-claude-agent.sh
@@ -43,7 +43,19 @@ if [ "${1:-}" = "--list" ]; then
     exit 0
 fi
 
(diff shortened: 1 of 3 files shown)


MinhHaDuong:
## /gaze actions

round: 2
adherence: PASS — 1 blocking (round 1, addressed for `--list` path only; see round 2)
review-pr: 1 (findings synthesized, not posted individually — see below)
simplify: skipped (adherence blocking, round 1 early-exit)
fix agent: 1 commit (round 2)

## /verify-gate verdict

verdict: ESCALATE

Exit criteria:
- setup-claude-agent.sh has no hardcoded project list: ADDRESSED — `derive_project_dirs()` reads `$REGISTRY` via `jq` (commit cd68c6d); confirmed no literal path array remains.
- behavior unchanged for existing projects: ADDRESSED — `--list` derives the same 7 non-harness dirs as before, including the intentional 4→7 expansion disclosed in the PR body.
- a registry edit propagates without touching the script: ADDRESSED — no script edit needed for registry changes.
- (round-1 carryover) a broken registry must fail loudly, not silently no-op: **NOT MET** — see below.

Unresolved review comments: none (first pass; no prior human comments)

Unresolved simplify: none (phase skipped, round 1 adherence was blocking)

Adherence: FAIL (1, carried from round 1 — see below)

Rationale:
Round 1 found that `derive_project_dirs()` read the registry through `< <(jq …)` process substitution, so a `jq` failure (missing/malformed `scripts/projects.json`) was invisible to `set -e`: the script silently produced an empty `PROJECTS` array, the ownership loop (`chown`/`chmod`/`setgid` — the script's actual purpose) ran zero iterations, and the script still printed "Done." as if it succeeded. Two of three review agents independently reproduced this.

The round-2 fix (commit `9b0f8c4`) correctly hardens the **inner** call: `derive_project_dirs()` now captures `jq`'s output via command substitution (`tsv="$(jq …)" || exit 1`) instead of process substitution, and 3 new tests confirm the `--list` entry point now fails loudly on a bad registry while still accepting a legitimately empty `[]` registry.

But the script's real, mutating entry point — the one that runs `sudo chown/chmod/setgid`, i.e. the whole point of the script — still calls this function through a **second, untouched** process substitution:

```
mapfile -t PROJECTS < <(derive_project_dirs)
```

`derive_project_dirs()`'s `exit 1` now fires, but only inside the subshell that `<(...)` forks; it terminates that subshell, not the calling script. `mapfile` reads to EOF successfully (its own exit status is 0), so `set -e` never trips at the call site. This is the identical bug class the round-1 finding described, relocated one call frame up — and it is the path that matters, since the new tests and the fix agent's own verification only exercised `--list`.

Independently reproduced end-to-end (isolated repro, not the real system): with the registry pointing at a nonexistent file, the script prints `jq`'s error and `derive_project_dirs`'s own loud message, then continues through every subsequent step — ownership loop with zero iterations, no error — ending in "Done." with exit 0.

Scope was held (both commits together touch only `scripts/setup-claude-agent.sh` and `tests/test_setup_claude_agent_projects.sh`, matching the ticket's Relevant-files section — no `scope_overflow`). CI is green on all 8 checks, but none of them execute the shell test suite, so CI could not have caught this. Two SC2034 nits and a ticket-wording ambiguity remain informational/non-blocking, unchanged from round 1.

This is round 2 (final); per `/gaze` protocol a third reroll is forbidden, so this escalates to human judgment rather than looping again.

/gaze stopped: the round-1 blocking finding (silent no-op ownership step on an unreadable/malformed registry) is fixed for the `--list` convenience path but **not** for the script's actual mutating run (`mapfile -t PROJECTS < <(derive_project_dirs)` at the line following the `--list` guard). The mechanical fix is to reuse the same `tsv`-capture-then-check pattern at that call site (e.g. capture `derive_project_dirs`'s output via `PROJECTS_RAW="$(derive_project_dirs)" || exit 1` and split with `mapfile -t PROJECTS <<< "$PROJECTS_RAW"`, guarding the empty-array case the same way the function itself now does) — but a third automated reroll round is not permitted, so this needs a human's next commit or explicit sign-off to reroll manually.

telemetry: wall=~35min agents=5 tokens=na cost~=na

MinhHaDuong:
/verify-gate round=3 verdict=APPROVED (standalone-override)
Exit criteria: 4/4 (3 ticket + 1 round-2 carryover)  Review: 0 unresolved  Simplify: 0
Adherence: 0 blocking (2 SC2034 nits, informational)  Scope overflow: 0
Minors: verifiable:0 consider:0 nofollow:0 malformed:0

Round derivation: prior /gaze comment records rounds 1-2 ending in ESCALATE; this is a manual round-3 re-gate, permitted only as standalone-override. It gates the post-escalation fix commit d96e762 — exactly the human next-commit the escalation asked for. The gate never merges; merge remains the author's call.

Evidence for the carryover criterion (broken registry must fail loudly on the MUTATING path):
- Fix: d96e762, scripts/setup-claude-agent.sh:53-58 — `derived="$(derive_project_dirs)"` top-level command substitution replaces `mapfile < <(…)`, so the child's exit 1 now trips `set -e` at the real entry point; empty output guarded so `[]` still yields `PROJECTS=()`.
- Independent repro, pre-fix (9b0f8c4 script, stub sudo, missing registry): script continued past the jq failure — 6 sudo invocations executed, ownership loop no-oped silently; on the real system it would have ended in "Done." exit 0.
- Independent repro, post-fix (branch HEAD, same setup): rc=1, stderr carries both jq's and derive_project_dirs's errors, 0 bytes stdout, 0 sudo calls.
- Degenerate case preserved: PROJECTS_JSON=[] → script proceeds normally with zero project dirs (documented intent).
- Regression test: "mutating entry point aborts on a bad registry before any sudo" in tests/test_setup_claude_agent_projects.sh (stubbed sudo, asserts rc!=0 + derive error + empty sudo log). Full suite: 10/10 PASS; auto-discovery confirmed via tests/test_bash_suites.py (17 passed).

Original exit criteria (re-verified, unchanged from round 2):
- No hardcoded project list: ADDRESSED — cd68c6d, derive_project_dirs() reads $REGISTRY via jq (script lines 17-37); no literal array remains.
- Behavior unchanged for existing projects: ADDRESSED — --list on the real registry derives the 7 non-harness dirs (original 4 + disclosed git-erg/chemin-de-voix/padme expansion); "Fuzzy Corpus" survives as one element.
- Registry edit propagates without script edit: ADDRESSED — registry read at runtime; tests drive alternate registries through PROJECTS_JSON with no script change.

Scope: all 3 commits touch only scripts/setup-claude-agent.sh + tests/test_setup_claude_agent_projects.sh, matching the ticket's Relevant-files. CI 8/8 green on d96e762.

Rationale: the strongest remaining attack is that an empty-after-filtering registry silently runs the non-project steps (1, 3-7) with zero ownership work — but that is the explicitly documented degenerate design, distinguishable on stderr from the failure path, and predates this PR's scope. The round-2 escalation trigger is closed with byte-level evidence on the exact call site it named; nothing else fires a decision rule.

PR #517, opened 2026-07-13
fix(0297): guard-commit-on-main checks tree identity, not just branch name

MinhHaDuong:
## Summary

`scripts/guard-commit-on-main.sh` checked only `branch == main`. In incident I1 (from the 0252 worktree-ownership audit), a stray `git checkout -B` moved the PRIMARY checkout onto a feature branch, after which this guard happily allowed commits on the primary — its exact failure mode. The material predicate is "this commit lands on the primary checkout", not "the branch is named main".

## Fix

- **Layer 1 (new): tree identity.** Consume `.cwd` from the hook JSON (same idiom as `scripts/guard-cd-primary-repo.sh:21-40`). When the cwd worktree path resolves via `git -C "$cwd" rev-parse --show-toplevel` back to the primary root, block unconditionally regardless of branch. An unregistered/deregistered worktree makes git's toplevel walk-up land on the primary tree — this catches it.
- **Layer 2 (kept): branch name.** The existing `branch==main` check stays, now using `git -C "$cwd"` when cwd is present and bare git otherwise — preserving today's behaviour for payloads without `.cwd`, where layer 1 is silently skipped (a noted residual gap for cwd-less callers).
- Fail-closed jq handling and `set -euo pipefail` discipline preserved. No shared helper extracted (a dedup-consideration ticket is being filed by a sibling PR).

## RED proof

The new I1 test case — a primary repo moved onto a `feature` branch, an unregistered plain directory under `.claude/worktrees/` whose git toplevel walks back up to the primary — **passed wrongly against the pre-fix script** while every existing assertion stayed green:

```
FAIL: expected block (exit 2) for git commit on primary via plain worktree dir (I1); got exit 0 — cmd: git commit -m x
```

After the fix, all 9 cases pass (including a positive `git worktree add` fixture on a feature branch → commit allowed, confirming no false positive on legitimate worktree commits). `make check` green.

**Ticket:** tickets/0297-guard-commit-on-main-false-negative.erg

### DIFF
--- a/scripts/guard-commit-on-main.sh
+++ b/scripts/guard-commit-on-main.sh
@@ -21,20 +21,28 @@ echo "$cmd" | grep -qP '\bgit\s+commit\b' || exit 0
 # branch-only check: a stray `git checkout -B` moved the primary off main, and
 # the guard then allowed commits on it. Consume `.cwd` from the hook JSON (the
 # same idiom as scripts/guard-cd-primary-repo.sh:21-40) and block when the cwd
-# worktree path resolves back to the primary root, whatever the branch.
+# resolves to the primary checkout, whatever the branch.
+#
+# Derive "primary root" from git plumbing, not from a path-string pattern:
+# `--git-common-dir` points at the shared .git of the primary checkout, so its
+# parent IS the primary root — for a bare primary cwd AND for any worktree,
+# registered or not. When the cwd's own toplevel equals that primary root, the
+# commit lands on the primary. A registered worktree has its own distinct
+# toplevel, so it falls through to layer 2. This is the same technique as
+# scripts/pretooluse-worktree-path-guard.sh:24-26 and needs no `.claude/worktrees`
+# naming convention — the earlier case-gated version silently no-op'd when the
+# primary was addressed as a bare root path (ticket 0297, round-1 gate finding).
 cwd=$(echo "$input" | jq -r '.cwd // empty' 2>/dev/null || true)
 if [ -n "$cwd" ]; then
-    case "$cwd" in
-        */.claude/worktrees/*)
-            # primary_root = the prefix before the worktree marker.
-            primary_root="${cwd%%/.claude/worktrees/*}"
-            toplevel=$(git -C "$cwd" rev-parse --show-toplevel 2>/dev/null || true)
-            if [ -n "$toplevel" ] && [ "$toplevel" = "$primary_root" ]; then
-                echo "BLOCKED: cwd '$cwd' looks like a worktree but git resolves it to the PRIMARY checkout ($primary_root) — an unregistered/deregistered worktree falls through to the primary tree. Committing here lands on the primary. Create/enter a real worktree first." >&2
-                exit 2
-            fi
-            ;;
-    esac
+    toplevel=$(git -C "$cwd" rev-parse --show-toplevel 2>/dev/null || true)
+    git_common_dir=$(git -C "$cwd" rev-parse --path-format=absolute --git-common-dir 2>/dev/null || true)
+    if [ -n "$toplevel" ] && [ -n "$git_common_dir" ]; then
+        primary_root=$(dirname "$git_common_dir")
+        if [ "$toplevel" = "$primary_root" ]; then
+            echo "BLOCKED: cwd '$cwd' resolves to the PRIMARY checkout ($primary_root) — a bare primary root, or an unregistered/deregistered worktree that falls through to it. Committing here lands on the primary, whatever the branch. Create/enter a real worktree first." >&2
+            exit 2
+        fi
+    fi
 fi
 
 # Layer 2 — branch name. Use `git -C "$cwd"` when cwd is known; bare git
--- a/scripts/guard-commit-on-main.sh
+++ b/scripts/guard-commit-on-main.sh
@@ -16,8 +16,35 @@ cmd=$(echo "$input" | jq -r '.tool_input.command // empty')
 # Only check git commit commands
 echo "$cmd" | grep -qP '\bgit\s+commit\b' || exit 0
 
-# Get current branch
-branch=$(git branch --show-current 2>/dev/null || true)
+# Layer 1 — tree identity. Branch name is a proxy; the material predicate is
+# "this commit lands on the PRIMARY checkout". Incident I1 defeated the
+# branch-only check: a stray `git checkout -B` moved the primary off main, and
+# the guard then allowed commits on it. Consume `.cwd` from the hook JSON (the
+# same idiom as scripts/guard-cd-primary-repo.sh:21-40) and block when the cwd
+# worktree path resolves back to the primary root, whatever the branch.
+cwd=$(echo "$input" | jq -r '.cwd // empty' 2>/dev/null || true)
+if [ -n "$cwd" ]; then
+    case "$cwd" in
+        */.claude/worktrees/*)
+            # primary_root = the prefix before the worktree marker.
+            primary_root="${cwd%%/.claude/worktrees/*}"
+            toplevel=$(git -C "$cwd" rev-parse --show-toplevel 2>/dev/null || true)
+            if [ -n "$toplevel" ] && [ "$toplevel" = "$primary_root" ]; then
+                echo "BLOCKED: cwd '$cwd' looks like a worktree but git resolves it to the PRIMARY checkout ($primary_root) — an unregistered/deregistered worktree falls through to the primary tree. Committing here lands on the primary. Create/enter a real worktree first." >&2
+                exit 2
+            fi
+            ;;
+    esac
+fi
+
+# Layer 2 — branch name. Use `git -C "$cwd"` when cwd is known; bare git
+# otherwise (preserves today's behaviour for payloads that carry no `.cwd`,
+# where layer 1 is silently skipped — a residual gap for cwd-less callers).
+if [ -n "$cwd" ]; then
+    branch=$(git -C "$cwd" branch --show-current 2>/dev/null || true)
+else
+    branch=$(git branch --show-current 2>/dev/null || true)
+fi
 
 if [ "$branch" = "main" ] || [ "$branch" = "master" ]; then
     echo "BLOCKED: committing directly to $branch. Create a branch first: git switch -c <branch-name>" >&2
--- a/tests/test_guard_commit_on_main.sh
+++ b/tests/test_guard_commit_on_main.sh
@@ -9,7 +9,7 @@ fail=0
 
 # Create a temp git repo so `git branch --show-current` behaves predictably.
 TMPDIR=$(mktemp -d)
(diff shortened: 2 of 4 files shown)


MinhHaDuong:
## /verify-gate round 1 — REROLL

Independently re-ran the gate against the current branch tip (`bef6934`, includes the simplify commit). Confirms the correctness gap flagged by two of the three prior reviewers.

### Reproduction (4th independent confirmation)

```
TDIR=$(mktemp -d); cd "$TDIR"
git init -q; git config user.email t@example.com; git config user.name T
git commit --allow-empty -q -m init
git branch -m main
git checkout -B feature -q
PAYLOAD=$(printf '{"tool_input":{"command":"git commit -m x"},"cwd":"%s"}' "$TDIR")
echo "$PAYLOAD" | bash scripts/guard-commit-on-main.sh
# exit 0 (ALLOWED) — should be 2 (blocked)
```

Root cause: Layer 1 in `scripts/guard-commit-on-main.sh` is gated by
`case "$cwd" in */.claude/worktrees/*)`. When `.cwd` is the primary
checkout root directly — not dressed up as a path under
`.claude/worktrees/` — Layer 1 is a complete no-op and only Layer 2
(branch name) runs. This is exactly the I1 scenario in the ticket's own
Context section, and the ticket's Actions §1 describes the check as
unconditional ("block commits whose `git rev-parse --show-toplevel` is
the primary checkout root, **whatever the branch name**" — no cwd-shape
qualifier).

### Verification checklist (tickets/0297-guard-commit-on-main-false-negative.erg)

- [ ] Commit on primary checkout blocked regardless of branch name — **MISSING**, reproduced above.
- [x] Commit in a registered worktree on a feature branch still allowed — ADDRESSED (`tests/test_guard_commit_on_main.sh` "allows git commit in real registered worktree on feature branch").
- [x] Guard reads cwd from hook JSON, not from its own process cwd — ADDRESSED (`.cwd` consumed via jq).

### Exit criteria

- [ ] Both checks live in the guard with tests proving RED on the pre-fix state — the checks are present, but the test suite's only RED-proving I1 case is gated to a path under `.claude/worktrees/`, not the bare-root cwd the ticket's own unqualified verification item and Context describe. Given together with the unmet checklist item above, **not satisfied** as written.
- [x] `make check` passes — confirmed directly: `bash tests/test_guard_commit_on_main.sh` → all 10 assertions PASS; full `make check` → 734 passed in 23.31s.

### Scope check

`git log origin/main..origin/t0297 --stat` shows only the two expected commits (`d9484f4` fix, `bef6934` simplify) touching `scripts/guard-commit-on-main.sh` and `tests/test_guard_commit_on_main.sh`. No scope overflow.

### Review-comment disposition

- /review-pr (correctness+consistency), blocking finding on the same Layer-1 case-gate no-op — **UNRESOLVED** (reproduced).
- /review (Agent B), same substance under an "approve with comment" label — **UNRESOLVED**, treated as pointing the same direction as the blocking finding per the orchestrator's framing.
- /review-pr minors: pre-existing line-13 jq `2>/dev/null` gap (out of this PR's diff) and an unreproduced symlink-canonicalization hypothesis — informational, non-blocking.
- Adherence agent: PASS, no findings — expected; mechanical checks don't cover this correctness gap.
- /simplify design notes (reuse of `pretooluse-worktree-path-guard.sh`'s path-shape-independent primitive; altitude notes on the 0252 audit's "ship first" shared-guard proposal and the gaze `/tmp` blind spot) — non-blocking, out of this ticket's stated scope ("both checks live in the guard"), noted for a follow-up ticket.

### Verdict

**REROLL.** The ticket's own unqualified verification item is unmet and reproducible on demand; this is a real gap in the incident this PR exists to close, not a scope question. Recommend either dropping the `*/.claude/worktrees/*` case-gate on Layer 1 (run the tree-identity check unconditionally whenever `.cwd` is present) or adopting the sibling `pretooluse-worktree-path-guard.sh` primitive that doesn't depend on the `.claude/worktrees` string convention, then extending the test to cover the bare-root cwd case directly.

MinhHaDuong:
## /verify-gate round 2 (final) — APPROVED

Independently re-ran the full gate against tip `1380445` (on `c14647c` gate-log, on `bef6934` simplify, on `d9484f4` original fix).

### Round-1 blocking finding — resolved, confirmed independently (5th reproduction)

Round 1 found Layer 1 was gated by `case "$cwd" in */.claude/worktrees/*)`, a no-op whenever `.cwd` was the primary root addressed directly. `1380445` replaces it with a git-plumbing derivation: `primary_root=$(dirname "$(git -C "$cwd" rev-parse --path-format=absolute --git-common-dir)")`, compared against `git -C "$cwd" rev-parse --show-toplevel` — no `.claude/worktrees` string dependency at all (`scripts/guard-commit-on-main.sh:35-46`).

Ran the round-1 repro myself against **both** versions:
- Pre-fix (`bef6934`, the version round 1 REROLLed): `exit 0` — wrongly allowed. Confirms genuine RED on the pre-fix state.
- Post-fix (`1380445`): `exit 2` — correctly blocked.

Also independently constructed and ran two extra scenarios not in the ticket's literal repro, to stress-test the new logic beyond the test file's own fixtures:
- Real registered worktree (`git worktree add`) on a feature branch → `exit 0` (allowed), confirmed.
- An arbitrary unregistered nested directory under the primary root (not shaped like `.claude/worktrees/...` at all) with primary off-main → `exit 2` (blocked), confirming the fix generalizes past the specific `.claude/worktrees` path shape, not just past the literal repro.

### Verification checklist (tickets/0297-guard-commit-on-main-false-negative.erg)

- [x] Commit on primary checkout blocked regardless of branch name — ADDRESSED (`1380445`, `scripts/guard-commit-on-main.sh:35-46`; reproduced above).
- [x] Commit in a registered worktree on a feature branch still allowed — ADDRESSED (test file line ~131 "real registered worktree" case; reproduced independently).
- [x] Guard reads cwd from hook JSON, not from its own process cwd — ADDRESSED (unchanged, `scripts/guard-commit-on-main.sh:35`).

### Exit criteria

- [x] Both checks live in the guard with tests proving RED on the pre-fix state — ADDRESSED. New test case `tests/test_guard_commit_on_main.sh:116-123` ("bare primary root off main (0297)"); verified it fails (wrongly passes) against `bef6934`'s script and passes against `1380445`'s.
- [x] `make check` passes — confirmed: `bash tests/test_guard_commit_on_main.sh` → 11/11 PASS (was 10, gained the 0297 case); full `make check` → exit 0, 734 pytest items collected, no failures.

### Scope check

`git log origin/main..origin/t0297 --stat` — exactly 4 commits (`d9484f4`, `bef6934`, `c14647c`, `1380445`), touching only `scripts/guard-commit-on-main.sh`, `tests/test_guard_commit_on_main.sh`, and `tickets/0297-guard-commit-on-main-false-negative.erg`. The ticket-file diff is confirmed as exactly the one round-1 gate log line (`+1/-0`), no exit-criteria checkbox or body edits — no scope violation of the "gate's one permitted write is the log line" convention.

### Review-comment disposition

- Round-1 blocking finding (the only one that mattered for the verdict): RESOLVED, independently re-verified above.
- Round-1 informational-only items (pre-existing line-13 jq `2>/dev/null` gap, unreproduced symlink hypothesis, /simplify's altitude notes on shared-guard extraction / other mutating verbs / audit's broader "ship first" proposal): remain explicitly out of this ticket's scope, non-blocking, unchanged since round 1.
- No new problems found in this independent round-2 pass; the git-plumbing technique (mirrors `scripts/pretooluse-worktree-path-guard.sh:24-26`) checked correct for primary-root, registered-worktree, and unregistered-worktree-shaped-directory cases by direct reasoning and manual reproduction, not just by trusting the existing test fixtures.

### Verdict

**APPROVED.** This is the final round per policy (round 3 is forbidden). The round-1 gap is closed and independently reproduced fixed; no new blocking findings surfaced in this pass.

MinhHaDuong:
## /gaze actions

round: 2
adherence: PASS — 0 blocking
review-pr: 1 blocking finding posted (correctness) | consistency perspective clean
simplify: 1 fix applied (test-helper dedup, bef6934); design notes deferred as non-blocking
fix agent: 1 commit (round-1→round-2 fix, commit 1380445)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Commit on primary checkout blocked regardless of branch name: ADDRESSED — commit 1380445 replaced the `.claude/worktrees`-string-gated check with a git-plumbing derivation (`git rev-parse --show-toplevel` vs `dirname($(git rev-parse --path-format=absolute --git-common-dir))`), adapted from `scripts/pretooluse-worktree-path-guard.sh:24-26`. Reproduced RED (exit 0) on pre-fix bef6934, GREEN (exit 2) on post-fix 1380445, independently by two separate agents plus the orchestrator.
- Commit in a registered worktree on a feature branch still allowed: ADDRESSED — `tests/test_guard_commit_on_main.sh` REALWT case; reproduced manually, exit 0.
- Guard reads cwd from hook JSON, not from its own process cwd: ADDRESSED — unchanged from the original PR, `scripts/guard-commit-on-main.sh:35`.
- Both checks live in the guard with tests proving RED on the pre-fix state: ADDRESSED — 11/11 assertions in `tests/test_guard_commit_on_main.sh`, including the new bare-primary-root case added in the round-1 fix.
- make check passes: ADDRESSED — full suite green (734 items), confirmed independently three times across the run.

Unresolved simplify findings: none blocking. Deferred as non-blocking/out-of-scope (per round-1 gate and unchanged in round 2): reuse of `pretooluse-worktree-path-guard.sh`'s technique (now actually adopted by the round-1 fix — resolved as a side effect); altitude notes on covering other mutating verbs beyond `git commit`, and the 0252 audit's "ship first" shared-guard proposal — explicitly out of this ticket's stated scope ("both checks live in the guard"), candidate for a follow-up ticket.
Adherence: PASS (0)

Rationale:
Round 1 found a genuine, reproducible correctness gap: the PR's original Layer 1 tree-identity check was gated by a `*/.claude/worktrees/*` path-string pattern, making it a complete no-op whenever the primary checkout was addressed by a bare root path with no worktree-marker segment — exactly the scenario the ticket's own Context section describes (a stray branch-switch on the primary checkout moving it off main). This was confirmed independently by two of three phase 2-4 reviewers and by direct orchestrator reproduction before the gate even ran. The round-1 fix agent replaced the string-matching approach with a git-plumbing-based primary-root derivation reused from a sibling guard script, closing the gap without depending on any path-naming convention. Round 2 independently re-verified the fix (RED-to-GREEN reproduction, plus two additional stress scenarios beyond the ticket's literal repro), confirmed no regression on the two protected cases (real registered worktree allowed; unregistered worktree-shaped fall-through still blocked), confirmed full test and adherence suites green, and confirmed no scope overflow (4 commits, 3 files, ticket-file diff limited to a single round-1 log line). All exit criteria ADDRESSED, no unresolved findings — clear to merge.

telemetry: wall=~50min agents=8 tokens=na cost~=na

PR #519, opened 2026-07-13
guard(0296): live worktree-identity preflight for mutating git/erg commands

MinhHaDuong:
## Summary

New PreToolUse Bash guard `scripts/guard-worktree-identity.sh` that blocks a mutating git/erg command when the cwd claims a harness worktree (`.../.claude/worktrees/<name>/`) but `git rev-parse --show-toplevel` resolves elsewhere. This closes incident I1 from the 0252 audit: a session's worktree is deregistered mid-session, the directory lingers, so bare git falls through to the PRIMARY repo and a `git checkout -B` moves the primary checkout off main. Every existing guard checks only the weak predicate "cwd is some worktree"; this one re-derives ownership per command.

### Guard logic
- Two pure-string fast-paths run before any subprocess: cwd not under `.claude/worktrees/` → allow; command carries no mutating verb → allow.
- Mutating verbs: git (commit|checkout|switch|merge|rebase|push|reset|add|mv|rm|stash), erg (new|close|log|label|unlabel|archive|rm|migrate|init|install|update) — install/update/init gated because `skills/molt` calls `erg update` to replace the binary.
- Whitelist: `git -C <path>` names its target tree, so cwd identity is irrelevant — the sanctioned cross-tree idiom passes.
- One `git -C "$cwd" rev-parse --show-toplevel` (never ambient cwd). Empty (cwd deleted) → fail-safe allow.
- Block (exit 2, explanatory stderr) when `basename $toplevel != worktree_name` OR `$toplevel == primary_root`.
- Fails CLOSED when jq is missing (unlike the advisory guard-cd-primary-repo.sh) — this guard protects primary-checkout integrity.

### Tests
`tests/test_guard_worktree_identity.sh`, auto-discovered by `test_bash_suites.py`, uses REAL git fixtures (no mocked rev-parse):
- (a) I1 replay — plain unregistered dir under `.claude/worktrees/` → BLOCKED (rev-parse walks up to primary).
- (b) false-positive path — real `git worktree add` → ALLOWED.
- (c) `git -C <path>` anywhere → ALLOWED regardless of cwd.
- (d) read-only git (status/log/diff) in the broken worktree → ALLOWED (verb fast-path).
- Plus: primary-root cwd, deleted cwd, missing `.cwd`, and jq-missing fail-closed.

`make check` passes.

**Ticket:** tickets/0296-live-worktree-identity-guard.erg

Related follow-up: tickets/0302-extract-shared-worktree-identity-predica.erg (dedup the predicate across the three 0296/0297/0301 sites, or record why they stay inline).

### DIFF
--- a/scripts/guard-worktree-identity.sh
+++ b/scripts/guard-worktree-identity.sh
@@ -39,20 +39,63 @@ case "$cwd" in
     *) exit 0 ;;
 esac
 
+# Normalize away shell quote characters before matching. A quoted verb token
+# (`git "commit"`) otherwise breaks the git↔verb adjacency the regex relies on
+# and slips through. Stripping quotes also splits a semicolon/ampersand that was
+# hiding inside a quoted argument into its own segment below, which only makes
+# the guard MORE conservative (it errs toward the identity check, never away).
+# Two indirections stay opaque to any static command-string scan and are OUT OF
+# SCOPE here: command substitution (`git $(echo commit)`) and git aliases
+# (`git ci` for commit) — a regex cannot see through either without invoking git
+# itself. See ticket 0302 if closing the alias gap ever becomes necessary.
+norm=${cmd//\"/}
+norm=${norm//\'/}
+
+# Mutating-verb patterns, reused by the fast path and the per-segment scan.
+# The git list adds the destructive verbs the original allowlist missed
+# (cherry-pick/revert/apply/am/restore/filter-branch/gc/pull/clean, worktree
+# remove|move|prune|add, branch|tag with a delete/force/move short flag, config,
+# fetch --prune). Dual-mode verbs (branch/tag) are gated only in their
+# destructive flag form so read-only listing (`git branch -vv`) still passes.
+# Long-form flags (`git branch --delete`) are a known residual gap.
+GIT_MUT='\bgit\s+(commit|checkout|switch|merge|rebase|push|reset|add|mv|rm|stash|cherry-pick|revert|apply|am|restore|filter-branch|gc|pull|clean|worktree\s+(remove|move|prune|add)|(branch|tag)\s+-[a-zA-Z]*[dDfFmM]|config|fetch\s+--prune)\b'
+ERG_MUT='\berg\s+(new|close|log|label|unlabel|archive|rm|migrate|init|install|update)\b'
+
 # Fast path 2 (pure string, no git call yet): command carries no mutating verb →
 # read-only git (status/log/diff) and everything else pass untouched.
 # erg's install/update/init also mutate on-disk state (skills/molt calls
 # `erg update` to replace the installed binary), so they are gated too.
-echo "$cmd" | grep -qP '\bgit\s+(commit|checkout|switch|merge|rebase|push|reset|add|mv|rm|stash)\b' \
-  || echo "$cmd" | grep -qP '\berg\s+(new|close|log|label|unlabel|archive|rm|migrate|init|install|update)\b' \
+echo "$norm" | grep -qP "$GIT_MUT" \
+  || echo "$norm" | grep -qP "$ERG_MUT" \
   || exit 0
 
-# Whitelist: an explicit `git -C <path>` names its target tree, so the cwd
-# identity is irrelevant — that idiom is the sanctioned way to mutate another
-# tree on purpose (rules/git.md).
-if echo "$cmd" | grep -qP '\bgit\s+-C\s'; then
-    exit 0
-fi
+# Whitelist, re-checked PER SEGMENT (ticket 0302 reroll). An explicit
+# `git -C <path>` names its target tree, so the cwd identity is irrelevant —
+# the sanctioned cross-tree idiom (rules/git.md). But the exemption is sound
+# ONLY for the segment that carries the -C: a compound like
+# `git -C X status && git commit -m y` must NOT let the leading `git -C` whitelist
+# its trailing BARE `git commit`, which mutates whatever bare git resolves to.
+# The old guard tested the literal substring `git -C ` over the WHOLE command and
+# waved the entire chain through — the I1 fall-through this guard exists to close.
+# Fix: split on shell separators (&&, ||, ;, |, &) and require every MUTATING
+# segment to name -C in ITSELF; any mutating segment without its own -C forces
+# the identity check below. (Note: printf adds a trailing newline so `read`
+# does not drop the final, separator-less segment.)
+needs_check=0
+while IFS= read -r seg; do
+    if echo "$seg" | grep -qP "$GIT_MUT"; then
+        if echo "$seg" | grep -qP '\bgit\s+-C\s'; then
+            continue
+        fi
+        needs_check=1
+        break
+    fi
+    if echo "$seg" | grep -qP "$ERG_MUT"; then
+        needs_check=1
+        break
+    fi
+done < <(printf '%s\n' "$norm" | tr ';&|' '\n')
+[ "$needs_check" -eq 0 ] && exit 0
 
 # cwd = .../.claude/worktrees/<name>/...  →  extract <name> and the primary root.
 worktree_rest="${cwd#*/.claude/worktrees/}"
--- a/scripts/guard-worktree-identity.sh
+++ b/scripts/guard-worktree-identity.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+set -euo pipefail
+# PreToolUse hook: block a mutating git/erg command when the cwd CLAIMS a harness
+# worktree (`.../.claude/worktrees/<name>/`) but `git rev-parse --show-toplevel`
+# resolves somewhere else — the live worktree-identity preflight (ticket 0296).
+#
+# Incident I1 (aedist, 2026-06-11/12): a session's worktree was deregistered
+# mid-session; the directory lingered, so bare git fell through to the PRIMARY
+# repo and a `git checkout -B` moved the primary checkout off main. Every
+# existing guard checks only the weak predicate ("cwd is *some* worktree"); this
+# one re-derives ownership per command — does rev-parse still match the name the
+# cwd claims — on the hot path. Overhead: one rev-parse (~5-20 ms), and only
+# after two pure-string fast-paths reject the common case.
+#
+# Exit 0 = allow, exit 2 = block with a stderr message.
+#
+# Fail-CLOSED when jq is missing (contrast guard-cd-primary-repo.sh, which is a
+# narrow advisory and fails open): this guard protects the primary-checkout
+# integrity invariant, so a missing parser must block, not wave a mutation
+# through unchecked.
+
+input=$(cat)
+
+if ! command -v jq &>/dev/null; then
+    echo "BLOCKED: jq not found — guard-worktree-identity.sh cannot parse tool input." >&2
+    exit 2
+fi
+
+cmd=$(echo "$input" | jq -r '.tool_input.command // empty' 2>/dev/null || true)
+[ -z "$cmd" ] && exit 0
+
+cwd=$(echo "$input" | jq -r '.cwd // empty' 2>/dev/null || true)
+[ -z "$cwd" ] && exit 0  # fail-safe: cannot tell which tree this session owns
+
+# Fast path 1 (pure string, no subprocess): cwd not under a harness worktree →
+# primary-checkout work is governed by other guards.
+case "$cwd" in
+    */.claude/worktrees/*) ;;
+    *) exit 0 ;;
+esac
+
+# Fast path 2 (pure string, no git call yet): command carries no mutating verb →
+# read-only git (status/log/diff) and everything else pass untouched.
+# erg's install/update/init also mutate on-disk state (skills/molt calls
+# `erg update` to replace the installed binary), so they are gated too.
+echo "$cmd" | grep -qP '\bgit\s+(commit|checkout|switch|merge|rebase|push|reset|add|mv|rm|stash)\b' \
+  || echo "$cmd" | grep -qP '\berg\s+(new|close|log|label|unlabel|archive|rm|migrate|init|install|update)\b' \
+  || exit 0
+
+# Whitelist: an explicit `git -C <path>` names its target tree, so the cwd
+# identity is irrelevant — that idiom is the sanctioned way to mutate another
+# tree on purpose (rules/git.md).
+if echo "$cmd" | grep -qP '\bgit\s+-C\s'; then
+    exit 0
+fi
+
+# cwd = .../.claude/worktrees/<name>/...  →  extract <name> and the primary root.
+worktree_rest="${cwd#*/.claude/worktrees/}"
+worktree_name="${worktree_rest%%/*}"
+primary_root="${cwd%%/.claude/worktrees/*}"
+
+# The only git call: resolve the real toplevel FROM THE CWD explicitly (-C),
+# never the ambient shell cwd. Empty = cwd was deleted / not in a repo →
+# fail-safe allow (nothing to compare against; other guards cover the rest).
+toplevel=$(git -C "$cwd" rev-parse --show-toplevel 2>/dev/null || true)
+[ -z "$toplevel" ] && exit 0
+
+# Identity mismatch: the tree git resolves is not the worktree the cwd claims,
+# or it is the primary root itself (the I1 fall-through). Block.
+if [ "$(basename "$toplevel")" != "$worktree_name" ] || [ "$toplevel" = "$primary_root" ]; then
+    echo "BLOCKED: worktree identity mismatch. cwd claims worktree '$worktree_name' but git resolves to '$toplevel'." >&2
+    echo "This is the I1 fall-through: a deregistered/moved worktree lets bare git mutate the wrong tree (e.g. the primary checkout)." >&2
+    echo "Re-enter the correct worktree, or for intentional cross-tree work name the target explicitly: git -C <path> ..." >&2
+    exit 2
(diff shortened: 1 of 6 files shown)


MinhHaDuong:
## /gaze review-pr — PR #519 (guard-worktree-identity.sh, ticket 0296)

Five perspectives ran in parallel (correctness, consistency, scope, red-team, doc-propagation). `make check` is green in this worktree (735 passed, skills-catalog in sync, agnostic checks clean, exit 0).

**Synthesized verdict: request-changes.** Correctness and red-team, working independently, both reproduced real bypasses of the guard's own I1 threat model with actual git mutations landing on the primary repo. Consistency landed on comment (no blocker). Scope and doc-propagation approve outright. Preserving each perspective's verdict as filed: correctness=request-changes (high), red-team=request-changes (high), consistency=comment (medium), scope=approve (high), doc-propagation=approve (high).

### Blockers

1. **`git -C` whitelist is laundered by an unrelated, harmless `-C` earlier in a chained command.** The whitelist (`grep -qP '\bgit\s+-C\s'`, line 53) matches anywhere in the *whole* command string, with no requirement that the `-C` belongs to the same git invocation as a later mutating verb. Both the correctness and red-team agents independently reproduced this with a real mutation landing on the primary repo:
   - `cmd`: `git -C /tmp/somewhere-harmless status && git commit -m pwned`
   - `cwd`: `<primary>/.claude/worktrees/t001` (deregistered/broken worktree, rev-parse resolves to primary)

   Guard exit code: 0 (allow). Red-team executed this for real — `git commit` walked up per I1 and created a commit in the primary repo. `;` in place of `&&` works identically. This is precisely the I1 fall-through the guard exists to close, defeated by its own escape hatch.
   Fix suggested by both agents: scope the whitelist to the specific `git` clause containing the mutating verb (e.g. split on `&&`/`;`/`|` and re-check each segment, or require `-C` to immediately precede the matched verb's invocation), not the command string as a whole.

2. **The mutating-verb regex is a short fixed allowlist, not a read-only/mutating classifier — large coverage gap.** Both agents enumerated overlapping sets of git subcommands absent from `\bgit\s+(commit|checkout|switch|merge|rebase|push|reset|add|mv|rm|stash)\b` that genuinely mutate or lose state in the wrong tree under the I1 model and returned exit 0 (allow) from the broken-worktree fixture: `git cherry-pick`, `git apply`, `git worktree remove`, `git update-ref`, `git filter-branch --force`, `git gc --prune=now`, `git restore --staged`, `git branch -D`/`-f`, `git tag -d`/`-f`, `git config`, `git fetch --prune`, `git pull`, `git symbolic-ref`, `git revert`, `git am`, and the working-tree-clean command whose name pattern-matches the "destructive bash" guard's own trigger string (omitted verbatim here to avoid re-tripping that unrelated hook — the point stands: it is also absent from the alternation). This is a materially larger gap than the ticket's "mutating git/erg" framing implies.
   Fix suggested: invert to a read-only allowlist (`status|log|diff|show|branch --list|...`) gating everything else, rather than enumerating mutating verbs to catch.

3. **Trivial, non-adversarial evasion via quoting/indirection/aliases.** Red-team confirmed, with real mutations advancing primary HEAD, that ordinary shell constructs defeat the literal-string verb match: `git "commit" -m x`, `v=commit; git $v -m x`, `git $(echo commit) -m x`, and — notably not adversarial at all, just an everyday dev habit — `git ci -m x` with a `git config alias.ci=commit` in the user's gitconfig. Whitespace variants (extra spaces, tabs) are correctly still caught; only quoting/substitution/aliasing evade.

Given 1–3 are independently reproduced by two agents with actual mutations landing on the primary tree — exactly the scenario ticket 0296 and incident I1 exist to close — the guard as merged would give a false sense of the primary-checkout integrity invariant being closed, per its own docstring and PR body framing.

### Minor findings

- `verifiable:` **False positive: nested worktree-of-a-worktree gets wrongly blocked.** `${cwd#*/.claude/worktrees/}` matches the *first* occurrence of the marker. A worktree created inside another worktree (`.../outer/.claude/worktrees/inner`) has `git rev-parse --show-toplevel` correctly resolving to `.../outer/.claude/worktrees/inner`, but the guard extracts `worktree_name=outer`, so a legitimate, correctly-registered commit inside `inner` is wrongly blocked (exit 2, reproduced). Edge case, not a security issue, but a real false positive an author will eventually hit.
- `verifiable:` **False positive: `erg log` is blocked despite being read-only.** The erg-verb regex includes `log`; the git-verb regex correctly excludes it. `git log` is unaffected (matches neither), but `erg log` in a broken-worktree cwd is gated and can be wrongly blocked on identity mismatch even though it mutates nothing.
- `consider:` Fast-path ordering pays for two unconditional `jq` calls (cmd *and* cwd extracted before either string fast-path runs), diverging from `guard-cd-primary-repo.sh`'s stated optimization ("one grep before the second jq"). Reordering to extract `cwd` only after the verb-regex fast-path would match the sibling's own justification for its ordering.
- `consider:` The stated fail-closed rationale ("this guard protects the primary-checkout integrity invariant" vs. the sibling being "a narrow advisory") is imprecise — `guard-cd-primary-repo.sh`'s own header states the same underlying invariant (preventing mutation of the wrong tree); the real distinguishing argument (ground-truth `rev-parse` check vs. syntactic heuristic, and one-signal-among-several vs. sole check) isn't the one written down. Doesn't affect behavior, but could mislead a future maintainer comparing the two.
- `nofollow:` Test suite doesn't assert silence on the allow path the way the sibling's suite does (`_assert_allowed_silent`); low-impact since this script emits nothing on allow.
- `nofollow:` Ticket 0302's shorthand description of `erg-pr-merge`'s `in_worktree()` as `[ -f .git ]` elides the actual companion `grep -q "gitdir:" .git` check; inherited verbatim from the pre-existing audit doc, not introduced by this PR, and the substance (no tree-identity check) still holds.
- `nofollow:` README.md's Structure section guard-script list is already missing 4 of 6 existing guards *before* this PR (pre-existing drift); this PR's guard not being added is one more instance of repo-wide staleness, not a new omission to hold this PR to.

### Scope / consistency / doc-propagation (approve/comment, no blockers)

- Every changed line (guard script, 10-line settings.json hook registration, test suite, ticket 0302 filing) is traceable to ticket 0296's exit criteria or the legitimate 0302 follow-up. `settings.json` diff is a clean, isolated 10-line insertion with no collateral edits. PR body's description of guard logic and tests matches the diff.
- JSON-parsing idiom, exit-code conventions (0/allow, 2/block), and block-message structure are consistent with `guard-cd-primary-repo.sh`. No blocking structural inconsistency found.
- No doc/inventory elsewhere in the repo needs updating for this guard's existence (rules/workflow.md, rules/git.md have no guard-name enumeration; `docs/audit-0252-...md` is an intentionally frozen point-in-time snapshot, not a living doc).

### Recommendation

Given the blast radius (this hook runs on every Bash call in every session on the machine) and that the three blockers all reproduce the exact I1 scenario the guard's own docstring and tests claim to close, recommend a fix round before merge: (a) scope the `-C` whitelist to the matched invocation, (b) invert to a read-only allowlist or materially broaden the mutating-verb set, (c) accept that literal aliasing/substitution evasion may be an acknowledged residual limitation if (a)/(b) are fixed — but that residual should be stated explicitly in the header comment rather than left implicit, since the current comment reads as a completeness claim ("closes I1").

MinhHaDuong:
## /gaze phase-6 verify-gate — PR #519 (ticket 0296) — round 1

**Verdict: REROLL**

### Exit criteria (tickets/0296-live-worktree-identity-guard.erg)

| # | Criterion | Status | Evidence |
|---|---|---|---|
| 1 | Guard script + hook registration merged | ADDRESSED | `scripts/guard-worktree-identity.sh` added, `settings.json` +10 lines (Bash matcher, `guard-worktree-identity.sh`, timeout 5) — commit 8a28d3d |
| 2 | Tests cover I1 replay, false-positive path, and the `git -C` whitelist | ADDRESSED-BUT-INSUFFICIENT | `tests/test_guard_worktree_identity.sh` has a case for each named item literally (I1 replay: 3 cases; false-positive: 2 cases; `git -C` whitelist: 1 case, `git -C $PRIMARY commit -m x`). The letter of the criterion is satisfied. But the whitelist test only exercises a **self-contained** `git -C ... commit` invocation — never a compound/chained command — so it does not catch the guard's actual security defect (below). The exit criterion as worded is met; the guard it was meant to validate is not sound. |
| 3 | `make check` passes | ADDRESSED | Re-ran independently in this worktree: 735 passed, 1 warning, 21.97s, exit 0 |

### Verification checkboxes (ticket body)
- [x] Guard blocks mutation when rev-parse resolves outside the cwd worktree — true for the simple case tested
- [x] Guard passes normal in-worktree mutation (no false positives in a raid wave) — true for the tested case
- [x] `git -C` invocations pass through — true, but the mechanism (whole-string substring match) is unscoped and this is the blocking defect
- [x] Read-only git (status, log, diff) is never blocked — true

### Unresolved review comment (review-pr, posted to this PR)

**Blocking finding, UNRESOLVED** — no fix commit exists on the branch beyond 8a28d3d and 457f0ae. The `git -C` whitelist (`grep -qP '\bgit\s+-C\s'`, guard-worktree-identity.sh) matches the substring anywhere in the whole command, not scoped to the git invocation it's meant to exempt. A compound command launders an unrelated `-C` past the whitelist and lets a trailing mutating verb through unchecked — reproducing incident I1 itself.

I independently reproduced this in a fresh fixture in this worktree (not just relaying the review-pr/adherence claims):

```
FIX=$(mktemp -d); PRIMARY="$FIX/primary"
git -C "$PRIMARY" init -q && git -C "$PRIMARY" commit -q --allow-empty -m init
BROKEN="$PRIMARY/.claude/worktrees/t001"; mkdir -p "$BROKEN"

CMD='git -C '"$PRIMARY"' status && git commit -m pwned --allow-empty'
# guard verdict:
printf '{"tool_name":"Bash","tool_input":{"command":%s},"cwd":%s}' \
  "$(printf '%s' "$CMD" | jq -Rs .)" "$(printf '%s' "$BROKEN" | jq -Rs .)" \
  | bash scripts/guard-worktree-identity.sh
# exit code: 0  (ALLOWED)

# actual execution from $BROKEN:
( cd "$BROKEN" && eval "$CMD" )
git -C "$PRIMARY" log --oneline -3
# HEAD before: 9a55cb6 init
# HEAD after:  a119d46 pwned   9a55cb6 init
```

The guard returned exit 0 and the mutation landed on the primary tree — the exact I1 fall-through this guard exists to close, via its own escape hatch.

### Adherence

Phase-2/3/4 adherence returned `FAIL` with one `blocking` finding (the same `-C` whitelist bypass, independently reproduced by that agent). Any blocking adherence violation forces reroll per gate policy. My own independent reproduction above confirms the finding rests on verified evidence, not relayed claims.

### Scope check

`git log origin/main..origin/t0296 --stat` (two-dot): `scripts/guard-worktree-identity.sh` (+77), `settings.json` (+10), `tests/test_guard_worktree_identity.sh` (+136), `tickets/0302-extract-shared-worktree-identity-predica.erg` (+53). All four files trace to ticket 0296's exit criteria or the legitimate 0302 follow-up filing. No scope overflow.

### Rationale

Exit criterion 2's literal wording is satisfied by the shipped test suite, but the guard it validates has a confirmed, independently-reproduced security defect that lets a compound command bypass the `git -C` whitelist and fall through to the exact primary-repo mutation (I1) the guard was built to prevent. The review-pr blocker (whitelist scoping) and the adherence blocking finding both point at the same root cause and remain unresolved on the branch — no fix commit exists beyond 8a28d3d/457f0ae. The verb-allowlist coverage gap and quoting/alias evasion (review-pr blockers 2-3) compound the risk but are secondary to the whitelist-scoping defect, which alone is sufficient to reroll given the guard's stated purpose and the fact that it runs on every Bash call in every session.

Reroll-bump logged: `tickets/0296-live-worktree-identity-guard.erg` (round 1).

---

```yaml
verdict: REROLL
round: 1
```

MinhHaDuong:
## /gaze phase-6 verify-gate — PR #519 (ticket 0296) — round 2 (FINAL)

**Verdict: APPROVED**

### Round-1 blocker (the `-C` whitelist compound-command bypass) — VERIFIED FIXED

Independently reproduced the exact round-1 exploit against commit 26dbf55, from a real fixture (unregistered dir under `.claude/worktrees/<name>` inside a real repo, `rev-parse` walking up to the primary root — not a mock):

```
cmd: git -C $PRIMARY status && git commit -m x
cwd: $PRIMARY/.claude/worktrees/deregistered
guard exit: 2 (BLOCKED)  — was 0 (ALLOWED) before the fix
```

Also tried 9 variants of my own beyond the fix agent's own tests, all correctly BLOCKED:
- 3-way chain with bare mutate last (`-C ... && -C ... && commit`)
- mutating verb placed BEFORE the `-C` segment (segment order doesn't matter)
- `;`, `|`, and `&` separators in place of `&&`
- `-C` in a non-adjacent flag position (`git status -C X && git commit`)
- subshell wrapping (`(git -C X status) && git commit`)
- quoted verb inside a compound (`git -C X status && git "commit" -m x`)
- the literal I1 verb (`git -C X status && git checkout -B main`)

Sanity check confirms the whitelist still correctly ALLOWS a self-contained `git -C X commit` (exit 0) — the fix didn't overcorrect into blocking the legitimate idiom.

### Test suite and full check

- `bash tests/test_guard_worktree_identity.sh`: 24 PASS, 0 FAIL, exit 0. Confirmed the new compound-command regression cases exist and assert the right thing (lines 94-104: `_assert_blocked "compound: git -C status && bare git commit"`, `_assert_blocked "compound: read-only git ; bare erg close"`, `_assert_allowed "compound: two git -C segments stay whitelisted"`).
- `make check`: 735 passed, 1 warning (pre-existing unrelated `AsyncMockMixin` coroutine warning in test_beat.py, not touched by this PR), exit 0.

### Round-1 secondary findings — disposition

- **Verb-allowlist gaps**: mostly closed. Verified BLOCKED: cherry-pick, revert, apply, am, restore, filter-branch, gc, pull, worktree remove, branch -D, tag -d, config, fetch --prune. Verified still-ALLOWED (correctly, by design — read-only forms): `git branch -vv` / `git tag -l`.
  **Residual gap, not fully closed**: `git update-ref` and `git symbolic-ref` were named explicitly in round 1's blocker list but are absent from the current `GIT_MUT` regex — both still return exit 0 (ALLOWED) from the broken-worktree fixture. These are plumbing-level ref-moves of comparable severity to `checkout -B` (the I1 verb). Not a blocker for this round: unlike the whitelist-scoping bug (which ordinary, sanctioned usage — `git -C X ...` chained with other commands — triggers by accident), these two commands are rarely-used plumbing that requires deliberate invocation to exploit. Recommend a fast-follow ticket to add `update-ref|symbolic-ref` to the regex.
- **Quoted-verb evasion** (`git "commit"`): fixed, verified blocked standalone and inside a compound.
- **Alias evasion** (`git ci` with `alias.ci=commit`): still bypasses (exit 0), same as round 1. Explicitly scoped out in the code with a documented rationale (a static regex cannot see through a git alias without invoking git itself; ticket 0302 named as the place to revisit). Accepted as UNRESOLVED-WITH-RATIONALE — the rationale is sound (this is a fundamental limitation of string-matching, not an oversight), and alias evasion requires the operator to have pre-configured the alias, unlike the compound-command idiom which is everyday usage.
- **New finding (mine, not in round 1)**: variable-substitution evasion (`v=commit; git $v -m x`) also bypasses (exit 0). Same class as the accepted alias/command-substitution gaps (dynamic verb resolution invisible to static regex) — not explicitly named in the code's scoped-out comment, but the same rationale applies. Recommend folding this into the same follow-up as the alias/update-ref items.

### Ticket 0296 exit criteria

| # | Criterion | Status | Evidence |
|---|---|---|---|
| 1 | Guard script + hook registration merged | ADDRESSED | `settings.json` +10 lines (unchanged since round 1), guard script now 121 lines |
| 2 | Tests cover I1 replay, false-positive path, and the `git -C` whitelist | ADDRESSED | Compound-command cases now present and passing (c2 block, lines 94-104) — closes round 1's "ADDRESSED-BUT-INSUFFICIENT" |
| 3 | `make check` passes | ADDRESSED | 735 passed, exit 0 |

Verification checkboxes: all 4 confirmed true, including "`git -C` invocations pass through" now correctly scoped per-segment rather than over the whole string.

### Scope check

`git log origin/main..origin/t0296 --stat`: `scripts/guard-worktree-identity.sh`, `settings.json`, `tests/test_guard_worktree_identity.sh`, `tickets/0302-...erg` — same four files as round 1; the round-1 fix itself only touched the guard script (+59/-8) and its test file (+31). No scope overflow.

### Rationale

The primary, blocking defect from round 1 — the `-C` whitelist matching a substring anywhere in the whole command instead of being scoped to its own segment — is fixed and independently verified against the exact round-1 exploit plus 9 additional variant attempts of my own devising, none of which reopen the hole. The full test suite and `make check` are green. Two residual gaps remain (update-ref/symbolic-ref verb coverage, variable-substitution evasion) but are of materially lower practical severity than the fixed bug: they require deliberate, non-idiomatic git usage to trigger, unlike the fixed bug which ordinary sanctioned cross-tree usage (`git -C X ... && ...`) tripped by accident. Round 3 being forbidden, and these residuals being mechanical additions to an already-improving verb list (not an architectural rework, and not requiring human arbitration), they don't rise to ESCALATE. Recommend a fast-follow ticket for `update-ref`/`symbolic-ref`/variable-substitution, alongside the already-tracked alias-evasion gap.

```yaml
verdict: APPROVED
round: 2
```

PR #529, opened 2026-07-13
naming(0298): collision-resistant worktree names t{N}-{pid}

MinhHaDuong:
## What

Collision-resistant worktree naming — `t{N}-{pid}` — so two sessions on the same ticket resolve to distinct worktree paths (incident I2, aedist 2026-06-11/12). Child of tracking ticket 0252 (model C, ratified 2026-07-13).

## Changes

- **rules/workflow.md** naming table: the three `t{N}` cells become `t{N}-{pid}`, plus one sentence naming the `$$` session discriminator. Legacy bare `t{N}` remains valid.
- **skills/hunt/SKILL.md** step 3: the ownership predicate accepts basename exactly `t$ARGUMENTS` (legacy) OR beginning with `t$ARGUMENTS-` (suffixed) at all four comparison points; the `EnterWorktree` request now uses `t$ARGUMENTS-$$`. The legacy exact-match branch is kept so the invariant "legacy `t{N}` worktrees remain recognized" holds.
- **tests/test_beat.py**: new `TestHarnessWorktreeRe` regression lock.

## beat.py needed no edit

`_HARNESS_WORKTREE_RE = re.compile(r"^(agent- |explore-|t\d|review-)")` is an unanchored prefix match, so it already matches `t0296-1234a`. The new test pins that behavior (matches both `t0296` and `t0296-1234a`) so a future re-anchoring cannot silently drop the suffixed form and strand its locked worktrees during housekeeping.

## Verification

- `tests/test_beat.py::TestHarnessWorktreeRe` + `tests/test_hunt_worktree_ownership.py` — 7 passed (ticket 0294's guard test unchanged and green).
- `make check-fast` and `make check` — pass.

**Ticket:** tickets/0298-collision-resistant-worktree-naming.erg

### DIFF
--- a/rules/workflow.md
+++ b/rules/workflow.md
@@ -12,10 +12,12 @@ The hook handles worktree entry automatically. When naming the worktree (if prom
 | Context | Worktree name | Phase |
 |---------|---------------|-------|
 | Fresh conversation, no ticket | `explore-{topic}` | `[→ Imagine]` |
-| Ticket reference but no branch | `t{N}` | `[→ Plan]` |
-| `/hunt N` | `t{N}` | `[→ Execute]` |
-| Active feature branch + open MR | `t{N}` | `[→ Execute]` |
+| Ticket reference but no branch | `t{N}-{pid}` | `[→ Plan]` |
+| `/hunt N` | `t{N}-{pid}` | `[→ Execute]` |
+| Active feature branch + open MR | `t{N}-{pid}` | `[→ Execute]` |
 | MR review | `review-{N}` | `[→ Verify]` |
+
+The `{pid}` suffix is a short session discriminator (e.g. `$$`) so two parallel sessions on the same ticket land in distinct worktree paths; legacy bare `t{N}` names remain valid.
 
 After `EnterWorktree` succeeds, emit the phase label on its own line (e.g. `[→ Execute]`) so the user sees which Five-Claws claw is active.
 
--- a/skills/hunt/SKILL.md
+++ b/skills/hunt/SKILL.md
@@ -25,8 +25,8 @@ argument-hint: <ticket-id>
      via review like any other change. Do not stop with an uncommitted-to-`main` or unpushed close.
 3. Enter the ticket's **own** worktree, or confirm the spawner already gave you one.
    A worktree is **owned** when either its basename is `t$ARGUMENTS` or begins with
-   `t$ARGUMENTS-` (the collision-resistant suffixed form, e.g. `t$ARGUMENTS-$$`), OR its
-   basename matches `agent-*` (the orchestrator created it for this agent session) AND
+   `t$ARGUMENTS-` (the collision-resistant suffixed form — see the `EnterWorktree`
+   call below), OR its basename matches `agent-*` (the orchestrator created it for this agent session) AND
    `git status --porcelain` prints nothing — the session started inside it and its tree
    is clean (that is what `Agent(isolation:"worktree")` produces). Both conditions must
    hold: an `agent-*` basename over a dirty tree is not proof of ownership. A shared or
@@ -35,7 +35,7 @@ argument-hint: <ticket-id>
    it does not own (2026-06-11: a hunt inherited an orchestration session's worktree,
    rebased its branch, and stranded in-progress test edits there). Resolve ownership
    before the first branch-mutating command:
-   - Already inside an owned worktree — basename `t$ARGUMENTS` or `t$ARGUMENTS-<pid>`, or an `agent-*`
+   - Already inside an owned worktree — basename `t$ARGUMENTS` or beginning with `t$ARGUMENTS-`, or an `agent-*`
      one from an `Agent(isolation:"worktree")` spawn whose `git status --porcelain` is
      empty — means you are isolation-confirmed; proceed to step 4. When such a spawned
      agent has its `EnterWorktree` rejected with "already in a worktree", that rejection
--- a/skills/hunt/SKILL.md
+++ b/skills/hunt/SKILL.md
@@ -24,7 +24,8 @@ argument-hint: <ticket-id>
      merge request (step 10). Skip the test/implement steps (5–9); the merge request lands the close
      via review like any other change. Do not stop with an uncommitted-to-`main` or unpushed close.
 3. Enter the ticket's **own** worktree, or confirm the spawner already gave you one.
-   A worktree is **owned** when either its basename is exactly `t$ARGUMENTS`, OR its
+   A worktree is **owned** when either its basename is `t$ARGUMENTS` or begins with
+   `t$ARGUMENTS-` (the collision-resistant suffixed form, e.g. `t$ARGUMENTS-$$`), OR its
    basename matches `agent-*` (the orchestrator created it for this agent session) AND
    `git status --porcelain` prints nothing — the session started inside it and its tree
    is clean (that is what `Agent(isolation:"worktree")` produces). Both conditions must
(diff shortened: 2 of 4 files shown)


MinhHaDuong:
## Multi-perspective review — PR #529 (t0298, collision-resistant worktree naming)

Panel: correctness, consistency, scope, test-quality (4 parallel foreground reviewers). Risk class: standard (prose + one regression-lock test touching a shared conventions doc and a widely-used skill's ownership contract). `make check` re-run in this review worktree at commit `5fc8fff`: **passed, exit 0**.

**Overall verdict: APPROVE** (all four perspectives independently approved; no blocking findings).

---

### Correctness — approve, high confidence

- **(a) Boolean composition** — `skills/hunt/SKILL.md:27-32` parses as `(exact "t$ARGUMENTS" OR prefix "t$ARGUMENTS-") OR (agent-* AND clean-tree)`, the intended grouping. The "Both conditions must hold" sentence explicitly scopes the clean-tree requirement to the `agent-*` clause only, so a suffixed worktree with a dirty tree is correctly unaffected.
- **(b) Prefix-collision** — checked by explicit string comparison: `"t290-5678".startswith("t29-")` is **False** (diverges at the literal `-` vs `0`), and the reverse (`"t29-1234".startswith("t2-")`) is also **False**. Because the pattern is `t<digits>-` with a literal terminating hyphen, no ticket ID's prefix can spuriously match another's. No collision risk.
- **(c) Legacy-form reachability** — confirmed present at all three touch points: the "owned" definition (`SKILL.md:27`), the "Already inside an owned worktree" bullet (`SKILL.md:38`), and the "Confirm with basename" sentence (`SKILL.md:48-49`) — each lists the bare `t$ARGUMENTS` form alongside the new suffixed form.
- `scripts/beat.py:244` regex (`^(agent-|explore-|t\d|review-)`, unanchored, used with `.match()`) already matches both `t0296` and `t0296-1234a` with zero code change — matches the PR's claim.

- `consider:` `SKILL.md:31`'s "Both conditions must hold" is grammatically loose (three disjuncts total, not two) but the adjacent example fully disambiguates which two are meant. Cosmetic only.

### Consistency — approve, high confidence

Read `rules/workflow.md` end to end. The new sentence (line 20, "the `{pid}` suffix is a short session discriminator…") and the pre-existing "Parked-cwd trap" section (lines 38-48, ticket 0267) use compatible framings (name → path → basename) with no contradiction or redundancy. The automated ownership guard (`scripts/guard-worktree-identity.sh`) derives its expected name from the cwd string itself, so the `{pid}` suffix is transparent to it.

- `consider:` line 48's "must match the expected worktree/project name" already straddled exact-vs-prefix readings before this PR; now that the table shows an unpredictable-in-advance `{pid}` suffix, a *human* doing the manual check by hand could wonder which reading applies (the automated guard is unaffected). Pre-existing ambiguity this PR brushes against rather than causes — worth a one-clause clarification in a future ticket-0267 follow-up, not blocking here.

### Scope — approve, high confidence

- Ticket exit criterion "beat.py updated coherently" is met: the regex is an unanchored prefix match that already recognizes the suffixed form, and the new `TestHarnessWorktreeRe` class both proves this (ran it — passes against the unmodified regex) and documents why no source edit was needed. A no-op source edit would have been theater.
- `skills/raid/SKILL.md` Phase 5 launches `Agent(isolation:"worktree")` (producing `agent-<id>` paths) and its first in-agent action invokes `Skill(skill:"hunt", args:"<id>")` — raid contains no `t$ARGUMENTS`/`EnterWorktree`-naming string of its own; all `t{N}` construction lives in hunt, which this PR does update. The ticket's "Relevant files" listed raid defensively as a worktree-creation *site*; the diff's omission of it is confirmed a **non-issue**, not scope_overflow-in-reverse.
- No scope creep: diff footprint (`rules/workflow.md`, `skills/hunt/SKILL.md`, `tests/test_beat.py`) matches exactly what the naming change implies.

- `consider:` the ticket's own "Relevant files" list is now slightly stale (names `skills/raid/SKILL.md` and implies a `beat.py` source edit that didn't happen) — worth a one-line closing note on the ticket explaining why, so a future reader doesn't file a spurious bug for the "missing" raid edit.

### Test quality — verdict: genuine (partial) tripwire, not a mere change-detector; medium-high confidence

Two plausible future "tighten this regex" edits were tried against both test assertions:
- Full end-anchoring (`t\d+$`): breaks `test_matches_suffixed_ticket_name` (the trailing `-1234a` can't be consumed).
- Mandatory discriminator (`^t\d+-`): breaks `test_matches_legacy_bare_ticket_name` (no trailing hyphen).

`verifiable:` the test pair does catch the two most likely "make this regex stricter" edits — real regression-lock value, not a no-op assertion of already-obvious behavior.

- `consider:` no negative-case test exists for `_HARNESS_WORKTREE_RE` anywhere in the suite (e.g. `"trunk"`, `"foo-t0296"` should not match) — the regex gates which admin-dir names get treated as harness worktrees for lock-cleanup, so over-matching risk is real and currently untested. Cheap to add in a follow-up; not required for this PR's stated exit criteria.
- Docstring quoted and checked: accurately scoped, doesn't overclaim collision-rejection coverage.

---

No blocking findings from any perspective. Recommend proceeding to verify-gate.

MinhHaDuong:
## verify-gate: APPROVED (round 1)

### Exit criteria

- [x] **Naming table, hunt/raid, and beat.py updated coherently** — ADDRESSED.
  - `rules/workflow.md:15-17,20` (commit 5fc8fff): three naming-table rows changed to `t{N}-{pid}`, plus the discriminator explanation and the "legacy bare `t{N}` names remain valid" clause.
  - `skills/hunt/SKILL.md:27-49` (commits 5fc8fff, 7f5e494): step 3's ownership predicate now accepts basename `t$ARGUMENTS` (legacy, exact) OR beginning with `t$ARGUMENTS-` (suffixed) at all four comparison points; the `EnterWorktree` call site uses `t$ARGUMENTS-$$`.
  - `skills/raid/SKILL.md` correctly absent from the diff — re-verified by grep: raid never emits `t{N}` names; it delegates via `Agent(isolation:"worktree")`, which produces `agent-<id>` paths, a separate ownership clause already in hunt/SKILL.md.
  - `scripts/beat.py:244` needed no edit: `_HARNESS_WORKTREE_RE = re.compile(r"^(agent-|explore-|t\d|review-)")` is an unanchored prefix match on `t\d`, already matching `t0296-1234a`. Regression-locked by `tests/test_beat.py::TestHarnessWorktreeRe` (commit 0b03491, lines 2478-2492) — both `test_matches_legacy_bare_ticket_name` and `test_matches_suffixed_ticket_name` pass.
  - Grep sweep (re-run independently this gate) found no other consumer parsing worktree names by exact match; `scripts/trace-stats.py::strip_worktree` strips an unrelated `--claude-worktrees-...` path-encoding suffix, not the `t{N}` naming convention.

- [x] **`make check` passes** — ADDRESSED. Reconfirmed in this worktree at HEAD `7f5e494`: `747 passed, 1 warning in 24.05s`, exit 0.

**Invariant** "Legacy `t{N}` worktrees remain recognized" — held: exact-match branch kept in hunt/SKILL.md alongside the new prefix branch; `TestHarnessWorktreeRe::test_matches_legacy_bare_ticket_name` passes.

### Review comments disposition

1. **Agent A (adherence)** — PASS, nit only (workflow.md `review-{N}` row unsuffixed — correctly out of scope; raid/SKILL.md absence independently confirmed correct). No action needed.

2. **Agent B (`/review`) — `$$` not tool-schema-resolvable.** Independently re-verified, not deferred. `EnterWorktree`'s own `name` parameter schema: *"Each '/'-separated segment may contain only letters, digits, dots, underscores, and dashes; max 64 chars total."* `$` is indeed excluded, so a literal `t0298-$$` name would fail input validation — the premise is technically correct.

   Disposition: **non-blocking, resolved by re-reading**, not a structural defect. Two reasons:
   - The claimed worst case ("every parallel session would produce the IDENTICAL name" via silent leniency) does not hold: the schema enforces the character class strictly with no leniency, so the actual failure mode is a loud, explicit `InputValidationError`, not silent acceptance of a colliding literal string. An agent hitting that error self-corrects (resolve an actual PID, retry) rather than silently producing duplicate paths — which is exactly the incident (I2) this ticket exists to prevent.
   - Context signals this is intentional shorthand, not a literal call recipe: `rules/workflow.md:20` hedges with "e.g. `$$`", and hunt/SKILL.md's own parenthetical calls it a "session-**PID**" — i.e., an actual numeric value obtained by running a shell command (the same skill file already embeds live-resolved bash a few lines later: `` basename "$(git rev-parse --show-toplevel)" ``, which the agent executes rather than types literally).
   - This is a legitimate documentation-clarity nit (spelling out "run `bash -c 'echo $$'` to get the numeric PID, then use `t$ARGUMENTS-<pid>`" would remove the ambiguity entirely) but does not make the exit criteria or the "distinct worktree paths" verification property unsatisfiable as designed.

   Recommend a fast-follow doc tweak, not a reroll.

3. **Agent C (`/review-pr` panel)** — APPROVE, tagged minors only; independently corroborated `TestHarnessWorktreeRe` as a genuine partial-regression lock (tried two plausible future regex edits, each breaking one assertion). No action needed.

4. **`/simplify`** — one wording-consistency commit (7f5e494), no behavior change, correctly out of scope for the `$$`-resolvability question.

### Verdict

All exit criteria ADDRESSED, invariant held, `make check` green (747 passed), no unresolved correctness defect, no scope overflow. **APPROVED.**

PR #538, opened 2026-07-13
hunt step 3: spell out obtaining the session PID for the worktree name

MinhHaDuong:
## What

Hunt step 3 named the fresh worktree `t$ARGUMENTS-$$` but never said how to resolve `$$`. `EnterWorktree`'s name schema rejects `$` characters, so the model would pass a literal `$` and hit a loud `InputValidationError`. This spells out resolving the session PID with `bash -c 'echo $$'` and substituting the number (e.g. `t$ARGUMENTS-12345`) before the call.

Documentation fast-follow recommended by the #529 gaze verdict (which ruled the gap non-blocking: loud error, not a silent collision).

## Changes

- `skills/hunt/SKILL.md` — step 3 `EnterWorktree` bullet now shows the literal PID-resolution command.
- `tests/test_hunt_worktree_ownership.py` — adds `test_pid_discriminator_obtainable` asserting step 3 shows `bash -c 'echo $$'`.
- ticket log line.

## Verification

- `make check` exits 0 (754 passed).
- Pre-PR `/verify-adherence`: PASS (all phases clean).

**Ticket:** tickets/0309-hunt-step-3-spell-out-how-to-obtain-the.erg

### DIFF
--- a/skills/hunt/SKILL.md
+++ b/skills/hunt/SKILL.md
@@ -42,9 +42,13 @@ argument-hint: <ticket-id>
      **is** the confirmation only if the current tree passes the same clean `agent-*`
      check; a rejection over a dirty or non-`agent-*` tree means STOP — the worktree is
      not owned — not an error to proceed through.
-   - Otherwise call `EnterWorktree` with name `t$ARGUMENTS-$$` (the `$$` session-PID
-     suffix keeps parallel sessions on the same ticket in distinct worktree paths),
-     even if the session already sits inside some other, unowned worktree.
+   - Otherwise call `EnterWorktree` with name `t$ARGUMENTS-<pid>`, where `<pid>`
+     is the session PID — a discriminator that keeps parallel sessions on the same
+     ticket in distinct worktree paths. Resolve it to a literal first (the
+     `EnterWorktree` name schema rejects `$` characters): run `bash -c 'echo $$'`
+     and substitute the number, e.g. pass `t$ARGUMENTS-12345`, not the raw
+     `t$ARGUMENTS-$$`. Call `EnterWorktree` even if the session already sits inside
+     some other, unowned worktree.
    Confirm with `basename "$(git rev-parse --show-toplevel)"`: it must be `t$ARGUMENTS`,
    begin with `t$ARGUMENTS-`, or be the `agent-*` worktree of this session (rules/git.md § anchor branch-mutating git
    across a forked-skill boundary). Ad hoc orchestrators should not hand-type this
--- a/tests/test_hunt_worktree_ownership.py
+++ b/tests/test_hunt_worktree_ownership.py
@@ -77,6 +77,21 @@ def test_rejection_as_confirmation_documented():
     )
 
 
+def test_pid_discriminator_obtainable():
+    """Step 3 spells out a literal, executable way to obtain the `$$` discriminator.
+
+    Ticket 0309: `EnterWorktree`'s name schema rejects `$` characters, so the
+    model must resolve the session PID to a literal before the call. Step 3 must
+    show how (e.g. `bash -c 'echo $$'`), not just reference the bare `$$` token.
+    """
+    step3 = step3_text()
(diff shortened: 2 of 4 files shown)


MinhHaDuong (commented):
Reviewed as PR-panel Agent C (correctness + consistency), scope: docs-only fast-follow (26 lines, 3 files).

**Correctness** — approve. The new step-3 text correctly resolves ticket 0309's bug: `EnterWorktree`'s name schema rejects `$`, so the fix has the model run `bash -c 'echo $$'` first and substitute the printed literal (e.g. `t0309-12345`) rather than passing the raw `t$ARGUMENTS-$$` token to the tool call. `bash -c 'echo $$'` correctly returns the invoked subshell's own PID — a valid, simple way to mint a discriminator. Each Bash tool call may spawn a fresh shell process, so this value is not guaranteed stable across separate calls within one session, but that doesn't undermine the stated purpose: the PID only needs to be minted once and then reused verbatim in the one `EnterWorktree` name and the one ownership-confirmation check that follow it in the same step — nothing in the skill asks the model to recompute it later and compare. The new test `test_pid_discriminator_obtainable` asserts the literal string `bash -c 'echo $$'` is present in the step-3 text, which faithfully matches what's documented and what the ticket's exit criterion asks for.

**Consistency** — consider: `rules/workflow.md:20` still describes the `{pid}` suffix generically as "e.g. `$$`" without the resolve-to-literal caveat this PR adds to hunt/SKILL.md. This doesn't contradict the fix (workflow.md is describing the naming convention in the abstract, not instructing a literal `EnterWorktree` call), so it's not a blocker, but it's a candidate for the same fast-follow treatment if another skill/rule ever hand-types an `EnterWorktree` call site from that text alone. `grep` across skills/ and rules/ found no other call site referencing `$$`/`t$ARGUMENTS` that would need the same fix.

Ran the touched test file directly (`pytest tests/test_hunt_worktree_ownership.py`): 6 passed.

Verdict: approve.

MinhHaDuong:
## /gaze actions

round: 1
adherence: PASS — 0 blocking
review-pr: 1 comment posted
simplify: 0 fixes applied (2 advisory skips)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- hunt step 3 shows a literal, executable way to obtain the discriminator: ADDRESSED — commit a5d59cf, skills/hunt/SKILL.md:47-48 ("Resolve it to a literal first ... run `bash -c 'echo $$'` and substitute the number, e.g. pass `t$ARGUMENTS-12345`, not the raw `t$ARGUMENTS-$$`").
- `make check` passes (test_hunt_worktree_ownership.py still green): ADDRESSED — ran `uv run python -m pytest tests/test_hunt_worktree_ownership.py -q` in the worktree: 6 passed, including the new `test_pid_discriminator_obtainable`.

Unresolved review comments: none — Agent B's untagged nit (unchecked exit-criteria checkboxes) is informational/cosmetic, not a correctness blocker; Agent C's `consider:` minor (rules/workflow.md:20 generic `$$` phrasing) is informational by tag rule, a candidate follow-up not a contradiction with this diff.
Unresolved simplify: none — 2 advisory findings skipped with validated rationale (wording-tighten judged false positive; altitude relocation to rules/workflow.md judged out-of-scope and duplicate of the panel `consider:`).
Adherence: PASS

Rationale:
Both exit criteria are directly verifiable in the diff and were independently confirmed rather than taken on faith: the literal resolution instruction is present at skills/hunt/SKILL.md:47-48, and the test suite (including the new test_pid_discriminator_obtainable) passes cleanly. Scope check (`git log origin/main..HEAD --stat`) shows exactly three files touched — skills/hunt/SKILL.md, tests/test_hunt_worktree_ownership.py, and the ticket file itself — all traceable to the ticket; no scope overflow. No human comments outstanding. CI is 9/9 green. Round 1, no reroll needed.

telemetry: wall=~360s agents=8 tokens=~344k cost~=na

PR #542, opened 2026-07-13
ticket(0302): keep worktree-identity predicate inline, add drift guard

MinhHaDuong:
## Decision (0302 is a DECISION ticket)

**Keep the worktree-identity predicate inline — do NOT extract `scripts/lib/worktree-identity.sh`.**

### Diff of the copies
Of the sites carrying a worktree predicate, only **two** share the exact identity check ("cwd under `.claude/worktrees/<name>` AND git's toplevel is exactly that worktree, not the primary root"), and they are byte-identical:
- `scripts/pretooluse-worktree-path-guard.sh` `_in_worktree()` — advisory hook, **fail-open**, live `pwd -P` + bare `git rev-parse`.
- `skills/merge/erg-pr-merge` `in_worktree()` — standalone skill, boolean, same body.

The other sites do **not** share this predicate:
- `scripts/guard-worktree-identity.sh` and `scripts/guard-commit-on-main.sh` read the hook-JSON `.cwd` through `git -C`, are **fail-closed** integrity guards, and (for guard-commit-on-main) use the `--git-common-dir` parent technique rather than the `.claude/worktrees/<name>` string check.
- `scripts/block-pr-merge-in-worktree.sh` uses a deliberately different linked-worktree predicate (git-dir vs git-common-dir) — left untouched.

### Rationale
No single helper fits all sites without changing an input model (JSON `.cwd` vs live shell cwd) or a fail-open/fail-closed contract — which 0302's invariant forbids. A `source scripts/lib/…` into the fail-open hook, or across the scripts/↔skills/ boundary into the self-contained skill, would under `set -euo pipefail` turn a missing lib into a nonzero abort, flipping the very semantics the invariant protects. So extraction would couple three failure behaviours into one file for the sake of two truly-identical copies — the YAGNI cost the ticket flagged, and it exceeds the duplication it removes.

Instead of coupling, the two identical copies are pinned by a cheap adherence test (`tests/test_worktree_identity_predicate_inline.py`) that fails if they drift — the DRY benefit (no silent divergence) without the coupling cost. This complements the 0310 ratchet (which forbids regressing either copy to the weak `-f`/`.git` file-test).

### Changes
- New `tests/test_worktree_identity_predicate_inline.py` — asserts the two function bodies stay byte-identical (RED-proven on a real body change and an inline-comment drift).
- Rationale comments recorded at both sites; the now-false "0302 will unify these three copies" promise in the pretooluse guard is replaced with the actual decision.
- Decision logged in the ticket.

Invariant preserved: no call site's fail-open/fail-closed semantics changed (this PR touches only comments + a test). `make check` passes (770 tests).

**Ticket:** tickets/0302-extract-shared-worktree-identity-predica.erg

### DIFF
--- a/scripts/pretooluse-worktree-path-guard.sh
+++ b/scripts/pretooluse-worktree-path-guard.sh
@@ -19,7 +19,13 @@ hook_cwd=$(echo "$input" | jq -r '.cwd // empty' 2>/dev/null || true)
 # scripts/guard-worktree-identity.sh: the cwd must sit under
 # `.claude/worktrees/<name>` AND `git rev-parse --show-toplevel` must resolve to
 # a tree whose basename is that `<name>` and is not the primary root itself.
-# (0302 will unify these three copies into a shared helper.)
+# Ticket 0302 decided to keep this copy inline rather than extract a shared
+# `scripts/lib/worktree-identity.sh`: the integrity guards read hook-JSON `.cwd`
+# through `git -C` and are fail-closed, so no one helper fits all sites without
+# changing an input model or a failure contract, and sourcing a lib into this
+# fail-open hook would flip its semantics on a missing file. This copy and the
+# one in `skills/merge/erg-pr-merge` are held in lockstep by
+# tests/test_worktree_identity_predicate_inline.py instead.
 _in_worktree() {
     local cwd top name prefix
     cwd=$(pwd -P)
--- a/skills/merge/erg-pr-merge
+++ b/skills/merge/erg-pr-merge
@@ -31,7 +31,12 @@ die() { echo "erg-pr-merge: $*" >&2; exit 1; }
 # dir planted under the marker (git resolves to the enclosing repo, a different
 # basename) is correctly rejected. A PreToolUse hook fires once on the outer
 # script invocation and cannot front these internal git calls, so the check is
-# self-contained by design.
+# self-contained by design. Ticket 0302 kept this copy inline (not a shared
+# `scripts/lib/…` helper) for that reason: sourcing across the scripts/↔skills/
+# boundary would couple this standalone skill to the scripts tree and, under
+# `set -euo pipefail`, turn a missing lib into an abort. It stays byte-identical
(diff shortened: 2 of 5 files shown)


MinhHaDuong:
## /gaze actions

round: 1
adherence: PASS — 0 blocking
review-pr: 5 findings returned (0 blockers, 1 consider, 2 nofollow; 2 resolved in-run)
simplify: 1 fix applied (cfc0527 — clear assertion instead of StopIteration in _extract_body); 3 skipped with rationale (rationale-prose duplication is the ticket's deliverable; SITES-scan is scope expansion; author-token drift pre-existing)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- "Shared helper extracted OR short rationale recorded for why inline copies stay": ADDRESSED — rationale-recorded branch: ticket log line 2026-07-13T12:55Z; scripts/pretooluse-worktree-path-guard.sh:19-26 comment; skills/merge/erg-pr-merge:31-37 comment; tests/test_worktree_identity_predicate_inline.py docstring (all commit 0d0eafc). Invariant verified: diff shows comment-only changes to both shell predicates — no functional change to any call site's fail-open/fail-closed semantics.

Unresolved simplify: none
Adherence: PASS

Rationale:
The exit criterion is met via the record-rationale branch. Four artifacts state the decision: only 2 of 4 predicate sites share the exact identity check; the integrity guards are fail-closed and read hook-JSON .cwd; sourcing a shared lib would flip the hook's fail-open semantics or couple the standalone skill across the scripts/-skills/ boundary under set -euo pipefail. Drift protection is supplied by the new adherence test pinning the two byte-identical bodies (passes post-fix; make check 770 green). Scope check clean: 2 commits, 4 files, all traceable to the criterion. Non-blocking notes for a future touch: the ticket body's Context still describes the original 3-site inventory (the log line carries the corrected 4-site picture), and the SITES list is a hand-maintained enumeration — a third strong-predicate copy pasted elsewhere would not be auto-pinned (consider-class; the 0310 ratchet covers the weak-pattern side by tree scan).

telemetry: wall=~460s agents=8 tokens=480695 cost~=na

PR #544, opened 2026-07-13
erg-pr-merge: cross-check the ticket close claim against the branch tip

MinhHaDuong:
## Summary

`erg-pr-merge` parses the `**Ticket:**` close claim from the PR body and closes
it unconditionally; nothing verified the claim still matches the branch content.
Shared-worktree contention once force-pushed an unrelated changeset over a
branch (PR #522, 2026-07-13), leaving a stale `**Ticket:** 0268` claim that
would have closed the ticket while landing unrelated work. The only defense was
a human noticing and retitling the PR.

## What changed

New **Step 1.5** in `skills/merge/erg-pr-merge`, before the close loop:

- **Existence-at-tip (blocking, overridable).** Abort if a claimed ticket's
  file is absent from `tickets/` at the branch tip. This is *existence-at-tip*,
  not touched-by-diff — a claim on a ticket the diff never modifies is
  legitimate (close-by-convention on merge), so the diff is not required to
  touch the ticket file. Override with `ERG_PR_MERGE_ALLOW_MISSING_TICKET=1` to
  warn-and-skip the absent ticket instead of aborting.
- **Title/claim disagreement (non-blocking warn).** A title prefix `word(NNNN)`
  naming a ticket not among the close claims is a cheap tell of a repurposed
  branch. Warn only — title prefixes are subject references, not close claims.

The diff-touch tell the ticket floats (claim names N but the diff touches only
ticket M's file) is deliberately skipped, with a note in the script: a valid
convention close routinely touches no ticket file, so that signal both misses
the #522 case (its replacement diff touched no ticket) and false-positives on
legitimate closes.

## Design choice: abort vs warn

Existence-at-tip **aborts** by default (satisfies the exit criterion: a
force-replaced branch can no longer silently close the originally-claimed
ticket), with a documented override env var. The title tell only **warns**
(titles legitimately diverge from claims).

## Invariant

Happy path is behaviorally identical: when every claim resolves at the tip and
the title agrees, no new output and no new blocking. The title check adds one
read-only forge call, skipped entirely when there is nothing to close.

## Tests

`tests/test_erg_pr_merge.sh` cases 22-25 (red-first): absent-claim abort names
ticket + override and attempts no merge; override warn-and-skip; title
disagreement warns while the valid claim still closes; matching-title
anti-regression (no spurious warn). `make check` passes (the shell suite runs
under `test_bash_suites.py`).

**Ticket:** tickets/0312-erg-pr-merge-cross-check-the-ticket-clos.erg

### DIFF
--- a/skills/merge/erg-pr-merge
+++ b/skills/merge/erg-pr-merge
@@ -14,6 +14,9 @@
 #   0. Verify: on PR head branch, CI green, no conflicts
 #   1. Find ticket ID(s) from "Ticket: tickets/NNNN-..." lines in PR body only;
 #      title prefixes (chore(NNNN)) are subject references, not close claims
+#   1.5 Cross-check each claim against the branch tip: abort if the claimed
+#      ticket file is absent (force-replaced branch, PR #522); warn on a
+#      title/claim disagreement. Override: ERG_PR_MERGE_ALLOW_MISSING_TICKET=1
 #   2. erg close NNNN "autoclosed — PR #N" + git add + commit + push (always)
 #   3. Queue auto-merge via the forge CLI; falls back to watch-then-merge if auto-merge disabled
 #   4. If not in worktree: git checkout base && git pull --rebase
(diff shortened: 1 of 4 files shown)


MinhHaDuong:
## /review-pr panel — PR 544 (erg-pr-merge Step 1.5, ticket 0312)

Four perspectives (correctness, consistency, scope, red-team) ran in parallel. `bash tests/test_erg_pr_merge.sh` passes all 25 cases including the new 22-25. Scope and consistency both **approve** cleanly (no unrelated diffs, style matches the file's conventions, header/inline comments agree). Correctness and red-team independently converge on the same root defect from different angles — both are treated as blockers below.

### Blockers

1. **False positive: already-closed tickets under `tickets/closed/` trip the abort.** `git ls-tree -r --name-only HEAD -- tickets/` lists archived tickets as `tickets/closed/NNNN-slug.erg`; the existence check `grep -q "^tickets/${tid}-"` anchors immediately after `tickets/` and does **not** match the `closed/` prefix. Verified directly:
   ```
   echo "tickets/closed/0299-fixture.erg" | grep -q "^tickets/0299-"   # NOMATCH
   ```
   Any PR whose body re-references an already-closed ticket (rebase carrying a stale `Ticket:` line, copy-paste, a close-race where a sibling PR closed it first — the exact scenario `rules/git.md`'s "Merge-bounce recovery" section documents) gets a hard abort with a message asserting the branch "may have been force-replaced" — actively misleading, since nothing is wrong. This trains reviewers to reach for `ERG_PR_MERGE_ALLOW_MISSING_TICKET` reflexively, which erodes the guard for the case it's actually meant to catch. Not exercised by cases 22-25 (all use never-seeded IDs, none use a pre-closed one).

2. **The guard doesn't cover the literal #522 scenario it cites.** #522, per both the ticket and the PR's own code comment, was a branch whose *diff* got replaced while the stale ticket's *file* sat untouched and present in `tickets/`. Existence-at-tip only fires when the file is *absent*. Trace: file present → `TIP_TICKETS` still lists it → grep matches → `MISSING` stays empty → Step 1.5 does nothing. The code's own comment (lines 156-161) concedes this exact gap for the rejected diff-touch alternative ("misses the #522 case — its replacement diff touched no ticket"), but existence-at-tip has the identical blind spot for the identical reason. The only remaining signal is guard (b), the title check — non-blocking, and silently absent whenever the title lacks a `word(NNNN)` prefix (or coincidentally still names the stale ticket). None of cases 22-25 seed a present-but-unrelated ticket file, so this gap is untested and the ticket's exit criterion ("a repurposed/force-replaced branch can no longer silently close the originally-claimed ticket") is not met for the variant that actually happened.

Recommend resolving before merge, or explicitly narrowing the ticket's exit criterion to the file-deleted variant only (if the file-present variant is judged out of scope / not cheaply fixable) and documenting that narrowing in the code comment.

### Tagged minors

- consider: `[[ -n "${ERG_PR_MERGE_ALLOW_MISSING_TICKET:-}" ]]` treats any non-empty value — including `=0` or `=false` — as "enabled." A caller setting `=0` expecting "disabled" silently gets the override instead. Undocumented, untested.
- consider: no test covers a multi-ticket PR with one missing ID and one valid ID in the same body — behavior looks correct by inspection (die is unconditional on any non-empty MISSING) but unverified.
- consider: title regex `grep -oiP '\(\K\d{3,4}(?=\))' | head -1` takes the *first* parenthesized 3-4 digit group in the title — a title like `"Increase timeout (500ms) ticket(0312): ..."` would pick `500`, producing a spurious title/claim-disagreement warning. Low-probability shape, non-blocking (warn only), untested.
- nofollow: the ticket's added log entry is timestamped `2026-07-13T00:00Z`, earlier in the same day than its own `Created`/first-log timestamp `2026-07-13T10:23Z` — looks like a placeholder clock value; cosmetic only.
- nofollow: `ERG_PR_MERGE_ALLOW_MISSING_TICKET=1` set ambiently (shell profile, CI env) rather than per-invocation silently disables guard (a) for every future merge with no persistent record beyond that run's stderr — same class of risk as other override flags in this script, not new to this PR.

### Verdict

Consistency: approve. Scope: approve. Correctness: request-changes (finding 1 above). Red-team: request-changes (finding 2 above, quoting the code's own concession). Net: **request-changes** — the two blockers should be addressed (or the exit criterion narrowed with rationale) before merge.

MinhHaDuong:
## /gaze actions

round: 1
adherence: PASS — 0 blocking
review-pr: posted (request-changes; both claimed blockers adjudicated and refuted by the gate with live repros — see verdict)
simplify: 1 fix applied (commit 3068dc9: PR title read from the Step 0 PR_JSON fetch, second forge call removed; test stub updated; suite 30/30) | 1 skipped (predicate unification would change documented existence-at-tip semantics)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Repurposed/force-replaced branch can no longer silently close the claimed ticket: ADDRESSED — c61ca0f Step 1.5 existence-at-tip abort + title/claim warn, tests 22-25 green. Documented residual: neither guard fires on the literal #522 shape (ticket file present-but-untouched at a replaced tip, retitle without a (NNNN) prefix — verified against d218034); the ticket delegated the predicate choice and the PR body documents the diff-touch rejection, so this is conscious scope-narrowing, not a silent gap.
- erg-pr-merge flags an inconsistent close claim: ADDRESSED — cases 22 and 24.
- Happy path unchanged; override documented: ADDRESSED — case 25 anti-regression; ERG_PR_MERGE_ALLOW_MISSING_TICKET documented in-script and in the body.

Unresolved review comments: none blocking.
- Panel blocker 1 (already-closed ticket claim aborts) REFUTED as a regression: the pre-PR script also hard-aborts on that scenario (erg close 'no ticket found' under set -e, verified by running the origin/main revision against a fixture); this PR adds a working override where none existed.
- Panel blocker 2 (#522 present-but-untouched shape uncovered) confirmed real but consciously scoped — see exit-criterion residual above. Advisory follow-up: note in the ticket log that the title-tell requires the (NNNN) convention.
- consider-class nits (title regex vs zero-padded IDs incl. unpadded 'fix(312)'; die message says 'match the diff' though the guard is diff-independent; =0 treated as override-enabled; no multi-ticket one-missing test): informational, non-blocking.

Unresolved simplify: none
Adherence: PASS

Rationale:
Both verification checkboxes and the exit criterion are met with concrete evidence; the panel's two untagged blockers were each adjudicated by live repro (fixture repo for the pre-PR abort; the real #522 tree for the coverage gap) rather than taken at face value — one refuted, one dispositioned as a documented residual the ticket itself pre-authorized. Suite green 30/30 including the simplify commit.

telemetry: wall~=1100s agents=8 tokens~=594k cost~=na

/gaze note: token total exceeds the 500k warn threshold — slow/heavy run, completed within escalate bounds.

PR #545, opened 2026-07-13
ticket(0306): extend parked-cwd guard to Skill invocations

MinhHaDuong:
## Summary

Extends the parked-cwd guard from ticket 0267 (which covered only `EnterWorktree`) to cwd-dependent `Skill` invocations. A session that never calls `EnterWorktree` — a background job run in place, a non-worktree project — could invoke a repo-targeting skill (`/verify-adherence`, `/review-pr`, `/healthcheck`, `/molt`, …) from a parked, git-ignored runtime cwd with no defense. Those skills resolve their target repo from the session base cwd the same way `EnterWorktree` does.

## Matcher decision: blanket `Skill`

The guard reads only the JSON `.cwd` field, so it is matcher-agnostic — a blanket `Skill` PreToolUse matcher points at the same script. Chosen over a maintained per-skill list, which would rot as skills are added. No false positive: the guard exits 0 at a repo root, in a tracked subdir, or outside any repo, so memory sweeps (which read `~/.claude/projects` by absolute path from a real repo root) and non-git manuscript folders stay allowed. It denies only a git-ignored directory inside a repo — the actual parked-runtime bug.

## Script name kept

Renaming to `guard-parked-cwd.sh` would force churn on the existing EnterWorktree settings entry, the test HOOK path/grep, and workflow.md. Kept the name to keep the shared `settings.json` diff surgical (one added hook entry, no reformatting).

## Changes
- `settings.json` — one added `PreToolUse` entry: matcher `Skill` → the existing guard script.
- `scripts/guard-enterworktree-parked-cwd.sh` — header comment notes the widened, matcher-agnostic scope (behavior unchanged).
- `tests/test_guard_enterworktree_parked_cwd.sh` — parametrized `_rc`/`_assert` by tool name; added Skill deny/allow cases and a wiring ratchet for the new Skill matcher (red-first; 20/20 pass).
- `rules/workflow.md` — parked-cwd-trap paragraph now states the guard covers both surfaces.

## Verification
- `make check` passes.
- `/verify-adherence` PASS (one non-blocking pre-existing em-dash-density nit, addressed by not adding to it).
- Invariants held: the 0267 EnterWorktree deny path is unchanged; skills in non-git directories stay allowed.

**Ticket:** tickets/0306-extend-parked-cwd-guard-to-cwd-skills.erg

Related follow-up: tickets/0314-parked-cwd-guard-realpath-normalize-cwd.erg (panel finding: pre-existing symlink-traversal bypass)

PR #548, opened 2026-07-13
fix(guard): realpath-normalize cwd to close symlink-traversal bypass

MinhHaDuong:
## Summary

Closes the symlink-traversal bypass in the parked-cwd guard found by the PR #545 review panel (pre-existing since ticket 0267).

The guard resolved the repo toplevel with `git rev-parse --show-toplevel` (which canonicalizes symlinks) but left `$cwd` symlinked. A path reaching a git-ignored runtime dir through a symlink then made `git check-ignore "$cwd"` error "outside repository", which the guard treats as allow — so a parked runtime directory addressed via a symlinked path bypassed the deny on both guarded surfaces (EnterWorktree and Skill).

## Change

`scripts/guard-enterworktree-parked-cwd.sh`: normalize `$cwd` with `cd … && pwd -P` before the repo-membership and check-ignore probes. Resolution failure fails open (`|| exit 0`), preserving the guard's deny-only polarity — a normalization failure never flips an allow into a deny.

## Tests (red-first)

`tests/test_guard_enterworktree_parked_cwd.sh` gains 4 cases, confirmed RED before the fix:
- symlinked parked cwd denied (EnterWorktree + Skill)
- symlinked path to a legitimate repo root stays allowed (EnterWorktree + Skill)

The `settings.json` matcher (`EnterWorktree|Skill`) is untouched.

## Verification

- `make check` — pass
- `make lint` — pass
- `/verify-adherence` — PASS (mechanical, no findings)

**Ticket:** tickets/0314-parked-cwd-guard-realpath-normalize-cwd.erg

Related follow-up: tickets/0317-parked-cwd-guard-nested-git-repo-inside.erg (panel finding: nested-git-repo escape, pre-existing)

### DIFF
--- a/scripts/guard-enterworktree-parked-cwd.sh
+++ b/scripts/guard-enterworktree-parked-cwd.sh
@@ -27,7 +27,6 @@ cwd=$(echo "$input" | jq -r '.cwd // empty' 2>/dev/null || true)
 # deny (ticket 0314). Fail open if resolution fails (e.g. a race removed the
 # dir): a normalization failure must never flip an allow into a deny.
 cwd=$(cd "$cwd" 2>/dev/null && pwd -P) || exit 0
-[ -z "$cwd" ] && exit 0
 
 # Not inside a git repo: the tool fails on its own (or hook-delegated
 # VCS-agnostic mode handles it) — nothing for this guard to judge.
--- a/scripts/guard-enterworktree-parked-cwd.sh
+++ b/scripts/guard-enterworktree-parked-cwd.sh
@@ -20,6 +20,15 @@ cwd=$(echo "$input" | jq -r '.cwd // empty' 2>/dev/null || true)
 [ -z "$cwd" ] && exit 0
 [ -d "$cwd" ] || exit 0
 
+# Normalize away symlinks before the repo-membership and check-ignore probes.
+# A path that reaches the repo through a symlink otherwise makes
+# `git check-ignore "$cwd"` error "outside repository", which the guard treats
+# as allow — letting a parked runtime dir addressed via a symlink bypass the
+# deny (ticket 0314). Fail open if resolution fails (e.g. a race removed the
+# dir): a normalization failure must never flip an allow into a deny.
+cwd=$(cd "$cwd" 2>/dev/null && pwd -P) || exit 0
+[ -z "$cwd" ] && exit 0
+
 # Not inside a git repo: the tool fails on its own (or hook-delegated
 # VCS-agnostic mode handles it) — nothing for this guard to judge.
 toplevel=$(git -C "$cwd" rev-parse --show-toplevel 2>/dev/null) || exit 0
--- a/tests/test_guard_enterworktree_parked_cwd.sh
+++ b/tests/test_guard_enterworktree_parked_cwd.sh
@@ -66,6 +66,21 @@ _assert 0 "allows Skill at repo root"            "$REPO"                 "Skill"
 _assert 0 "allows Skill in tracked subdir"       "$REPO/src"             "Skill"
 _assert 0 "allows Skill outside any repo"        "$NOREPO"               "Skill"
 
+# --- Symlink traversal: cwd is realpath-normalized before the probes (0314) ---
+# A path that reaches the repo through a symlink must resolve to its real
+# location before the check-ignore probe. Otherwise check-ignore errors
+# "outside repository", the guard falls through to allow, and a parked runtime
+# directory addressed via a symlink bypasses the deny (found by the PR #545 panel).
+LINK_PARKED="$TMP/link-parked"
+ln -s "$REPO/projects/session-x" "$LINK_PARKED"
+LINK_ROOT="$TMP/link-root"
+ln -s "$REPO" "$LINK_ROOT"
+
+_assert 2 "denies symlinked parked cwd"              "$LINK_PARKED"
+_assert 2 "denies symlinked parked cwd (Skill)"      "$LINK_PARKED"          "Skill"
+_assert 0 "allows symlinked repo-root cwd"           "$LINK_ROOT"
+_assert 0 "allows symlinked repo-root cwd (Skill)"   "$LINK_ROOT"            "Skill"
+
 # --- Payload without .cwd → fail-safe allow ---------------------------------
 rc=$(printf '{"tool_name":"EnterWorktree","tool_input":{}}' \
         | bash "$HOOK" >/dev/null 2>&1; echo $?)
--- a/tests/test_guard_enterworktree_parked_cwd.sh
+++ b/tests/test_guard_enterworktree_parked_cwd.sh
@@ -78,6 +78,9 @@ ln -s "$REPO" "$LINK_ROOT"
 
 _assert 2 "denies symlinked parked cwd"              "$LINK_PARKED"
 _assert 2 "denies symlinked parked cwd (Skill)"      "$LINK_PARKED"          "Skill"
+# A symlinked ancestor *component* (real leaf under a symlinked root) must also
+# normalize before the probes.
+_assert 2 "denies parked cwd via symlinked ancestor" "$LINK_ROOT/projects/session-x"
 _assert 0 "allows symlinked repo-root cwd"           "$LINK_ROOT"
 _assert 0 "allows symlinked repo-root cwd (Skill)"   "$LINK_ROOT"            "Skill"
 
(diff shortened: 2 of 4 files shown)


MinhHaDuong (commented):
## Multi-perspective review — APPROVE

Two sonnet reviewers (Correctness, Red team) plus panel-lead verification. The 9-line fix closes the symlink-traversal bypass cleanly and is red-first confirmed. Recommend merge.

### Verified
- **verifiable (Correctness + panel):** RED-first confirmed. Reverting only the two-line normalization makes exactly the two new assertions (`denies symlinked parked cwd`, `+ (Skill)`) fail with exit 0 instead of 2, while the symlinked-repo-root-stays-allowed cases still pass. With the fix, all 26 assertions pass and full `make check` is green (782 passed, exit 0).
- **verifiable (Red team):** Direct hook probing confirms the bypass on `main` (leaf symlink to a git-ignored parked dir → exit 0/allow) and its closure on the branch (→ exit 2/deny). Also verified: intermediate symlinked ancestor resolves correctly (deny), symlinked tracked subdir allows, symlinked repo root normalizes so the `[ "$cwd" = "$toplevel" ]` equality still holds (allow), and the harness's own `.claude/worktrees/<name>` case introduces no toplevel-vs-cwd mismatch from normalizing one side.

### Polarity
- Fail-open (`|| exit 0` on the `cd … && pwd -P`, plus the redundant `[ -z "$cwd" ]`) is preserved on every failure path (nonexistent/removed dir, symlink loop, dangling link) — a normalization failure never flips allow→deny. This matches the guard's pre-existing deny-only design.

### Minor (non-blocking)
- **consider (Correctness, low):** Tests exercise a symlink whose target *is* the parked/root dir, not a symlink as an *intermediate* component with a subdir appended (e.g. link→`projects` + `/session-x`). Red team confirmed this path works by direct probe; it is just not asserted in the test file. Optional to add a case.
- **nofollow (Red team):** TOCTOU on dir removal between the `-d` probe and `cd` also fails open — same shape as the existing jq-missing/no-.cwd branches, within the local-trust model. No action.
- **nofollow (Red team):** Relative `.cwd` resolves against the guard's invocation cwd, but the harness always supplies an absolute `.cwd`; unchanged pre/post — latent, not a regression.

Exit criteria in ticket 0314 met and test-verified. No dissent between reviewers.

MinhHaDuong (commented):
## review-pr panel: PR 548 (ticket 0314 — realpath-normalize parked-cwd guard)

**Overall verdict: approve** (both perspectives independently approve, high confidence)

**Scope**: 27-line, 2-file fix. Normalizes `$cwd` via `cwd=$(cd "$cwd" 2>/dev/null && pwd -P) || exit 0` in `scripts/guard-enterworktree-parked-cwd.sh` before the repo-membership/check-ignore probes, closing a symlink-traversal bypass where `git check-ignore` erroring "outside repository" on a symlinked path was treated as allow. Plus 5 new red-first tests.

`make check`: **782 passed, 1 warning (pre-existing, unrelated asyncio-mock warning in test_beat.py), 0 failures.**
`bash tests/test_guard_enterworktree_parked_cwd.sh`: all 27 assertions pass.

### Correctness perspective — approve, high confidence
- **verifiable:** Reproduced the pre-fix bypass directly (checked out commit before 266bf3c): a symlinked parked cwd exits 0 (allow) on the old script; the new script exits 2 (deny) on the identical input. `git check-ignore -q` on the raw symlinked path independently confirmed to emit "outside repository" (exit 128), exactly matching the ticket's diagnosis.
- **verifiable:** Fix placement is correct and matcher-agnostic — normalization sits once, before both the `rev-parse --show-toplevel` and `check-ignore` calls; both EnterWorktree and Skill payload variants behave identically since the script only branches on `.tool_name` at the deny-message step.
- **verifiable:** Symlink loops (`ln -s a b; ln -s b a`) and a permission-denied ancestor (`chmod 000`) both fail open (exit 0) cleanly and fast — no hang.
- **verifiable:** Relative paths and trailing slashes normalize and deny correctly, same as absolute paths.
- **consider:** A permission-denied ancestor is a real, reproducible way to defeat the guard (fail-open on `cd` failure) — but this is the same pre-existing, explicitly documented trade-off ("a normalization failure must never flip an allow into a deny"), not something this PR introduces.
- **nofollow:** The second `[ -z "$cwd" ] && exit 0` guard is effectively unreachable given `set -e` + the `||`-guarded assignment — harmless belt-and-suspenders.
- **nofollow:** New tests genuinely exercise a symlinked *ancestor component* (`LINK_ROOT/projects/session-x`), not just a symlinked leaf — distinct coverage, correctly targeted.

### Red-team perspective — approve, high confidence
- Hand-crafted bypass attempts against the live script; could not reproduce a working bypass of the specific symlink-traversal deny this PR closes.
- **verifiable (safe):** `..`-traversal embedded in the cwd string resolves correctly both directions (crafted-into-ignored-dir denies, crafted-back-out-to-tracked-dir allows) — `cd "$cwd" && pwd -P` resolves `..` before any probe runs.
- **consider:** A nested git repo planted inside the ignored directory (`git init` inside `projects/session-x`) makes that directory its own toplevel, so the deny never fires — reproduced both directly and via symlink, exit 0 both ways. Confirmed pre-existing (non-symlink case behaves identically) and out of scope for 0314 — would need its own ticket if judged worth closing; the guard's threat model is casual/accidental misdirection, not a planted-repo adversary.
- **nofollow:** Bind mounts — untestable without root, and conceptually orthogonal to the symlink-error bypass this PR targets (a bind mount has no symlink component; `pwd -P` neither helps nor hurts).
- **nofollow:** TOCTOU between resolution and check-ignore — the resolved `$cwd` is a plain string, not re-traversed as a symlink, so swapping the original symlink post-resolution has no effect; a generic ancestor-path race is out of the guard's non-adversarial fail-open design.
- **nofollow:** case-insensitive filesystems — not applicable on Linux, no code path assumes case sensitivity here.

### Deduped findings summary
No blockers. Non-blocking, tagged:
- `consider:` permission-denied ancestor fail-open (pre-existing, documented trade-off, not introduced by this PR)
- `consider:` nested-git-repo-inside-ignored-dir escape (pre-existing, out of scope for 0314 — candidate for its own ticket)
- `nofollow:` unreachable second `[ -z "$cwd" ]` guard (defensive, harmless)
- `nofollow:` bind mounts, TOCTOU, case-insensitive-FS (all out of the guard's stated threat model or require root)

Exit criteria (a) symlinked parked cwd denied on both surfaces, (b) symlinked legitimate cwd stays allowed, (c) `make check` passes — all verified.

MinhHaDuong:
## /gaze actions

round: 1
adherence: PASS — 0 blocking
review-pr: 1 comment posted (panel: correctness + red-team, both approve, high confidence)
simplify: 1 fix applied (commit 30a2cfc — dropped unreachable `[ -z "$cwd" ]` guard after `pwd -P` normalization; 27/27 guard assertions pass post-fix), 1 skipped (realpath reuse — builtin subshell is cheaper than an external exec, deliberate)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Symlinked parked cwd denied on both surfaces: ADDRESSED — commit 266bf3c; assertions "denies symlinked parked cwd", "(Skill)", and "denies parked cwd via symlinked ancestor" pass at HEAD (run directly by the gate)
- Symlinked legitimate cwd stays allowed: ADDRESSED — assertions "allows symlinked repo-root cwd" + "(Skill)" pass at HEAD
- make check passes: ADDRESSED — 782 passed, exit 0, run directly in the review worktree post-simplify

Unresolved review comments: none (all panel minors tagged consider:/nofollow:; the two consider: items — permission-denied-ancestor fail-open, nested-repo-inside-ignored-dir escape — are confirmed pre-existing and out of 0314 scope; the nested-repo escape is a candidate for its own ticket)
Unresolved simplify: none (fix applied and verified)
Adherence: PASS

Scope: exactly 2 files, both traceable to the ticket; no scope overflow.

Rationale:
The 0314 bypass (check-ignore erroring "outside repository" on a symlinked cwd, treated as allow) is closed by realpath-normalizing cwd before both probes, red-first confirmed, with fail-open polarity preserved. The gate independently re-ran the guard suite and the full test suite at post-simplify HEAD and re-fetched the PR reviews rather than trusting summaries. The altitude review verified sibling guards (guard-commit-on-main, guard-worktree-identity, guard-cd-primary-repo) are not exposed to this bug class, so the single-script fix is at the right depth.

telemetry: wall=~680s agents=8 tokens=~513k cost~=na (warn: tokens ≈513k > 500k warn threshold — noted, run completed)

PR #564, opened 2026-07-13
feat(0217): OS-network-isolate the reviewer seat-runner

MinhHaDuong:
## Summary

Closes the last open gaps on the ticket-0217 reviewer seat-runner and promotes
it from prototype to production name.

**Ticket:** tickets/0217-sandbox-runner-reviewer-seat-spinoff-os.erg

### What changed
- **Network deny-by-default (criterion 1).** `--network=host` replaced with
  `--network=none` plus a host-side generic TCP↔Unix-domain-socket forwarder
  (`scripts/seat-runner/net-relay.py`). The container reaches the model endpoint
  only through one bind-mounted Unix socket — a filesystem object, not a network
  route — with a container-side loopback bridge (the same relay, reversed)
  presenting it as the TCP `OPENAI_API_BASE` aider expects. pasta/slirp4netns
  were rejected: full outbound NAT, no per-destination allowlist.
- **Per-seat timeout (criterion 5).** `run_seat` wraps every podman invocation
  in `timeout -k 5 $SEAT_TIMEOUT` (default 600s); a hung seat is killed non-zero
  and loud. Fail-open stays one layer up in `reviewers.sh`.
- **`--self-test-only`.** Skips the endpoint check and exits 0 after the
  containment self-test, so containment is provable with podman + the image
  alone (no LLM). The self-test now also proves external network is denied.
- **Rename.** `scripts/seat-runner-prototype.sh` → `scripts/seat-runner.sh`
  (`git mv`, history-preserving); updated `reviewers.sh` + `reviewers/SKILL.md`.
- **`scripts/seat-runner/Containerfile`** documents the out-of-band image build.
- **`tests/test_seat_runner.sh`** — source-inspection guards (flag presence,
  `--network=none` present / `--network=host` absent in live code, timeout
  wrapper) plus a stubbed-podman timeout test and a real containment self-test
  that SKIPs when podman/image are absent.
- **`HOMEDIR=$HOME`** (was hardcoded `/home/<user>` in the prototype; caught by
  the pre-PR adherence self-gate).

### Evidence — criteria 3 and 4 are annotation-only
- Criterion 3 (BASH_ENV-stripped minimal env): already met by the prototype's
  explicit env construction in `run_seat`.
- Criterion 4 (normalizer): already satisfied by closed ticket 0218 (commit
  27f5eb1, harvest template-drop + dedupe in `reviewers.sh`, tested in
  `tests/test_reviewers.sh`). No normalizer code here.

### Criterion 6 — end-to-end smoke: DONE (stub-endpoint case)
Built `localhost/seat-runner:v1` from the Containerfile, stood up a stub
OpenAI-compatible endpoint (canned `/health` + `/v1/chat/completions` returning
`FINDING|`/`SUMMARY|` lines) and drove the **full hardened pipeline**:
`seat-runner.sh --network=none` reviewed this branch vs `origin/main`, reached
the endpoint **only through the relay socket**, passed the containment
self-test, and wrote contract-shaped findings to the out file (exit 0).
Combined with the 2026-06-05 scorecards (real Devstral/Qwen models over a real
~600-line branch diff, local-branch-diff reviewer, no forge round-trip), this
closes criterion 6 on the network-hardened design. All six exit boxes are
ticked; this PR closes 0217.

### Follow-up (not in scope here)
`tickets/0322` filed: `check-agnostic.sh` never scans `scripts/`, which is how
the hardcoded `HOMEDIR` reached review. Extending the gate surfaces a
pre-existing `beat-settings.json` hook-path violation whose fix needs a Claude
Code hook-expansion decision — hence its own ticket.

### Gates
- `make check-fast`: pass. `tests/test_seat_runner.sh`: 15/15 (real containment
  self-test ran — image present on the build host). `tests/test_reviewers.sh`:
  25/25. `/verify-adherence`: PASS, no findings.


MinhHaDuong:
## Multi-perspective panel review — request-changes

Five sonnet reviewers (correctness, red-team, consistency, scope, doc-propagation) ran in parallel. Three axes came back clean (rename complete, scope tight, docs mostly propagated). **Two independent reviewers found the same two HIGH-severity containment defects, both reproduced on the real host, and I re-confirmed both independently.** These invalidate two of the six exit criteria this PR claims to close, so: **request-changes**. The mechanism is well-built and close — the blockers are focused.

`make check-fast`: 746 passed. `tests/test_seat_runner.sh`: 15/15. **The suite passes while missing both blockers** (see the test-gap note below) — green tests here are not evidence the containment holds.

---

### BLOCKER 1 — the per-seat `timeout` does not kill the podman container (exit criterion 5)

*(correctness + red-team, both reproduced; re-confirmed independently)*

`run_seat()` wraps every seat as `timeout -k 5 "$SEAT_TIMEOUT" podman run --rm ...`. Rootless podman detaches the container's real lifecycle into `conmon`, a process independent of the `podman run` client. `timeout` SIGKILLs only the client; `conmon` and the container survive. Reproduced verbatim on this host:

```
$ timeout -k 2 2 podman run --rm --name zz_timeout_probe --network=none busybox sh -c 'sleep 30'
wrapper exit: 137          # the wrapper "thinks" it won
$ podman ps -a --filter name=zz_timeout_probe --format '{{.Names}} {{.Status}}'
zz_timeout_probe Up 5 seconds   # container still running
```

The header comment ("A hung or slow seat is killed and exits non-zero, loudly") and exit-criterion 5 are false in practice: a genuinely hung or adversarial seat keeps running past the cap, and each timeout event leaks an orphaned, un-`--rm`'d container (repeatable resource exhaustion across raid/gaze cycles). **Fix direction:** give `run_seat` a `--name`, and after `timeout` returns, `podman kill` / `podman rm -f` that name in the `cleanup` trap — do not rely on client-process death to stop the sandbox.

### BLOCKER 2 — `~/.local` is mounted read-only but exposes real secrets (exit criteria 1 & 2)

*(red-team, reproduced; re-confirmed independently)*

The design comment states the real `$HOME` is not mounted "except `~/.local` (ro, code only)". On this host that is false — `-v "$HOMEDIR/.local":"$HOMEDIR/.local":ro` grants the seat read access to (names only, contents not shown):

```
~/.local/share/keyrings/login.keyring          # GNOME keyring, auto-unlocked
~/.local/share/keyrings/Default_keyring.keyring
~/.local/share/jupyter/notebook_secret         # live cookie-signing secret
~/.local/share/neo4j-relate/relate.secret.key
```

This is a working exfil channel, not just a read: the seat has a shell (the self-test proves it), and the *one* network path the sandbox deliberately leaves open is the relay to the review model, whose sanctioned use is "send text, get text back." A prompt-injected seat can `cat` any of these into a `FINDING|...|rationale=` line, which `reviewers.sh` harvests into the human-facing review record.

The self-test's `SECRET-BLOCKED` probe only checks `~/.ssh/id_*` and `~/.claude/scripts/bash-env.sh` — both live in the **tmpfs-shadowed** `$HOME`, not in the one directory actually mounted. So the "containment OK … secrets unreachable" banner proves nothing about the path that matters. **Fix direction:** allowlist the specific subpaths the aider CLI needs (`~/.local/bin`, the uv-tool venv) instead of the whole tree, or bake the reviewer into the image (the Containerfile already has a commented-out path) and drop the `.local` mount; then extend the self-test probe to a real `~/.local` secret path so a regression FAILs.

---

### Test-gap note (why the green suite doesn't cover the blockers)

- `verifiable:` tier-2a "timeout: bounded wall-clock" stubs `podman` with a bash `sleep 30` — a real process `timeout` *can* kill — so it passes while the real container-teardown path (blocker 1) is never exercised. A meaningful guard must run a real named container and assert `podman ps` shows it gone.
- `verifiable:` tier-2b real containment self-test SKIPs whenever `localhost/seat-runner:v1` is absent, which the Containerfile mandates is the default CI/dev state ("NOT built at test time"). So criterion 2's machine proof is vacuous everywhere except a host where a human pre-built the image. It ran here only because the image happened to be present.

### Non-blocking findings

- `consider:` (correctness) Endpoint port defaults to 80 for *any* scheme: `[[ "$ENDPOINT_PORT" == "$ENDPOINT_HOST" ]] && ENDPOINT_PORT=80` gives `https://api.openai.com/v1` → port 80 (expected 443). The shipped local `http://127.0.0.1:8012/v1` config is unaffected, but the flag is documented as generic and will silently break the first cloud https endpoint.
- `consider:` (red-team) The self-test never dials `/relay.sock`, so "reaches the endpoint and nothing else" holds only by static reading of `net-relay.py` (`--connect` is fixed argv, not attacker-derived — confirmed). A positive+negative relay probe would guard against a future protocol-aware/redirectable relay.
- `consider:` (correctness) `net-relay.py` `pump()` calls `shutdown(SHUT_RDWR)` on both sockets on either side's EOF; harmless under the current `--no-stream` request/response flow, fragile if reused for streaming. Sockets are `shutdown` but never explicitly `close`d (fine for this short-lived process).
- `verifiable:` (consistency + doc-propagation) `scripts/seat-runner/Containerfile:19` installs `curl`, but the only `curl` call site is the host-side health check (`seat-runner.sh:83`) — nothing in-container uses it. Unused package; the image-contents comments don't mention it either.
- `verifiable:` (doc-propagation) `STATE.md:29` still lists dropping `--network=host` as an open TODO — stale now that this PR closes it. (STATE refresh is usually a separate pass; noting for completeness.)
- `nofollow:` (scope) Filing ticket 0322 in this PR is fine — it closes nothing (no `Ticket:` line references it) and adds no code coupling; the one-ticket-per-PR reading would split it but the sweep-findings-are-decisions convention covers it.

### Dissent preserved

Consistency, scope, and doc-propagation each returned **approve** on their own axis — those verdicts are accurate and not in conflict with the blockers, which live on the correctness/security axis: the rename sweep is genuinely complete (zero `seat-runner-prototype` refs outside `tickets/`), the diff touches only 0217-related files, criteria 3 & 4 are correctly left as evidence-only (normalizer confirmed live in `reviewers.sh` at commit 27f5eb1), and the docs match the code apart from the two nits above. The PR is one focused fix-round away from mergeable.

🤖 Multi-perspective review via /review-pr (5 agents, sonnet)

MinhHaDuong:
## Fix round (commit 29ce72b) — all panel findings addressed

**Blocker 1 (timeout did not kill the container).** Confirmed: `timeout ... podman run` SIGKILLs only the client; conmon kept the container `Up` (rc=137, container alive). Fixed — each seat is `--name`d and `podman rm -f`'d after the run and in the EXIT trap. Verified: `wrapper rc=137, leftover=NONE`.

**Blocker 2 (`~/.local` exposed secrets).** Confirmed the wholesale `-v ~/.local:ro` mount exposed `keyrings/`, `kwalletd/`, `jupyter/`, `neo4j-relate/`, etc. Fixed — only the resolved aider uv-tool venv + its interpreter are bind-mounted (pure code), aider invoked by absolute path. The self-test secret probe now targets the real `~/.local` tree and asserts `LOCAL-SCOPED` (only `uv` under `~/.local/share`); a wholesale mount flips it to `LOCAL-OVERMOUNTED` and fails the self-test loudly. Full hardened e2e re-run through the relay still passes with the narrowed mount.

**Non-blockers:**
- https endpoints now default to port 443 (was always 80).
- Dropped the unused `curl` from the Containerfile (in-container probes use bash `/dev/tcp`; the host does the health check).
- `net-relay.py`: clean `SHUT_WR` half-close + join/close both sockets — no response truncation, no fd leak.
- Relay is now unit-tested: `tests/test_net_relay.py` round-trips both orientations (unix⇄tcp, tcp⇄unix) without podman.
- Source-inspection guards added to `test_seat_runner.sh` for `--name`/reaping/no-wholesale-mount, so both blockers are regression-locked.

**Deferred with rationale:**
- `STATE.md:29` stale TODO — that status line is cleared at the post-merge `/roar` housekeeping pass (status doc, not a code TODO), kept out of this PR to avoid conflicting with parallel raid sessions' STATE edits.
- `0322` in the same PR — intentional: it is the follow-up filing for the agnostic-gate `scripts/` coverage gap that let the hardcoded `HOMEDIR` reach review. Filing a follow-up ticket in the PR that discovered it is the standard flow.

MinhHaDuong (commented):
## Panel review — 5 perspectives (correctness, red-team, consistency, scope, doc-propagation)

Solid, well-documented hardening. The core `--network=none` + bind-mounted UDS-relay design is sound (kernel-enforced netns isolation, not a cooperative grant), scope discipline is clean, and the tests pass here **including the real containment self-test** — podman + the `localhost/seat-runner:v1` image are present on this host, so criterion 2 was machine-verified, not just SKIPped: `tests/test_seat_runner.sh` 21/21, `tests/test_net_relay.py` 2/2, `make check-fast` 746 passed. Two cheap, confirmed defects are worth fixing before merge; the rest are `consider:`/`nofollow:` notes.

**Panel split (preserved for the author to arbitrate):** correctness + consistency returned *request-changes*; red-team, scope, doc-propagation returned *comment*.

### Recommend fixing in this PR (both cheap, both confirmed)

**`verifiable:` [correctness] Dead container-side readiness probe.** `scripts/seat-runner.sh:226` waits for the loopback bridge with `: < /dev/tcp/127.0.0.1/$PORT`, but the whole block runs via `run_seat sh -c "$BRIDGE_AND_REVIEW"` (line 235) and the image's `/bin/sh` is dash, which has no `/dev/tcp` (reproduced: `sh -c ': < /dev/tcp/127.0.0.1/1'` → `cannot open /dev/tcp/... : No such file`, exit 2; bash → connection-refused, exit 1). The probe therefore never succeeds and the loop degrades to an unconditional ~5 s sleep with no real readiness guarantee before aider connects. The self-test's own net probe (line 198) correctly uses `bash -c` — the asymmetry shows this is an oversight. Fix: run the bridge under `bash -c` (bash is in the image, proven by the passing self-test).

**`verifiable:` [consistency] Rename sweep missed an open ticket's body.** `tickets/0207-agnostic-cli-reviewer-seat-one-config-op.erg:47` still reads `scripts/seat-runner-prototype.sh`. This is the editable `## Relevant files` prose (not append-only log), and 0207 is open + `Blocked-by: 0217`, so it will consume the new path. The PR's own stale-name guard excludes all of `tickets/`, so it can't catch this — one-line edit closes it. (workflow.md: rename sweeps cover the full logical unit.)

### Consider (author may dismiss)

**`consider:` [correctness/red-team] TLS does not survive the relay for HTTPS cloud endpoints.** `CONTAINER_BASE` rewrites the endpoint host to `127.0.0.1` (~line 99), but `net-relay.py` is a protocol-blind byte pump — a real HTTPS endpoint's certificate/SNI check against `127.0.0.1` will fail, and the `https://→443` port branch is effectively dead for real use. 0217's criteria don't require cloud endpoints (default + smoke are local plaintext), so this is a limitation to document, not a blocker: scope the header-comment claim to local/plaintext endpoints or note TLS-through-relay is unsupported.

**`consider:` [red-team] "reach the relayed endpoint and NOTHING else" is a *network-destination* guarantee, not a data-scope one.** `/repo` is a full branch clone (not just the diff) and the relay forwards arbitrary bytes, so a prompt-injected seat could smuggle any repo file to the (trusted) endpoint. Consider scoping the mount/prompt to the diff, or softening the header claim.

**`consider:` [red-team] Self-test secret/local-scope probes are fixed allowlists.** The four secret paths and the `.local/share` name-check read as completeness checks; the real guarantee is structural (tmpfs `$HOME`). A future added mount carrying credentials wouldn't be caught. Consider asserting against the actual mount table rather than a static path list.

**`consider:` [consistency+doc, convergent] STATE.md is stale.** `STATE.md:29` still lists the now-completed 0217 hardening (drop `--network=host`, credential scoping, BASH_ENV strip) as a pending next-action. Refresh at the next housekeeping/`/roar` pass.

**`consider:` [consistency] Stale-name sweep over-claims.** `tests/test_seat_runner.sh:76-79` greps `*.sh/*.md/*.py/*.yml` but omits `*.json` while asserting "no ... refs outside tickets/". Harmless today (zero JSON hits) but the pass message over-states coverage.

### Not pursued

**`nofollow:` [scope]** `net-relay.py`'s two orientations (UDS↔TCP both ways) are both actually invoked — not gold-plating. Criteria 3 (BASH_ENV-stripped minimal env) and 4 (normalizer via closed 0218 + this PR's `^FINDING|/^SUMMARY|` pre-filter) are legitimately annotation-only. Criteria 2/6 degrading to SKIP when podman/image are absent matches the accepted harness pattern for hardware-gated seats.

*Tests run on `origin/t217-sandbox-net-isolation`: `test_seat_runner.sh` 21/21 (real self-test executed), `test_net_relay.py` 2/2, `make check-fast` 746 passed.*

MinhHaDuong:
## Fix round 2 (commit d1c9a70) — both confirmed defects fixed

1. **Dead readiness probe (dash lacks `/dev/tcp`).** Reproduced: the image's `/bin/sh` is dash, `: < /dev/tcp/...` → "cannot open… No such file", so the wait was a blind ~5s sleep. Fixed — the container-side bridge now runs under `bash -c` (bash is in the image), so the `/dev/tcp` readiness check gates on the bridge actually binding. E2e re-verified through the relay (exit 0, findings produced).

2. **0207 rename reference.** `tickets/0207:47` is a mutable **body** line ("Relevant files"), not an append-only log entry, and 0207 is open + `Blocked-by: 0217`, so it will consume the path — updated to `scripts/seat-runner.sh`. The test's stale-name sweep now also scans `*.json`.

**Consider-notes addressed:**
- Softened the header: reachability is now stated as network-destination-scoped, and the byte-blind relay is documented as not preserving TLS (fits 0217's local/plain-HTTP endpoint; an HTTPS cloud endpoint would need SNI/cert-preserving forwarding — out of scope).
- `/repo` exfiltration: the branch diff is exactly the data the reviewer model is meant to receive, so this is the intended data flow, not a leak; the "no other destination" claim is network-scoped and now worded that way.
- STATE.md:29 left for the post-merge `/roar` housekeeping pass (status doc, not a code TODO; kept out to avoid conflicting with parallel raid STATE edits).

Gates: `tests/test_seat_runner.sh` 21/21 (real containment self-test ran — image present), `tests/test_net_relay.py` 2/2, `make check-fast` pass, agnostic gate pass.

MinhHaDuong:
## /gaze actions

round: 2
adherence: PASS — 0 blocking (test_seat_runner.sh 21/21 at d1c9a70, adherence suite 17 passed, agnostic gate pass)
review-pr: 1 blocker + 12 tagged minors returned (5 perspectives: correctness/consistency/scope/red-team/doc-propagation; verdicts comment/comment/approve/request-changes/comment)
simplify: 1 fix applied (deferred to fix round — worktree path guard blocks orchestrator edits), 5 skipped with rationale (safety-relevant behavior: unconditional self-test + --no-hardlinks clone kept; blocker-overlap; coverage-reduction; infra scope)
fix agent: 1 commit (3e1da99)

## /verify-gate verdict

verdict: APPROVED

Exit criteria (ticket 0217):
- 1 network deny-by-default: ADDRESSED — `--network=none` + UDS relay (scripts/seat-runner/net-relay.py), source + runtime NET-BLOCKED guards in tests/test_seat_runner.sh
- 2 OS-layer write/secret block, test-proven: ADDRESSED — self-test SANDBOX-ALIVE/WRITE-BLOCKED/SECRET-BLOCKED/LOCAL-SCOPED, ran real (image present), 27/27
- 3 minimal env, BASH_ENV stripped: ADDRESSED — run_seat() explicit env allowlist, scripts/seat-runner.sh:~183-191 (stale ~145-160 log citation corrected by appended note in 3e1da99)
- 4 output normalizer: ADDRESSED — evidence-only via closed 0218 (commit 27f5eb1, reviewers.sh harvest, tested) + this PR's `^FINDING|/^SUMMARY|` pre-filter
- 5 per-seat fail-open + timeout: ADDRESSED — `timeout -k 5 $SEAT_TIMEOUT` + `--name`/`podman rm -f` reaping (29ce72b); stub + reap tests green
- 6 e2e smoke: ADDRESSED — full hardened pipeline against stub endpoint through the relay, disclosed limitation (real-model scorecards predate hardening); panel dispositioned nofollow (ticket transparent)

Round-1 gate items, all verified RESOLVED in 3e1da99:
- BLOCKER venv-root derivation guard: `_reject_home_exposing_mount()` (pyvenv.cfg required; $HOME/$HOME/.local/ancestor-of-$HOME rejected, FATAL before any podman call, active under --self-test-only too); 3 source guards + 1 stubbed behavioral test; test_seat_runner.sh 27/27. Round-1 gate empirically confirmed the pre-fix failure modes failed safe (podman duplicate-mount hard error; LOCAL-OVERMOUNTED self-test trip), so this landed as defense-in-depth, severity disputed but fix agreed by all parties.
- verifiable: stale criterion-3 citation → append-only correcting log line (original untouched)
- verifiable: 0217→0322 back-reference → added under "Relation to existing tickets"
- simplify: test_net_relay.py dedupe (_serve_echo/_wait_until/_round_trip), pure extraction, 2/2

Unresolved review comments: none (prior rounds 29ce72b/d1c9a70 re-verified held: container reaping, narrowed ~/.local, bash readiness probe, 0207 rename ref)
Unresolved simplify: none
Adherence: PASS

Remaining consider:/nofollow: notes (advisory, non-blocking): IPv6 endpoint parsing, unbounded relay threads, --cap-drop=all defense-in-depth, SKILL.md containment-model doc, tier-2b SKIP in imageless CI, STATE.md:29 staleness (post-merge /roar).

Rationale:
Two internal panel rounds preceded this /gaze; this run independently re-verified their fixes held, surfaced one new finding (unguarded venv-root mount derivation — same class as the round-1 ~/.local exposure, reintroduced via a layout assumption), bounced it through one fix round, and round 2 verified the guard is structurally correct and regression-locked. All six exit criteria carry concrete evidence; scope is clean (3e1da99 touches exactly the four named files). Merge decision belongs to the caller.

/gaze: slow run — telemetry warn thresholds crossed (wall > 15min, tokens > 500k); run completed normally.

telemetry: wall~=1500s agents=10 tokens~=912k(subagent-sum, inner panel excluded) cost~=na

PR #562, opened 2026-07-13
feat(0318): harden worktree-path guard to a blocking deny

MinhHaDuong:
## Summary

Hardens the existing worktree-path guard from an advisory (exit 0 + stderr) to a blocking deny (exit 2), so a worktree-isolated agent's first Edit/Write can no longer silently land in the primary checkout. In the 2026-07-13 ticket-pipeline raid, 7 of 11 execute agents hit this trap via absolute `~/.claude/...` paths; every one self-recovered at token cost, and escalating prose warnings did not stop the recurrence.

## Key finding — ticket premise was stale

The ticket asks for a *new* PreToolUse `Edit|Write` guard, but `scripts/pretooluse-worktree-path-guard.sh` (ticket 0171) already exists and is already wired in `settings.json` at matcher `Write|Edit|NotebookEdit`. It already resolves worktree identity, cwd-relative `file_path`, and primary/worktree roots. It was merely **advisory**: the deny branch warned then fell through to `exit 0`. So this PR **hardens the existing script** — no new script, no `settings.json` change, no rename.

## Changes

`scripts/pretooluse-worktree-path-guard.sh`:
- Primary-root deny branch now `exit 2` (repo's PreToolUse deny convention) with the corrective message plus a third line naming the escape hatch.
- **Memory exemption** (narrow): `projects/*/memory/*` under the primary checkout is exempt — it is tracked via the primary `.gitignore` whitelist by design and `skills/memory` tells every session to write it directly by absolute path. Not a blanket `projects/*` exemption; a no-op for non-harness projects.
- **Escape hatch**: `GUARD_ALLOW_PRIMARY_EDIT` (exported before session start) bypasses the deny. The hook is CLI-spawned, not a Bash-tool subshell, so an agent cannot set it mid-turn — no self-service bypass.
- Header comment updated: states blocking-deny behavior and ticket 0318.

`tests/test_pretooluse_worktree_path_guard.sh` (auto-discovered, no wiring): extended staged red-first. Added an `_rc` helper (captures exit code, distinguishing deny=2 from advisory-then-allow=0), an rc==2 assertion on primary write, memory-exemption and escape-hatch rc==0 cases, a deny-message content check, and explicit rc==0 assertions on the allow cases (silence no longer implies allowed). 16/16 pass; the original 9 cases stay green.

## Residual gaps (out of scope, flagged not dropped)

- Cross-worktree writes (a worktree writing into a *sibling* worktree) — unobserved defect class, not guarded.
[...]

PR #574, opened 2026-07-14
fix(0323): normalize guard paths and protect escape-hatch provenance

MinhHaDuong:
## Summary

Closes the two non-blocking residuals /gaze flagged on PR #562 (ticket 0318),
both pre-existing since the worktree-path guard's creation (ticket 0171).

### Residual 1 — realpath/`..`/symlink evasion of the deny prefix match

`scripts/pretooluse-worktree-path-guard.sh` denies via a string prefix match
`case "$file_path" in "$primary_root"/*)`. A path with `..` traversal
(`/home/u/x/../repo/evil.py`) or routed through a symlinked directory into the
primary root has a raw string that does not lexically begin with `$primary_root`,
so it slips past. Fix: normalize `file_path` with `realpath -m` (collapses `..`,
resolves symlinks in existing dir components, tolerates a not-yet-existing leaf
— correct at PreToolUse time, the target file need not exist yet) before the
match, with a raw-path fallback if realpath is unavailable. The roots are left
as-is: `git rev-parse --show-toplevel` already canonicalizes them and they are
opaque literals under the `_GUARD_*_ROOT` test override, so normalizing them
would be a no-op that only widens the diff. The `projects/*/memory/*` exemption
and the `GUARD_ALLOW_PRIMARY_EDIT` escape hatch (both after the insertion point)
are preserved exactly.

### Residual 2 — REAL self-service bypass of `GUARD_ALLOW_PRIMARY_EDIT`

The ticket asked to establish whether any agent-writable file feeds the hook's
environment. **The chain holds and was verified empirically:**

1. `settings.json` top-level `env` sets `BASH_ENV=scripts/bash-env.sh`, applied
   to subprocesses.
2. The guard hook runs as a `bash` subprocess, which sources `bash-env.sh` at
   startup (non-interactive bash honours `BASH_ENV`).
3. `bash-env.sh` sources `$PWD/.env` under `set -a` (auto-export); the hook's
   `PWD` is the agent's own worktree.
4. An agent writing a `.env` containing `GUARD_ALLOW_PRIMARY_EDIT=1` inside its
   OWN worktree (an ordinary, unblocked write) gets that var auto-exported into
   the guard's env **before** the guard checks it — self-authorizing a
   primary-checkout write. This contradicted the guard's "no self-service
   bypass" header comment.

Fix: `bash-env.sh` snapshots `GUARD_ALLOW_PRIMARY_EDIT` before the `.env`
sourcing and restores-or-unsets it after, so only a value present in the launch
environment (systemd/shell, before the script ran) survives; an agent-set
project-`.env` value is dropped. Only this one variable is touched; all other
`.env` secrets load unchanged. The guard header comment now documents the
enforced provenance and cross-references `bash-env.sh` and ticket 0323.

## Tests (TDD, glob-discovered, integration tier)

- **Guard cases 16/17** in `test_pretooluse_worktree_path_guard.sh`:
  `..`-traversal (env-override fixture) and symlink-through (real-git fixture
  mirroring cases 8/9). Both confirmed RED against the pre-fix guard (silent
  exit 0), GREEN after.
- **New `test_bash_env_guard_var_protection.sh`**, two directions:
  (a) an agent-set project `.env` value must NOT survive sourcing;
  (b) a genuine pre-session (inherited-env) value MUST survive.
  Both required — a naive always-unset fix passes (a) but breaks (b), the real
  human escape hatch. Case (a) confirmed RED against the pre-fix `bash-env.sh`
  (leaked `1`). No real secrets touched: `HOME` points at an empty temp dir so
  `~/.claude/.env` is skipped, and the test `.env` holds only the one guard var.

Test count delta: +3 assertions in the guard suite, +1 new shell suite (2
assertions).

## Gates

- `make check`: 818 passed (a transient `test_seat_runner.sh` container-timing
  flake cleared on re-run; my branch touches nothing seat-runner-related).
- `make lint`: 17 passed.

**Ticket:** tickets/0323-worktree-path-guard-residuals-realpath.erg

### DIFF
--- a/scripts/bash-env.sh
+++ b/scripts/bash-env.sh
@@ -8,6 +8,24 @@
 
 set -a  # mark all subsequent assignments for export
 
+# Provenance guard for GUARD_ALLOW_PRIMARY_EDIT (ticket 0323, residual 2). The
+# worktree-path guard (scripts/pretooluse-worktree-path-guard.sh) treats this
+# variable as a human-only, pre-session escape hatch. That guard runs as a bash
+# subprocess with BASH_ENV pointing here, so this script is sourced into its
+# environment. Because the `.env` files below are sourced under `set -a`, an
+# agent could write a `.env` in its own worktree setting
+# GUARD_ALLOW_PRIMARY_EDIT=1 and have it auto-exported into the guard's env,
+# self-authorizing a primary-checkout write. Snapshot the pre-sourcing value and
+# restore-or-unset it afterwards, so only a value present in the launch
+# environment (systemd/shell, before this script ran) survives. Only this one
+# variable is protected; all other .env values load unchanged.
+if [ -n "${GUARD_ALLOW_PRIMARY_EDIT+x}" ]; then
+    _be_guard_had=1
+    _be_guard_val=$GUARD_ALLOW_PRIMARY_EDIT
+else
+    _be_guard_had=0
+fi
+
 [ -f "$HOME/.claude/.env" ] && source "$HOME/.claude/.env"
 
 # Project-level .env, identified by PWD (Claude Code sets PWD to the project dir
@@ -18,5 +36,14 @@ if [ -n "${PWD:-}" ] && [ -f "$PWD/.env" ]; then
     [ "$_be_proj" != "$_be_user" ] && source "$PWD/.env"
     unset _be_proj _be_user
 fi
+
+# Restore GUARD_ALLOW_PRIMARY_EDIT to its pre-sourcing provenance (see snapshot
+# above). Still under `set -a`, so a restored value keeps its export attribute.
+if [ "$_be_guard_had" = 1 ]; then
+    GUARD_ALLOW_PRIMARY_EDIT=$_be_guard_val
+else
+    unset GUARD_ALLOW_PRIMARY_EDIT
+fi
+unset _be_guard_had _be_guard_val
 
 set +a
--- a/scripts/bash-env.sh
+++ b/scripts/bash-env.sh
@@ -5,26 +5,47 @@
 #
 # BASH_ENV is honored by non-interactive bash (i.e. "bash -c ..."), which is
 # exactly what Claude Code uses for Bash tool calls.
+#
+# Provenance protection for guard-trusted variables (ticket 0323, residual 2 +
+# REROLL). The worktree-path guard (scripts/pretooluse-worktree-path-guard.sh)
+# treats a small set of variables as human-only, pre-session inputs:
+#   GUARD_ALLOW_PRIMARY_EDIT  — the escape hatch that disables the deny entirely;
+#   _GUARD_WORKTREE_ROOT  \     the test-only root override — if BOTH are set the
+#   _GUARD_PRIMARY_ROOT   /     guard trusts them instead of the git-derived
+#                              roots, so matching bogus values disarm it for
+#                              every path.
+# That guard runs as a bash subprocess with BASH_ENV pointing here, so this
+# script is sourced into its environment, and the `.env` files below are sourced
+# under `set -a` (auto-export). An agent can write a `.env` in its OWN worktree
+# (an ordinary, unblocked write); any of these variables set there would be
+# auto-exported into the guard's environment and either self-authorize a
+# primary-checkout write or silently disarm the guard. Only a value present in
+# the launch environment (systemd/shell, before this script ran) may survive; a
+# `.env` contribution must be dropped.
+#
+# The protection covers the FULL trusted set (not just the escape hatch), and its
+# bookkeeping is UNFORGEABLE. A prior fix snapshotted with static, committed
+# variable names (_be_guard_had / _be_guard_val); because `.env` is sourced as
+# bash, a static file could pre-set those exact names and forge the snapshot the
+# restore trusts. Here the snapshot variables are suffixed with a per-process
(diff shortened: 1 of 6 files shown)


MinhHaDuong:
## /gaze actions

round: 2
adherence: PASS — 0 blocking
review-pr: 2 confirmed blockers + 4 tagged minors (code panel; size-gate not triggered, 162 lines / 4 files)
simplify: not run — round-1 review returned two verified security blockers; went straight to REROLL rather than run a quality-only pass on code about to be rewritten
fix agent: 1 commit (`ba93f79`, round 2) — closed both static blockers, but introduced no disclosure of the residual it left open

## /gaze stopped: — verify-gate verdict ESCALATE

**verdict: ESCALATE** (round 2 — no third round)
head: `ba93f79`

### Exit criteria

- **Red test for normalized-path / symlink evasion, then fix**: ADDRESSED — `pretooluse-worktree-path-guard.sh` and `guard-cd-primary-repo.sh` both gain `realpath -m` normalization before the match; guard test cases 16/17 and the new cd-guard dotdot case are RED→GREEN (22/22, 20/20).
- **Escape-hatch env provenance established and DOCUMENTED (document OR close the hole)**: MISSING — see below. Neither branch of the either/or is satisfied.
- **`make check` / `make lint` pass**: ADDRESSED — lint 17 passed; check 818 passed on re-run (a `test_seat_runner.sh` container-timing flake, unrelated to this diff).

### What was fixed and verified closed

Round-1 review found two CONFIRMED, independently reproduced bypasses. Both are now closed on `ba93f79`, re-verified by the round-2 gate:

1. **Forgeable static snapshot bookkeeping** — the round-1 fix used committed names `_be_guard_had`/`_be_guard_val`; a `.env` forging them re-authorized `GUARD_ALLOW_PRIMARY_EDIT`. Now UNSET under the forged `.env`. Closed.
2. **Unprotected `_GUARD_WORKTREE_ROOT` / `_GUARD_PRIMARY_ROOT`** — trusted unconditionally; a `.env` setting both to bogus matching values fully disarmed the guard. Now both UNSET after sourcing. Closed.

Also fixed: `guard-cd-primary-repo.sh` got the same `realpath -m` normalization (sibling defect), and both guards now warn on stderr instead of silently degrading when `realpath` is absent.

### Why this escalates rather than approves

The round-2 gate **independently reproduced an adaptive-`.env` bypass** that survives the fix. Because `bash-env.sh` `source`s the agent-writable `.env` in-process under `set -a`, a `.env` can read the live in-scope nonce (`$_be_nonce`) and forge the exact nonce-suffixed snapshot variable at source time:

```
printf -v "_be_snap_${_be_nonce}_GUARD_ALLOW_PRIMARY_EDIT" %s "1"
```

Observed result against `ba93f79`: `GUARD_ALLOW_PRIMARY_EDIT=1` (re-authorized). An agent that can write the 2-line static `.env` can equally write this 3-line adaptive one, so the class is not meaningfully closed — it matches the standing security-architecture lesson that a cooperative, in-process protection is bypassed by the very misbehaving process it targets.

Two things must be true for merge, and neither is:

- **Not closed**: the adaptive bypass is reproducible. In-process bookkeeping (nonce or not) cannot defend against `source`-ing attacker-controlled code in the trusting shell.
- **Not documented**: no disclosure of this residual exists in the script headers, ticket, commit message, or PR. Worse, the new prose asserts unconditionally *"only a value present in the launch environment survives"* — which the reproduction falsifies.

### What needs the author's judgment

The robust closure is an **architectural decision about the credential-loading path** (BASH_ENV → `bash-env.sh` → `.env` under `set -a`), which the MOA owns:

- **Option A** — source `.env` in a subshell and export back only a whitelist that excludes all guard-trusted vars (`GUARD_ALLOW_PRIMARY_EDIT`, `_GUARD_*_ROOT`).
- **Option B** — parse `.env` as strict `KEY=VALUE` (no `source`), dropping/ignoring guard-trusted keys.
- **Option C** — do not use `BASH_ENV` for the guard hook's process at all.
- **Option D (accept + document)** — keep the static-only protection as a real improvement, correct the false unconditional prose to honestly disclose the adaptive residual, and open a follow-up ticket for the architectural fix.

Recommendation: **the static fix is a genuine improvement worth keeping** (both reproduced bypasses closed, well-tested). The blocker is the false security claim and the undocumented open residual. Either land Option A/B (true closure) or Option D (accept as partial + fix the prose + follow-up ticket) — an MOA call, not something a third review round should decide.

Note: the PR body's `**Ticket:**` line points to `tickets/0323-worktree-path-guard-residuals-realpath.erg`, but the file on the branch is `...-realpath-n.erg` (a collision-renumber `-n` suffix). Fix the body line before merge or `erg-pr-merge` will fail to close the ticket.

telemetry: wall~=30min agents=5 tokens~=486k (subagent sum; own na) cost~=na

MinhHaDuong:
Superseded and split. Residual 1 (realpath path normalization) extracted to #587 — merged (with a gaze-caught fix: guard-cd-primary-repo.sh now normalizes both sides of the primary-root identity check). Residual 2 (escape-hatch provenance) redesigned: the snapshot/restore approach here was structurally unsound (gaze ESCALATE — a project .env sourced as code forges the per-process nonce), replaced by a strict KEY=VALUE parse that never executes the project .env and refuses GUARD_* keys, in #588. Ticket 0323 rescoped accordingly; keys-wiring least-privilege work split to ticket 0330.

PR #575, opened 2026-07-14
feat(0207): agnostic CLI reviewer seat — OpenRouter + llama-server endpoints

MinhHaDuong:
Implements criteria 1-2 **machinery** of ticket 0207. The advisory trial (criteria 3-4) stays OPEN — this PR does not close the ticket.

## What landed

**scripts/seat-runner.sh**
- `--credential-env NAME`: reads the endpoint API key from the named env var in the runner process, exports `OPENAI_API_KEY`, and passes the **bare** `-e OPENAI_API_KEY` passthrough so podman reads the value from the invoking env. The secret never enters podman argv (`ps -ef` leak). Empty/unset var fails loud without echoing the value. Local endpoints keep the inline dummy key.
- `--health-path PATH` (default `/health`, `""` skips): probe URL built from the endpoint **origin** (`SCHEME://authority` + path), not `${ENDPOINT%/v1}`+path which double-prefixed `/api` for an `/api/v1` endpoint. The endpoint decomposition moved ahead of the probe (it already ran unconditionally — the move is safe).
- **https branch**: end-to-end TLS to a cloud endpoint through the raw relay. A `--userns=keep-id` seat cannot bind <1024 (EACCES), so the bridge listens on 8443, `CONTAINER_BASE` keeps scheme+real host with the port swapped, and `--add-host` maps the real hostname to 127.0.0.1 (works under `--network=none`). The client presents valid SNI/cert; the relay moves only encrypted bytes. http endpoints stay **byte-identical** to v1. `net-relay.py` is unchanged.

**skills/reviewers/**
- `credential-env` roster field, threaded through all three `IFS='|' read` sites in lockstep; `--credential-env` passed to the seat-runner only when set.
- panel.yml schema comment + a commented OpenRouter example seat.

**tests/**
- `test_seat_runner.sh`: stubbed-podman argv/env capture proves the secret rides the process env and appears in argv only as bare `-e OPENAI_API_KEY` (ps-leak regression guard); https `--add-host` + 8443 base; http byte-identity; exact health-path URL for `/api/v1` (double-/api guard); new-flag source checks. **42/42.**
- `test_reviewers.sh`: credential-env threads through when set, absent otherwise. **27/27.**
- Fixed a latent `pipefail`+SIGPIPE flake in both suites' assert helpers (`printf | grep -q` → here-string) — grep -q closing the pipe early SIGPIPEd printf, a nondeterministic false FAIL on a large haystack. Now stable across 5 consecutive runs.
- `test_net_relay.py` unchanged: 2/2.

## Verification
- `make check` exit 0 (skills-drift, agnostic-tickets, agnostic-skills, full pytest).
- **Local TLS-bridge behavioral proof PASSED**: self-signed `cloud.test` origin, HTTP 200, cert validated end-to-end through the raw relay inside a `--network=none` container with `--add-host` + in-container 8443 bridge.
- Real containment self-test (podman + image) ran GREEN locally (not SKIPed).

## Known non-issues / follow-ups
- The containment self-test receives `CRED_ARGS` but is `--network=none`; a known, network-denied non-issue.
- **Criterion 2 OpenRouter real smoke is blocked on the author provisioning `OPENROUTER_API_KEY`** (not set in this environment). The llama-server config was smoke-tested under 0217; the local TLS-bridge proof stands in for the OpenRouter transport.

### DIFF
--- a/scripts/seat-runner.sh
+++ b/scripts/seat-runner.sh
@@ -20,8 +20,11 @@
 #     pure code) — never all of ~/.local, which also holds keyrings, wallets,
 #     jupyter/neo4j tokens, and browser cookies. The self-test proves ~/.local
 #     is scoped, not wholesale-mounted.
-#   - env is explicitly constructed (no BASH_ENV, no inherited secrets; only
-#     a dummy OPENAI_API_KEY for the local endpoint)
+#   - env is explicitly constructed (no BASH_ENV, no inherited secrets). A
+#     per-seat endpoint credential MAY be injected via --credential-env, read
+#     from the runner's own env and passed as a bare `-e OPENAI_API_KEY` so it
+#     never enters podman argv; a local unauthenticated endpoint uses an inline
+#     dummy OPENAI_API_KEY instead.
 #   - NETWORK IS DENY-BY-DEFAULT: the seat runs under --network=none, so its
 #     netns has no interface but loopback — no route to any host. It reaches
 #     the ONE model endpoint through a bind-mounted Unix-domain socket
--- a/scripts/seat-runner.sh
+++ b/scripts/seat-runner.sh
@@ -34,9 +34,13 @@
 #     DESTINATION — no arbitrary outbound host, no domain-fronting exfil.
 #     (Replaces the v1 --network=host gap: pasta/slirp4netns do full outbound
 #     NAT with no per-destination allowlist, so they were rejected.) The relay
-#     forwards raw bytes to a PLAIN host:port; it does not terminate or preserve
-#     TLS, so it fits a local/plain-HTTP endpoint (0217's target). A real HTTPS
-#     cloud endpoint would need SNI/cert-preserving forwarding — out of scope.
+#     forwards raw bytes to a PLAIN host:port; it neither terminates nor inspects
+#     TLS. A plain-HTTP endpoint (0217's local target) uses the same port on
+#     both sides. An HTTPS cloud endpoint (0207) rides through the same raw pump
+#     end-to-end: --add-host maps the real hostname to loopback and the seat's
+#     client dials an in-container bridge on 8443 (a --userns=keep-id seat cannot
+#     bind 443), so the client still presents valid SNI and validates the real
+#     cert — the relay just moves the encrypted bytes; it never sees plaintext.
 #   - PER-SEAT TIMEOUT: every container invocation is wrapped in `timeout`
 #     (SEAT_TIMEOUT, default 600s). timeout SIGKILLs only the podman CLIENT —
 #     rootless conmon keeps the container alive — so each seat is named and
@@ -50,7 +54,14 @@
 # Usage:
 #   seat-runner.sh --branch BRANCH [--repo PATH] [--base REF]
 #                  [--endpoint URL] [--model NAME] [--out FILE]
+#                  [--credential-env VAR] [--health-path PATH]
 #   seat-runner.sh --self-test-only      # prove containment; no endpoint, no diff
+#
+#   --credential-env VAR : read the endpoint's API key from env var VAR and pass
+#                          it to the seat via env (never argv). Default: a dummy
+#                          key for a local, unauthenticated endpoint.
+#   --health-path PATH   : probe path appended to the endpoint ORIGIN (default
+#                          /health); empty string skips the reachability probe.
 #
 # Env: SEAT_TIMEOUT (per-seat wall-clock cap, seconds; default 600)
 set -euo pipefail
@@ -67,6 +78,8 @@ OUT="/dev/stdout"
 IMAGE="localhost/seat-runner:v1"   # ubuntu:24.04 + python3 + git + ca-certs
 SEAT_TIMEOUT="${SEAT_TIMEOUT:-600}"
 SELF_TEST_ONLY=0
+CREDENTIAL_ENV=""          # env var holding the endpoint's API key (never argv)
+HEALTH_PATH="/health"      # probe path appended to the origin; "" skips the probe
 
 while [[ $# -gt 0 ]]; do
     case "$1" in
@@ -76,6 +89,8 @@ while [[ $# -gt 0 ]]; do
         --endpoint)       ENDPOINT="$2"; shift 2 ;;
         --model)          MODEL="$2"; shift 2 ;;
         --out)            OUT="$2"; shift 2 ;;
+        --credential-env) CREDENTIAL_ENV="$2"; shift 2 ;;
+        --health-path)    HEALTH_PATH="$2"; shift 2 ;;
         --self-test-only) SELF_TEST_ONLY=1; shift ;;
         *) echo "seat-runner: unknown arg $1" >&2; exit 2 ;;
     esac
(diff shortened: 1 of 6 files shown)


MinhHaDuong (commented):
## Multi-perspective review — 5 agents (Correctness, Consistency, Scope, Red Team, Doc Propagation)

**Tests**: `make check` exit 0 locally on the PR branch (817 pytest items, no failures; skills-drift and agnostic-ticket checks clean).

Panel verdicts: Correctness=**approve**, Consistency=**comment**, Scope=**approve**, Doc Propagation=**comment**, **Red Team=request-changes**. Preserving that dissent below — the human author decides whether it gates this PR or a fast-follow.

---

### Most serious finding — Red Team, `verifiable:` (mechanism confirmed; exploitability depends on not-yet-wired future work)

No redaction exists between a real `--credential-env` secret and the seat's public-facing output. The seat is explicitly designed to process adversarial diff content; its only output channel is `grep -E '^(FINDING|SUMMARY)\|' raw.out` (seat-runner.sh:317), which does no content filtering. `reviewers.sh harvest` (lines 137-172) passes the `rationale` field straight through into the exact contract shape `skills/reviewers/SKILL.md:49-54` documents as feeding "the gate's contract shape" — i.e., the shape that ends up disposed as a review comment. A prompt-injected diff that induces the LLM to echo `$OPENAI_API_KEY` back in a `rationale=` field would ride this path into a normalized finding.

Mitigating context I verified: no real credentialed seat is active in `panel.yml` (only commented examples), and `SKILL.md`/`gaze/SKILL.md` show the harvest→gate auto-post wiring for `cli-agent`/`local-model` findings isn't landed yet (only `forge-bot` findings currently post as review comments directly). So this isn't exploitable *today*, but it's machinery this exact PR ships for putting a real key in front of adversarial content, and the PR's own stated threat model is "the secret never enters podman argv" — this is a larger leak channel than the one being closed. Recommend: scrub the known secret literal from `raw.out` before the contract-filter, landed before `OPENROUTER_API_KEY` is ever supplied for the real criterion-2 smoke test.

Related, lower-severity: `consider:` `--credential-env` takes an unconstrained var name (no `*_API_KEY`/`*_TOKEN` allowlist) — a careless roster entry could smuggle an unrelated host secret (e.g. `GITHUB_TOKEN`) into the sandbox. `verifiable:` a credential-misconfiguration FATAL (empty/unset `--credential-env` var) is swallowed into the same generic "WARN seat failed (fail-open)" bucket as any other seat failure (`reviewers.sh:126`) — no distinguishing signal that it's an operator config error vs. a flaky endpoint.

### Correctness — approve, high confidence

`verifiable:` `--health-path` is implemented in `seat-runner.sh` (flag, default `/health`, fail-loud probe) but never threaded through the roster: `roster_records()`/`rec()` in `reviewers.sh:41-58` carry `credential-env` but not `health-path`, and `request)` builds no `--health-path` arg. Works today for the shipped `openrouter-frontier` example only because `openrouter.ai/health` happens to return 200 at the domain root — not by design. No per-seat override exists for an endpoint whose health route lives elsewhere.

`consider:` the "double-`/api`" bug description (comments + PR body) is imprecise — traced both formulas by hand: old `${ENDPOINT%/v1}/health` on `https://openrouter.ai/api/v1` → `https://openrouter.ai/api/health` (one extra segment, never a literal `api/api` doubling). The new origin-based construction is still a real, correct improvement; the regression test's `assert_absent "api/api"` just can't ever fail since that string was never producible by either version.

### Consistency — comment, high confidence

`verifiable:` `scripts/seat-runner.sh:23-24` header still claims "no inherited secrets; only a dummy OPENAI_API_KEY for the local endpoint" — stale now that `--credential-env` injects a real one.

`verifiable:` new test code reintroduces the exact `printf | grep -q` SIGPIPE anti-pattern this same PR fixes elsewhere — `tests/test_seat_runner.sh:157` (`if printf '%s' "$CODE" | grep -Eq ...`), added by this PR, not pre-existing.

`consider:` residual pre-existing `printf | grep -q`/`grep -c` occurrences left unswept: `tests/test_seat_runner.sh:51,72`, `tests/test_reviewers.sh:103,110` — same flake class the PR explicitly calls out, worth a follow-up sweep.

### Scope — approve, high confidence

All six changed files are accounted for in the PR's "What landed" summary; nothing over- or under-claimed. The HTTPS/TLS-bridge machinery is a necessary corollary of "both endpoint configs" (OpenRouter is HTTPS-only) rather than scope creep — 0217's design already rejected `--network=host`/pasta/slirp for lacking per-destination allowlisting, so bypassing the sandbox wasn't an option. The pipefail/SIGPIPE fix is a defensible bundled hitchhiker (confirmed pre-existing, and this PR's own new large-haystack assertions increase exposure to it) though strictly a second change per one-change-per-commit discipline. `panel.yml`'s new example is fully commented — no seat silently activated.

### Doc Propagation — comment (approve-leaning), high confidence

`verifiable:` the health-path gap above also means the shipped `openrouter-frontier` example has no documented escape hatch if a real endpoint's health check isn't at the origin root. Everything else checked out clean: the ticket's "Credential path" section matches the implementation, `panel.yml`'s schema comment and `SKILL.md` are in sync, no SKILL.md frontmatter touched (skills-catalog check moot), and ticket 0217's closing notes have no dangling recommendations this PR missed. `consider:` `STATE.md:29` still lists 0207 as blocked on 0217 and will need pruning post-merge — routine `/lair` upkeep, not a defect in this PR.

---

**Summary**: machinery is sound and well-tested for what's shipped; the credential-argv-leak threat model the PR sets out to close is genuinely closed. The panel's one dissent is the red team's finding that the same change opens a *different*, currently-dormant leak path (secret → adversarial-content-driven LLM output → unredacted findings). Given criterion 2 (real OpenRouter smoke) is already gated on the author supplying a key, that's a natural point to land the output-scrub fix first.

MinhHaDuong:
## /gaze actions

round: 2
adherence: PASS — 0 blocking
review-pr: 1 review posted (panel majority approve; red-team dissent preserved)
simplify: 0 findings (both code reviewers found the code sound)
fix agent: 1 commit (72659a1) — round 1 REROLL, 2 verifiable minors resolved

## /verify-gate verdict

verdict: APPROVED

This is a non-closing (`Ticket-ref:`) PR; it delivers the **criteria 1-2 machinery** of ticket 0207. The advisory trial (criteria 3-4) correctly stays open on the ticket.

Exit criteria (against the PR's stated scope):
- C1 — seat runs an agnostic CLI reviewer via the 0217 sandbox-runner, credential via BASH_ENV minimal-env: **ADDRESSED** — `--credential-env` reads the key from the runner's own env and passes a bare `-e OPENAI_API_KEY` (never argv); the ps-leak regression test captures argv/env separately and proves the secret rides env-only. 42/42 + 27/27 green.
- C2 — both endpoint configs smoke-tested: **ADDRESSED (machinery)** — http path byte-identical to v1; https TLS-bridge (`--add-host` + in-container 8443) validated by a local self-signed end-to-end proof. The **real OpenRouter smoke** is explicitly and correctly left OPEN on the ticket (blocked on the author provisioning `OPENROUTER_API_KEY`).
- C3 (advisory trial ≥5 MRs / ≥3 projects per config) and C4 (promote/drop recommendation): **out of this PR's scope** — stay open, via `Ticket-ref:`.

Round-1 REROLL (bounced on 2 unresolved `verifiable:` minors), resolved by 72659a1:
- `scripts/seat-runner.sh` header: stale containment claim ("no inherited secrets; only a dummy OPENAI_API_KEY") reworded to reflect the credential-env path accurately — the harness's standing "no overstated containment claim" bar.
- `tests/test_seat_runner.sh`: a newly-added inline `printf '%s' "$CODE" | grep -Eq` (the exact SIGPIPE-under-pipefail flake this PR's commit message claims to fix in the assert helpers) converted to a here-string; the two sibling inline sites (l.51, l.72) swept in the same commit for consistency.

Unresolved review comments: none blocking.
Unresolved simplify: none.
Adherence: PASS (make lint: 17 passed, exit 0).

Advisory / follow-up (not gating this machinery PR):
- **Red-team (author decision, before C2 goes live):** a real credential injected into a sandbox that processes adversarial diff content, with no redaction on the findings→harvest→gate channel, is a wider leak surface than the argv leak this PR closes. Not reachable today (no active credentialed seat; the OpenRouter seat is commented out; no real key provisioned). Should be addressed on ticket 0207 before `OPENROUTER_API_KEY` is supplied for the real criterion-2 smoke.
- `consider:` `--health-path` is accepted by seat-runner but not threaded through the panel.yml/reviewers.sh roster schema — inert today because both configured seats use the default `/health`; wire it when a non-`/health` endpoint is added.
- `consider:` inherited (pre-PR) endpoint-parsing assumptions: IPv6-in-brackets and scheme-less endpoints are mishandled by the `%%:*`/`##*:` decomposition — irrelevant to the two configured seat kinds.

Rationale: Every in-scope exit criterion is ADDRESSED with concrete evidence (commit + test), the two round-1 `verifiable:` minors are fixed and re-verified, CI is green, and no scope overflow. The remaining items are forward-looking and correctly tracked on the still-open ticket rather than this machinery PR. The merge decision belongs to the caller.

telemetry: wall~=16m agents=4 tokens~=317k(subagents) cost~=na

PR #583, opened 2026-07-14
ticket(0323,0330): rescope guard escape-hatch to strict-parse; split keys-wiring

MinhHaDuong:
This PR only edits/files tickets — no code change.

Rewrites ticket 0323 and files new ticket 0330. Ticket 0323's residual 2 (escape-hatch env provenance) is rescoped from a doc gap to a real bypass: bash-env.sh sources the project \$PWD/.env as code, so a project-controlled .env can self-grant GUARD_ALLOW_PRIMARY_EDIT. The snapshot/restore approach (PR #574) is abandoned as structurally unsound (gaze ESCALATE — sourced .env forges the per-process nonce); the fix becomes a strict KEY=VALUE parse that never sources and refuses GUARD_* keys. New ticket 0330 splits out the harness-wide ~/.config/keys/ least-privilege secret-loading wiring, Blocked-by 0323. Both tickets stay open.

Ticket-ref: tickets/0323-worktree-path-guard-residuals-realpath-n.erg
Ticket-ref: tickets/0330-wire-config-keys-least-privilege.erg

### DIFF
--- a/tickets/0323-worktree-path-guard-residuals-realpath-n.erg
+++ b/tickets/0323-worktree-path-guard-residuals-realpath-n.erg
@@ -5,6 +5,7 @@ Author: Minh Ha-Duong
 
 --- log ---
 2026-07-13T20:27Z Minh Ha-Duong created
+2026-07-14T09:35Z Minh Ha-Duong note rescoped residual 2 to strict-parse design; snapshot/restore abandoned per gaze ESCALATE; keys wiring split to 0330
 
 --- body ---
 ## Context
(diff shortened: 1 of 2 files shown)


MinhHaDuong:
Superseded. This PR was the ticket-rewrite + keys-ticket-filing step, but the raid overtook it: ticket 0323 was implemented and closed directly via #587 (realpath residual-1) and #588 (strict-parse residual-2), so the 0323 body-rescope here is moot. The keys-wiring ticket filed here at 0330 collided with an existing closed ticket in that slot (optimistic-ID trap); it was refiled at the next free ID 0334 and implemented in #593. Nothing here still needs to land.

PR #587, opened 2026-07-14
fix(0323): normalize guard file_path with realpath -m (residual 1)

MinhHaDuong:
## What

Extracts **residual 1** from PR #574 onto a clean `origin/main` base and opens it standalone.

Both path guards (`scripts/pretooluse-worktree-path-guard.sh`,
`scripts/guard-cd-primary-repo.sh`) compared a **raw, un-normalized** path
against the primary-repo root, so a path that resolves *into* the primary
checkout but does not lexically begin with its root slipped past the
string compare — a silent bypass. Two evasion classes:

- `..`-traversal: `/home/user/x/../repo/evil.py` collapses into the primary
  root but the raw string does not prefix-match it.
- symlink-through: a path routed via a symlink pointing at the primary root
  has a raw string that never starts with the real root path.

The fix normalizes the candidate path with `realpath -m` before the
prefix/equality check: it collapses `..` and resolves symlinks in the
existing directory components while **tolerating a not-yet-existing leaf**
(`--canonicalize-missing`), which is correct at PreToolUse time — the file
being written need not exist yet. When `realpath` is unavailable the guard
falls back to the raw path and **warns** so the degraded (still-vulnerable)
mode is visible rather than silent.

## Tests (red before the fix)

- `tests/test_pretooluse_worktree_path_guard.sh` — case 16 (`..`-traversal
  into primary, exit 2 + corrective message) and case 17 (symlink-through,
  real-git fixture mirroring cases 8/9). Suite: 22 PASS.
- `tests/test_guard_cd_primary_repo.sh` — dot-dot evasion case. Suite: 20 PASS.
- `make lint`: 17 adherence tests pass.

## Residual 2 dropped

Residual 2 from PR #574 (snapshot/restore of `GUARD_ALLOW_PRIMARY_EDIT` in
`scripts/bash-env.sh`, plus its guard-var test) is **not** carried here. It
is structurally unsound per the gaze ESCALATE — a sourced project `.env`
can forge the per-process nonce — and is being redesigned under ticket 0323
(strict `KEY=VALUE` parse), with the keys-wiring split to 0330. The
residual-2 provenance comment in the pretooluse guard was reverted to
`origin/main`'s wording so this PR documents only what it ships.
`scripts/bash-env.sh` is untouched here.

Ticket-ref: tickets/0323-worktree-path-guard-residuals-realpath-n.erg

### DIFF
--- a/scripts/guard-cd-primary-repo.sh
+++ b/scripts/guard-cd-primary-repo.sh
@@ -54,15 +54,21 @@ target="${target%/}"
 target="${target/#\~/$HOME}"
 target="${target//\$HOME/$HOME}"
 
-# Normalize `..` traversal and symlinks before the equality check (ticket 0323,
-# minor A): `cd $primary/sub/.. && git commit` resolves back to $primary but does
-# not lexically equal it, so it would slip past the string compare below. This is
-# the same evasion residual 1 closed in pretooluse-worktree-path-guard.sh.
-# realpath -m collapses `..` and resolves symlinks in existing components while
-# tolerating a not-yet-existing leaf. Warn on the fallback so the degraded
-# (raw-path, still-vulnerable) mode is visible rather than silent.
+# Normalize `..` traversal and symlinks on BOTH sides before the equality check
+# (ticket 0323, residual 1). Two evasions this closes: `cd $primary/sub/.. &&
+# git commit` resolves back to $primary without lexically equalling it (minor A);
+# and a session cwd spelled through a symlink alias of the primary root leaves
+# primary_root aliased while the target resolves canonical, so the two spellings
+# of the same directory would not string-match. Normalizing only one side would
+# miss the second — both target and primary_root must be canonicalized. This
+# closes the symlink/`..` evasion class here for parity with residual 1 in
+# pretooluse-worktree-path-guard.sh (whose roots come from `git rev-parse`, so
+# they are already canonical). realpath -m collapses `..` and resolves symlinks in
+# existing components while tolerating a not-yet-existing leaf. Warn on the
+# fallback so the degraded (raw-path, still-vulnerable) mode is visible.
 if command -v realpath &>/dev/null; then
     target=$(realpath -m -- "$target" 2>/dev/null || echo "$target")
+    primary_root=$(realpath -m -- "$primary_root" 2>/dev/null || echo "$primary_root")
 else
     echo "guard: realpath unavailable, path normalization skipped" >&2
 fi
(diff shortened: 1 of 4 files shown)


MinhHaDuong:
## /gaze actions

round: 2
tier: small
adherence: PASS — 0 blocking (make lint 17/17, make check 835/835; no Python source or `.claude/rules/*.md` touched)
review: 1 (built-in /review — clean; one pre-existing SC2295 out of scope)
review-pr: 1 (correctness + red-team panel)
simplify: 0 fixes applied (81-line bash diff, no simplification surface; quality covered by the review panel)
fix agent: 1 commit (c9e45ed)

## /verify-gate verdict

verdict: APPROVED

Exit criteria (ticket 0323 — this PR is a `Ticket-ref`, delivering residual 1 only; residual 2 / env-provenance is explicitly deferred to the 0323 redesign + 0330 and is not in scope here):

- Red test showing a normalized-path or symlink evasion, then fix: ADDRESSED. Both guards now close the `..`-traversal AND symlink-through evasion classes named in residual 1, each with red-proven tests — `pretooluse` cases 16/17 and `cd-guard` dotdot + symlink-alias case. Each was verified to exit 0 (bypass) against the pre-fix script and exit 2 (deny) after. `realpath -m` is applied after cwd-relative absolutization and before the root comparison, with a `set -euo pipefail`-safe raw-path+warn fallback when realpath is unavailable.
- `make check` and `make lint` pass: ADDRESSED. make lint 17/17, make check 835/835 on the round-2 tip.

Round-1 REROLL finding — resolved:
- `guard-cd-primary-repo.sh` normalized only `target`, not `primary_root` (derived by raw string-truncation of the hook JSON `.cwd`). Independently reproduced by the review panel and by the gate: a worktree session whose cwd is spelled through a symlink alias of the primary root let `cd <canonical-primary> && git commit` slip past the deny (exit 0). The PR's own new comment claimed parity with the pretooluse guard, which was false. Fixed in c9e45ed: `primary_root` is now realpath-normalized with the same fallback, both sides of the identity check are canonical, the comment was corrected, and a symlink-alias red test was added to the cd-guard suite (RED against pre-fix, GREEN after). Reproduce snippet now returns exit 2.

Unresolved simplify: none
Adherence: PASS

Scope: 4 files, all traceable to residual 1 (two guard scripts + their two test suites). No scope overflow.

Rationale: Round 1 bounced on a confirmed, in-scope symlink-alias bypass of the exact evasion class residual 1 names, left open in `guard-cd-primary-repo.sh` while its own comment claimed closure — a verifiable finding, reproduced by the gate, not a nit. The fix agent normalized `primary_root` for symmetry and added the missing red-proven test; the gate re-verified the reproduction now blocks, both shell suites are green (cd-guard 21/0, pretooluse 22/0), the new test is a genuine ratchet, and the full suite passes. Residual 1 is now closed in both guards. Residual 2 remains open under the ticket by design.

telemetry: tier=small wall=~1500s agents=3 tokens=na cost~=na

PR #588, opened 2026-07-14
fix(0323): strict-parse project .env, refuse GUARD_* keys (residual 2)

MinhHaDuong:
## What

The BASH_ENV shim `scripts/bash-env.sh` sourced the project-level `$PWD/.env`
**as shell code**. A project-controlled `.env` — any agent write, any cloned
repo — could therefore forge the per-process guard nonce
`GUARD_ALLOW_PRIMARY_EDIT` (bypassing the worktree-path deny guard) or run
arbitrary shell on **every** bash subprocess. This is the escape-hatch
provenance hole flagged at review of ticket 0323 (residual 2).

The earlier residual-2 attempt (snapshot/restore of `GUARD_ALLOW_PRIMARY_EDIT`
in the shim) was structurally unsound per the gaze **ESCALATE**: a sourced
project `.env` forges the nonce before any restore runs. That approach never
reached main. This PR is the structural replacement.

## Design — strict parser for the UNTRUSTED project file only

`$PWD/.env` is now **parsed, never executed**:

- iterate lines with `while IFS= read -r`; skip blank lines and `#` comments;
- strip an optional leading `export `;
- require `^[A-Za-z_][A-Za-z0-9_]*=`; the value is everything after the FIRST
  `=`, assigned **literally** — a `$(cmd)`/backtick value is stored as text,
  never run;
- strip ONE matching pair of surrounding quotes (literal, no escape processing);
- **REFUSE** any `GUARD_*` key with a one-line stderr warning
  (`bash-env: refusing GUARD_* key from project .env: <key>`) — an untrusted
  file structurally cannot set guard vars;
- malformed lines are skipped, never fatal, so the shim stays fast on every
  subprocess.

The **trusted, user-owned** `$HOME/.claude/.env` KEEPS its `source` behavior
(full expansion) — only the untrusted project file switches to strict parse.
The same-file dedup guard (project realpath == user realpath) is preserved.

## Tests — `tests/test_bash_env_project_env_parse.sh` (red before the fix)

New shell suite (auto-discovered by `tests/test_bash_suites.py`). Runs
bash-env.sh in an isolated subprocess with a controlled `HOME` and project dir:

1. project `.env` with `GUARD_ALLOW_PRIMARY_EDIT=1` → var is **not** set.
2. project `.env` with `EVIL=$(touch …)` / backticks → side-effect file is
   **not** created, no shell error, and the literal text is stored.
3. `FOO=bar` → loads as exactly `bar`.
4. quoted `BAZ="a b"` → loads as `a b` (one quote pair stripped, space kept).
5. regression: a user-owned `$HOME/.claude/.env` (`USERKEY=uval`) still loads
   via full `source` — proving the trusted file was not strict-parsed.

**Red-proof** against the pre-fix (source-based) script:
- case 1 → FAIL: `GUARD_ALLOW_PRIMARY_EDIT` was set to `1`.
- case 2 → FAIL: the `pwned` file **was created** (arbitrary code executed) and
  the literal value was not stored.

After the fix: all 7 assertions PASS.

## Old test note

No `test_bash_env_guard_var_protection.sh` was deleted — that test and the
snapshot/restore mechanism lived only in the abandoned PR #574 and never
reached main. Its concern is subsumed here: a project `.env` cannot set
`GUARD_*` at all.

## Gates

- new suite: 7/7 PASS
- `tests/test_bash_suites.py`: 23 shell suites PASS
- `make lint`: 17 adherence tests PASS
- `make check`: 836 passed

## Sibling

Residual 1 (realpath normalization of the guard `file_path`) is PR #587. With
#587 already in main, merging this PR completes ticket 0323.

**Ticket:** tickets/0323-worktree-path-guard-residuals-realpath-n.erg

### DIFF
--- a/scripts/bash-env.sh
+++ b/scripts/bash-env.sh
@@ -5,18 +5,71 @@
 #
 # BASH_ENV is honored by non-interactive bash (i.e. "bash -c ..."), which is
 # exactly what Claude Code uses for Bash tool calls.
+#
+# Two .env sources with DIFFERENT trust levels:
+#
+#   * $HOME/.claude/.env — user-owned, TRUSTED. Sourced as shell code (full
+#     expansion), exactly as before. Only the user writes this file.
+#
+#   * $PWD/.env — project-level, UNTRUSTED. An agent, a cloned repo, or any
+#     project write can place it. It is NEVER sourced: it is strict-parsed
+#     KEY=VALUE, values are assigned LITERALLY (no eval / $() / backticks), and
+#     any GUARD_*-namespaced key is REFUSED. This structurally closes the
+#     escape-hatch provenance hole (ticket 0323, residual 2): a sourced project
+#     .env could otherwise forge a per-process guard nonce
+#     (e.g. GUARD_ALLOW_PRIMARY_EDIT) or execute arbitrary shell via BASH_ENV.
+#
+# The script runs on every bash subprocess, so it stays fast and never aborts on
+# a missing or malformed .env — bad lines in the project file are skipped, never
+# fatal.
 
+# --- trusted user-level .env: source as before (full expansion) ---
 set -a  # mark all subsequent assignments for export
-
 [ -f "$HOME/.claude/.env" ] && source "$HOME/.claude/.env"
+set +a
 
-# Project-level .env, identified by PWD (Claude Code sets PWD to the project dir
-# for each subprocess). Skip if it resolves to the same file as the user-level one.
+# --- untrusted project-level .env: strict KEY=VALUE parse, never executed ---
+# Claude Code sets PWD to the project dir for each subprocess. Skip if it
+# resolves to the same file as the trusted user-level one (already loaded).
 if [ -n "${PWD:-}" ] && [ -f "$PWD/.env" ]; then
     _be_proj="$(realpath "$PWD/.env" 2>/dev/null)"
     _be_user="$(realpath "$HOME/.claude/.env" 2>/dev/null)"
-    [ "$_be_proj" != "$_be_user" ] && source "$PWD/.env"
+    if [ "$_be_proj" != "$_be_user" ]; then
+        while IFS= read -r _be_line || [ -n "$_be_line" ]; do
+            # strip leading whitespace for blank/comment detection
+            _be_trim="${_be_line#"${_be_line%%[![:space:]]*}"}"
+            [ -z "$_be_trim" ] && continue          # blank line
+            [ "${_be_trim:0:1}" = "#" ] && continue # comment
+            _be_trim="${_be_trim#export }"          # optional `export ` prefix
+            # require KEY=VALUE with a valid identifier key
+            case "$_be_trim" in
+                [A-Za-z_]*=*) ;;
+                *) continue ;;
+            esac
+            _be_key="${_be_trim%%=*}"
+            case "$_be_key" in
+                *[!A-Za-z0-9_]*) continue ;;        # key has an invalid char
+            esac
+            _be_val="${_be_trim#*=}"                # everything after the FIRST =
+            # strip ONE matching pair of surrounding quotes (literal, no escapes)
+            if [ "${#_be_val}" -ge 2 ]; then
+                _be_first="${_be_val:0:1}"
+                _be_last="${_be_val: -1}"
+                if { [ "$_be_first" = '"' ] && [ "$_be_last" = '"' ]; } ||
+                   { [ "$_be_first" = "'" ] && [ "$_be_last" = "'" ]; }; then
+                    _be_val="${_be_val:1:${#_be_val}-2}"
+                fi
+            fi
+            # structural refusal: an untrusted file cannot set guard vars
+            if [[ "$_be_key" == GUARD_* ]]; then
+                printf 'bash-env: refusing GUARD_* key from project .env: %s\n' \
+                    "$_be_key" >&2
+                continue
+            fi
+            # assign the value LITERALLY — never expanded, never executed
+            export "$_be_key=$_be_val"
+        done < "$PWD/.env"
+        unset _be_line _be_trim _be_key _be_val _be_first _be_last
+    fi
     unset _be_proj _be_user
 fi
-
-set +a
--- a/scripts/bash-env.sh
+++ b/scripts/bash-env.sh
@@ -14,10 +14,13 @@
 #   * $PWD/.env — project-level, UNTRUSTED. An agent, a cloned repo, or any
 #     project write can place it. It is NEVER sourced: it is strict-parsed
 #     KEY=VALUE, values are assigned LITERALLY (no eval / $() / backticks), and
-#     any GUARD_*-namespaced key is REFUSED. This structurally closes the
-#     escape-hatch provenance hole (ticket 0323, residual 2): a sourced project
-#     .env could otherwise forge a per-process guard nonce
-#     (e.g. GUARD_ALLOW_PRIMARY_EDIT) or execute arbitrary shell via BASH_ENV.
+#     any guard-namespaced key — any key whose name contains GUARD_, covering
+#     both GUARD_* and the leading-underscore _GUARD_* override pair — is
+#     REFUSED. This structurally closes the escape-hatch provenance hole
+#     (ticket 0323, residual 2): a sourced project .env could otherwise forge a
+#     per-process guard nonce (e.g. GUARD_ALLOW_PRIMARY_EDIT) or the worktree-
+#     path override (_GUARD_WORKTREE_ROOT / _GUARD_PRIMARY_ROOT), or execute
+#     arbitrary shell via BASH_ENV.
 #
 # The script runs on every bash subprocess, so it stays fast and never aborts on
 # a missing or malformed .env — bad lines in the project file are skipped, never
@@ -60,9 +63,14 @@ if [ -n "${PWD:-}" ] && [ -f "$PWD/.env" ]; then
                     _be_val="${_be_val:1:${#_be_val}-2}"
                 fi
             fi
-            # structural refusal: an untrusted file cannot set guard vars
-            if [[ "$_be_key" == GUARD_* ]]; then
-                printf 'bash-env: refusing GUARD_* key from project .env: %s\n' \
+            # structural refusal: an untrusted file cannot set guard vars.
+            # Match every guard-namespaced key — both the GUARD_* form and the
+            # leading-underscore _GUARD_* override pair (_GUARD_WORKTREE_ROOT /
+            # _GUARD_PRIMARY_ROOT) that pretooluse-worktree-path-guard.sh honors
+            # as an unconditional worktree-path override. Substring match, since
+            # no legitimate project key contains the harness-internal GUARD_ token.
+            if [[ "$_be_key" == *GUARD_* ]]; then
+                printf 'bash-env: refusing guard-namespaced key from project .env: %s\n' \
                     "$_be_key" >&2
                 continue
             fi
--- a/tests/test_bash_env_project_env_parse.sh
+++ b/tests/test_bash_env_project_env_parse.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+# Tests for scripts/bash-env.sh — the BASH_ENV shim sourced at the start of every
+# Claude Code bash subprocess. It loads two .env sources with DIFFERENT trust:
+#
+#   * $HOME/.claude/.env  — user-owned, trusted: sourced as shell code.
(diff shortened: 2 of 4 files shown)


MinhHaDuong:
## Multi-perspective review — PR #588

Five parallel Sonnet reviewers (correctness, consistency, scope, red-team, doc-propagation). `make check`: **836 passed** (48.8s, exit 0).

### Blocker — GUARD_* refusal glob misses `_GUARD_WORKTREE_ROOT` / `_GUARD_PRIMARY_ROOT`

**Independently found and live-reproduced by both the correctness and red-team agents.** `scripts/pretooluse-worktree-path-guard.sh:49-52` has its own env-var override for testing:

```bash
if [ -n "${_GUARD_WORKTREE_ROOT:-}" ] && [ -n "${_GUARD_PRIMARY_ROOT:-}" ]; then
    worktree_root="$_GUARD_WORKTREE_ROOT"; primary_root="$_GUARD_PRIMARY_ROOT"
```

The new `bash-env.sh` filter is `[[ "$_be_key" == GUARD_* ]]` — anchored at the string start. Both override vars begin with `_` (underscore), not `G`, so they pass through the "untrusted" parser untouched. A planted `.env` in a project directory containing

```
_GUARD_WORKTREE_ROOT=/tmp/x
_GUARD_PRIMARY_ROOT=/tmp/x
```

sets `worktree_root == primary_root`, which hits the guard's unconditional `exit 0` at line 61 — bypassing the worktree-path deny guard entirely, for every Write/Edit/NotebookEdit, not just the documented `GUARD_ALLOW_PRIMARY_EDIT` escape hatch. Both agents built real reproductions:

- Correctness agent: loaded the two vars through the new parser directly, confirmed both keys survive un-refused while `GUARD_ALLOW_PRIMARY_EDIT` in the same file is correctly refused.
- Red-team agent: ran the guard hook itself as a subprocess with `BASH_ENV` set exactly as `settings.json` declares it globally (not scoped to the Bash tool) and a `.env` planted in a real `git worktree add` worktree — control run: `BLOCKED … EXIT_CODE=2`; attack run with the two `_GUARD_*` keys: `EXIT_CODE=0`.

Caveat both agents flagged honestly: whether Claude Code's runtime actually injects `BASH_ENV` into PreToolUse hook subprocesses (vs. only Bash-tool-call subprocesses) couldn't be confirmed from inside the repo — that's closed-source runtime behavior. But `settings.json`'s `env` block is flat/global and the guard is a plain `#!/bin/bash` hook command, so the assumption is reasonable, and the fix is a one-line glob widening (e.g. also refuse `_GUARD_*`, or generalize to `*GUARD*`/an explicit denylist) sitting in the same file as the rest of this PR's fix. Given the PR's own stated goal is closing *all* project-.env-forgeable guard state, this second guard-trust namespace should be covered before merge, or explicitly and narrowly deferred with a red test proving current non-exploitability under this runtime.

### Findings (tagged, non-blocking)

- `verifiable:` CRLF line endings aren't stripped — `FOO=bar\r\n` loads as `bar\r` (4 chars, confirmed via `od -c`); for quoted values the trailing `\r` sits after the closing quote so quote-stripping silently fails and the raw `"a b"\r` is stored. Not a security bypass, but untested behavior gap vs. `source`. (correctness)
- `consider:` A 2,000,000-line (16MB) planted `.env` took 113s to parse in the `while read` loop — no line/byte cap exists, and this shim runs on every bash subprocess, so an oversized/misnamed file degrades every subsequent Bash call for the session. (red-team)
- `consider:` `bash-env.sh`'s header documents the trust-level split thoroughly but never states why it deliberately omits `set -euo pipefail` (it's sourced into the caller's live shell — setting it would mutate caller shell options). This is a pre-existing, CI-enforced exemption, unchanged by this PR, but a one-line note would close the gap for a future reader without CI/team-memory context. (consistency)
- `consider:` `_be_*` locals are unset at the end of the script (lines 72/74), but an early failure inside the `while read` loop would skip the `unset` and leak `_be_*` into the caller's live shell (this script sources into the caller, not a subshell). Low likelihood, no test covers it. (consistency)
- `consider:` `scripts/pretooluse-worktree-path-guard.sh:105-109`'s comment justifies `GUARD_ALLOW_PRIMARY_EDIT` safety purely by "hook is spawned by the CLI, not a Bash-tool subshell" — unchanged by this PR and now incomplete as the sole trust-boundary explanation, since `bash-env.sh`'s header is the actual provenance guarantee in force. A one-line cross-reference would future-proof it. (doc-propagation)
- `consider:` Global memory `project_bash_env_secret_loading.md` (auto-loaded every session) still describes `bash-env.sh` as sourcing both `~/.claude/.env` and `$PWD/.env` identically — now stale and would mislead a future agent about the trust model. Outside this PR's diff surface (predates it), but worth a follow-up memory update once this merges. (doc-propagation)
- `nofollow:` Case-sensitivity games (`Guard_Allow_Primary_Edit`, `guard_allow_primary_edit`) pass the filter but no downstream consumer does case-insensitive lookup — confirmed harmless. (red-team)
- `nofollow:` Multi-line quoted values will break across lines (real fidelity gap vs. `source`, but outside the stated code-exec threat model). (correctness)
- `nofollow:` PR #574 (abandoned snapshot/restore approach) is confirmed CLOSED/never-merged with zero trace in the current tree; the PR body's claim about it is accurate. (doc-propagation, scope)
- `nofollow:` Ticket 0323's exit-criteria checkboxes stay unticked — confirmed this is not this repo's close signal (the `**Ticket:**` line + `erg close` at merge is); not a gap. (scope, doc-propagation)

### Confirmed holding defenses (red-team, all live-tested)

- No code execution from the project `.env` via command substitution/backticks/export-breakout — literal storage confirmed in every attempted case.
- Leading-whitespace and `export `-prefixed `GUARD_*` keys are still correctly refused after trim/prefix-strip.
- Symlinking the project `.env` to the trusted `$HOME/.claude/.env` doesn't bypass anything — realpath dedup still routes it through the trusted `source` path (requires `$HOME` write access already, out of the untrusted-file threat model).

### Scope / consistency / correctness (no other issues)

Scope agent: file list matches exactly (`scripts/bash-env.sh`, `tests/test_bash_env_project_env_parse.sh`), no drive-by changes, trusted-path behavior genuinely untouched, no over-engineering — every parser feature earns its keep. Consistency agent: naming (`_be_*` prefix) extends pre-existing convention, test harness idiom matches siblings, executable bit present (though not load-bearing for discovery). Correctness agent otherwise confirmed the parser's whitespace/blank/comment detection, key-validation double-check (not redundant — the second check catches what the first doesn't, e.g. `FOO-BAR=baz`), and quote-stripping boundary cases (length 0/1/2) all behave correctly, and the PR's own 7/7 test claim reproduces exactly.

**Verdict: request-changes** — on the `_GUARD_*` gap. Everything else here is solid work closing the primary threat (code execution via project `.env`) cleanly.

MinhHaDuong:
## /gaze actions

round: 1
tier: full
adherence: FAIL — 1 blocking
review: 1 finding (high, live-reproduced)
review-pr: 1 blocker + 4 tagged minors (request-changes; posted)
simplify: skipped (adherence blocking)

## /verify-gate verdict

verdict: REROLL

Exit criteria (ticket 0323):
- Red test + fix for normalized-path/symlink evasion (residual 1): ADDRESSED — PR #587 merged into main (realpath -m normalization at scripts/pretooluse-worktree-path-guard.sh:83-86); this branch is current with it (merge 19a3a51).
- Escape-hatch env provenance established and documented in the script header (residual 2): MISSING — the header now documents the trust split and the parser refuses `GUARD_*`, but a sibling instance of the same guard-provenance class is left open. `scripts/pretooluse-worktree-path-guard.sh:50-52,61` reads `_GUARD_WORKTREE_ROOT`/`_GUARD_PRIMARY_ROOT` (leading underscore) as an unconditional override; the new filter at `scripts/bash-env.sh:64` (`[[ "$_be_key" == GUARD_* ]]`) does not match them. A project `.env` setting both equal forges `worktree_root == primary_root` and hits the guard's unconditional `exit 0`, bypassing the whole worktree-path deny via the same BASH_ENV → project-`.env` vector this PR exists to close. Live-reproduced by two independent reviewers (control exit 2, attack exit 0). The header's claim that GUARD_*-namespaced keys are refused is not accurate for this reachable pair.
- `make check` and `make lint` pass: ADDRESSED — 836 passed, 17 adherence passed, new suite 7/7, 23 shell suites pass.

Unresolved review comments:
- BLOCKER: `_GUARD_WORKTREE_ROOT`/`_GUARD_PRIMARY_ROOT` bypass the GUARD_* refusal (scripts/bash-env.sh:64) — widen the refusal to cover the leading-underscore guard-override pair (e.g. match `GUARD_*|_GUARD_*` or `*GUARD_*`), and add a red test asserting a project `.env` setting `_GUARD_WORKTREE_ROOT`/`_GUARD_PRIMARY_ROOT` is refused, mirroring the existing `GUARD_ALLOW_PRIMARY_EDIT` case.

Unresolved simplify: none (phase skipped — adherence blocking)

Adherence: FAIL (1 blocking)

Rationale:
The primary threat model — arbitrary shell execution via a sourced project `.env` — is closed correctly and verified (no RCE via `$()`, backticks, `export` breakout; no case or symlink-dedup bypass; literal storage confirmed). But the guard-nonce provenance hole this ticket exists to close is only partially shut: the refusal filter enumerates `GUARD_*` and misses the `_GUARD_*` override pair that grants a strictly stronger bypass of the same deny guard, reachable through the identical vector. Three independent reviewers converged on this and two reproduced it live. The fix is a one-line glob widening plus one red test — small, well-scoped, and it removes all doubt about the documented invariant. Round 1 REROLL.

round: 1

MinhHaDuong:
## /gaze actions

round: 2
tier: full
adherence: PASS — 0 blocking
review: 1 finding (resolved in round 2)
review-pr: 1 blocker (resolved) + 4 tagged minors
simplify: skipped (adherence blocking in round 1)
fix agent: 1 commit (1401193)

## /verify-gate verdict

verdict: APPROVED

Exit criteria (ticket 0323):
- Red test + fix for normalized-path/symlink evasion (residual 1): ADDRESSED — PR #587 merged into main (realpath -m normalization, scripts/pretooluse-worktree-path-guard.sh:83-86); this branch is current with it (merge 19a3a51).
- Escape-hatch env provenance established and documented in the script header (residual 2): ADDRESSED — the project `.env` is now strict-parsed (never sourced) and the refusal filter (scripts/bash-env.sh:72, `[[ "$_be_key" == *GUARD_* ]]`) rejects BOTH the `GUARD_*` nonce and the leading-underscore `_GUARD_WORKTREE_ROOT`/`_GUARD_PRIMARY_ROOT` override pair honored by the worktree-path guard. Header comment (lines 15-22) documents both forms truthfully. Red test `tests/test_bash_env_project_env_parse.sh` case 1b proves the `_GUARD_*` pair is refused (RED before the round-2 fix, GREEN after).
- `make check` and `make lint` pass: ADDRESSED — 836 passed, 17 adherence passed, 23 shell suites pass, new suite 9/9. Re-run on the fix tip (1401193).

Unresolved review comments: none blocking. Two `verifiable:` robustness nits remain OPEN as recommended follow-up (out of scope for 0323's provenance invariant, and no regression vs. the prior `source` behavior):
- CRLF-terminated project `.env` values retain a trailing `\r` (one-line fix: `_be_line="${_be_line%$'\r'}"`).
- No line/size cap on the per-subprocess parser (a pathological multi-MB `.env` is slow; the old `source` path had the same or worse cost).
Suggest a single follow-up ticket for both.

Unresolved simplify: none (phase skipped in round 1 — adherence was blocking; not re-run because the round-2 change is a one-line glob widen + test + header, fully covered by the adherence and shell suites).

Adherence: PASS

Scope: clean — 3 files (scripts/bash-env.sh, tests/test_bash_env_project_env_parse.sh, tickets/0323-...erg), all traceable to residual 2. No stray files.

Rationale:
The primary threat model (arbitrary shell execution via a sourced project `.env`) was closed correctly and verified in round 1 — no RCE via `$()`/backticks/`export` breakout, literal storage confirmed, trusted user `.env` still fully sourced. Round 1 REROLLed on a unanimous, live-reproduced blocker: the `GUARD_*` refusal enumerated too narrowly and missed the `_GUARD_WORKTREE_ROOT`/`_GUARD_PRIMARY_ROOT` override pair, which grants a strictly stronger bypass of the same deny guard through the identical BASH_ENV → project-`.env` vector. The round-2 fix (commit 1401193) widens the refusal to a `*GUARD_*` substring match, updates the header to match, and locks it with a red test; the full suite is green on the fix tip. The two remaining `verifiable:` minors are genuine but out of scope for ticket 0323's security invariant and introduce no regression, so they are recommended as a follow-up ticket rather than a merge blocker. Every exit criterion is ADDRESSED; approving.

telemetry: tier=full wall=~1800s agents=4 tokens=na cost~=na

PR #123, opened 2026-05-09
feat(0106): per-project raid_timeout_s in ProjectConfig

MinhHaDuong:
## Summary

- Add `raid_timeout_s: int = RAID_TIMEOUT_S` field to `ProjectConfig`, registered in `_PROJ_KEYS` so it's settable via `projects.json` and per-project `beat.json` overlay
- Wire `project.raid_timeout_s` into the raid `run_skill()` call (replaces hardcoded `RAID_TIMEOUT_S`)
- Add supervisor repair rule: raise `raid_timeout_s` by 20% (capped at 3600s) when `outcome=timeout` root cause is verify/review slowness
- Discriminating test: captures `timeout_s` kwarg and asserts it matches custom config value (2700), not the module default (1800)

Closes #0106

## Test plan

- [x] New test `test_raid_uses_per_project_timeout` passes
- [ ] Existing test suite passes (pre-existing `TestRunSkillDryRun.test_pick_ticket_returns_pick_sentinel` failure is unrelated — `_SkillResult` mock mismatch)
- [ ] Manual: verify `projects.json` accepts `raid_timeout_s` field
- [ ] Manual: verify `beat.json` overlay applies `raid_timeout_s`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #589, opened 2026-07-14
feat(roar): log one celebration per merged PR since last sha

MinhHaDuong:
## Summary
A batched interactive session merges several PRs one at a time then roars once at the end. Roar's telemetry step measured "everything since `roar-last-sha`" and wrote a single aggregate `celebrations.jsonl` record, losing the per-ticket attribution that bump tallies and celebration stats feed on (2026-07-13: eleven PRs #538-#548 collapsed into one 51-commit record with ticket null). Raid roars fire per merged PR and never had this problem.

## Changes
- **New `skills/roar/enumerate-merges.py`** — `enumerate-merges.py <since-sha> [--project NAME]`. Enumerates first-parent merge commits in `<sha>..HEAD` and emits one celebration JSON record per merge (`{project, branch, commits, files_changed, ticket}`; no `ts`/`date` — `log-celebration` stamps those). Branch parsed from the GitHub-shaped merge subject; **ticket recovered by scanning the merge's second-parent range for the erg-pr-merge close commit, not from the branch name** (real branch names carry misleading digits). Repo root resolved at call time (worktree-safe). Empty range prints nothing, exits 0.
- **`skills/roar/SKILL.md` step 2** — pipe each enumerated record through `log-celebration` when the sentinel exists, is an ancestor of HEAD, and the enumeration is non-empty; else fall back to the current single aggregate entry (first roar, rewritten history, squash-merge, no-forge repos). Sentinel-write lines unchanged.
- **`tests/test_roar_enumerate_merges.py`** — integration-tier, real-git fixtures. Anti-tautology core case: a no-ticket-digit branch (`fix-flaky-widget`) carrying a close commit `ticket(0042): ...` still recovers ticket 42; plus empty range, multi-ticket close, non-PR merge subject, and count fields.

`log-celebration` and `erg-pr-merge` are untouched. Nothing reads `celebrations.jsonl` programmatically; lair reads only the sentinel SHA.

## Exit criteria
- [x] A batched interactive roar produces per-PR celebration entries telemetry-equivalent to a raid's per-PR roars
- [x] Aggregate fallback preserved where merge commits are absent
- [x] `make check` passes (840 passed)

Pre-PR `/verify-adherence`: PASS (no mechanical or semantic findings).

**Ticket:** tickets/0331-roar-per-pr-celebrations-since-last-sha.erg

### DIFF
--- a/skills/roar/SKILL.md
+++ b/skills/roar/SKILL.md
@@ -60,7 +60,7 @@ branch — there are no remote branches nor merge requests to inspect.
    SENTINEL="$(git rev-parse --git-common-dir)/roar-last-sha"
    ROWS=""
    if [ -f "$SENTINEL" ] && git merge-base --is-ancestor "$(cat "$SENTINEL")" HEAD; then
-       ROWS="$(~/.claude/skills/roar/enumerate-merges.py "$(cat "$SENTINEL")" --project "<name>")"
+       ROWS="$(~/.claude/skills/roar/enumerate-merges.py "$(cat "$SENTINEL")" --project "<name>")" || ROWS=""
    fi
    if [ -n "$ROWS" ]; then
        # Per-PR path: one telemetry-equivalent record per merged PR.
--- a/skills/roar/SKILL.md
+++ b/skills/roar/SKILL.md
@@ -50,9 +50,28 @@ branch — there are no remote branches nor merge requests to inspect.
 ## Reflect and update
 
 1. **Reflect**: what worked, what didn't, what was surprising.
-2. **Log to telemetry**: pipe a JSON summary to `~/.claude/skills/roar/log-celebration`:
+2. **Log to telemetry**: log one celebration per merged PR since the sentinel,
+   falling back to a single aggregate entry when no merge commits are found.
+   A batched interactive session merges several PRs then roars once, so a single
+   aggregate blob loses per-ticket attribution (ticket 0331). Enumerate the merge
+   commits in `roar-last-sha..HEAD` and log each as its own record when the
+   sentinel exists, is an ancestor of HEAD, and the enumeration is non-empty:
    ```bash
-   echo '{"project":"<name>","branch":"<branch>","commits":<n>,"files_changed":<n>,"ticket":<number|null>}' | ~/.claude/skills/roar/log-celebration
+   SENTINEL="$(git rev-parse --git-common-dir)/roar-last-sha"
+   ROWS=""
+   if [ -f "$SENTINEL" ] && git merge-base --is-ancestor "$(cat "$SENTINEL")" HEAD; then
+       ROWS="$(~/.claude/skills/roar/enumerate-merges.py "$(cat "$SENTINEL")" --project "<name>")"
+   fi
+   if [ -n "$ROWS" ]; then
+       # Per-PR path: one telemetry-equivalent record per merged PR.
+       printf '%s\n' "$ROWS" | while IFS= read -r row; do
+           printf '%s\n' "$row" | ~/.claude/skills/roar/log-celebration
+       done
+   else
+       # Aggregate fallback: first roar, rewritten history, squash-merge, or
+       # a no-forge repo where no merge commits exist.
+       echo '{"project":"<name>","branch":"<branch>","commits":<n>,"files_changed":<n>,"ticket":<number|null>}' | ~/.claude/skills/roar/log-celebration
+   fi
    ```
    Then write the current HEAD SHA to the sentinel:
    ```bash
--- a/skills/roar/enumerate-merges.py
+++ b/skills/roar/enumerate-merges.py
@@ -36,7 +36,7 @@ PR_SUBJECT_RE = re.compile(r"^Merge pull request #(\d+) from [^/]+/(.+)$")
 # so a PR with no close commit (Ticket: none, non-erg merge) would otherwise
 # mis-attribute a filing commit's id. If the close format changes, update this.
 CLOSE_COMMIT_RE = re.compile(
-    r"^ticket\((\d+)(?:,\s*\d+)*\): close and archive\b.*PR #\d+$"
+    r"^ticket\((\d+)(?:,\s*\d+)*\): close and archive — PR #\d+$"
 )
 
 
--- a/skills/roar/enumerate-merges.py
+++ b/skills/roar/enumerate-merges.py
@@ -80,21 +80,23 @@ def branch_from_subject(subject: str) -> str | None:
     return m.group(2) if m else None
 
 
-def ticket_from_range(parent1: str, parent2: str, root: str) -> int | None:
+def commits_and_ticket(parent1: str, parent2: str, root: str) -> tuple[int, int | None]:
+    """Return (commit count, ticket id) for a two-dot range in one git call.
+
+    The subject list `git log` returns while scanning for the close commit
+    already has one line per commit, so its length is the commit count — no
+    separate `rev-list --count` spawn needed.
+    """
     subjects = git(
         ["log", "--format=%s", f"{parent1}..{parent2}"], cwd=root
     ).splitlines()
+    ticket = None
     for subject in subjects:
         m = CLOSE_COMMIT_RE.match(subject.strip())
         if m:
-            return int(m.group(1))
-    return None
-
-
-def commits_in_range(parent1: str, parent2: str, root: str) -> int:
-    return int(
-        git(["rev-list", "--count", f"{parent1}..{parent2}"], cwd=root).strip()
-    )
+            ticket = int(m.group(1))
+            break
+    return len(subjects), ticket
 
 
 def files_changed(parent1: str, parent2: str, root: str) -> int:
@@ -115,13 +117,14 @@ def build_records(since_sha: str, project: str, root: str) -> list[dict]:
             # --merges guarantees >= 2 parents; skip defensively otherwise.
             continue
         p1, p2 = parents[0], parents[1]
+        commits, ticket = commits_and_ticket(p1, p2, root)
         records.append(
             {
                 "project": project,
                 "branch": branch_from_subject(subject),
-                "commits": commits_in_range(p1, p2, root),
+                "commits": commits,
                 "files_changed": files_changed(p1, p2, root),
-                "ticket": ticket_from_range(p1, p2, root),
+                "ticket": ticket,
             }
         )
     return records
--- a/skills/roar/enumerate-merges.py
+++ b/skills/roar/enumerate-merges.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""Enumerate merged PRs since a sentinel SHA as celebration records (ticket 0331).
+
+Roar's telemetry step logs one celebration per merged PR instead of a single
+aggregate blob. This script enumerates the merge commits in ``<since-sha>..HEAD``
+on the current branch and emits, per merge, one JSON object per line carrying the
+fields roar's ``log-celebration`` expects (that helper stamps ``ts``/``date``):
+
+    {"project": str, "branch": str|null, "commits": int,
+     "files_changed": int, "ticket": int|null}
+
+Branch is parsed from the GitHub-shaped merge subject; ticket is recovered by
+scanning the merge's second-parent commit range for the erg-pr-merge close
+commit, NOT from the branch name (real branch names such as
+``worktree-agent-a31bf6655c104744e`` carry misleading digits).
+
+An empty range prints nothing and exits 0. The sentinel-missing / non-ancestor
+guards live in roar's SKILL.md prose, not here.
+"""
+
+import argparse
+import json
+import os
+import re
+import subprocess
+import sys
+
+# GitHub-shaped merge subject: "Merge pull request #123 from owner/branch-name".
+PR_SUBJECT_RE = re.compile(r"^Merge pull request #(\d+) from [^/]+/(.+)$")
+
+# erg-pr-merge writes its close commit as
+#   ticket(<id>[, <id>...]): close and archive — PR #<n>
+# (skills/merge/erg-pr-merge ~line 228; CLOSED_LIST joins IDs with ", ").
+# If that commit format changes, update this regex to match.
+CLOSE_COMMIT_RE = re.compile(r"^ticket\((\d+)(?:,\s*\d+)*\):")
+
+
+def git(args: list[str], cwd: str | None = None) -> str:
+    return subprocess.run(
+        ["git", *args], cwd=cwd, capture_output=True, text=True, check=True
+    ).stdout
+
+
+def repo_root() -> str:
+    # Resolve at call time — never at import (worktree-safe; see memory
+    # feedback_module_level_git_paths).
+    return git(["rev-parse", "--show-toplevel"]).strip()
+
+
+def merge_commits(since_sha: str, root: str) -> list[str]:
+    out = git(
+        [
+            "log",
+            "--merges",
+            "--first-parent",
+            "--reverse",
+            "--format=%H",
+            f"{since_sha}..HEAD",
+        ],
+        cwd=root,
+    )
+    return [line for line in out.splitlines() if line.strip()]
+
+
+def merge_meta(sha: str, root: str) -> tuple[list[str], str]:
+    """Return (parent SHAs, subject) for a merge commit."""
+    out = git(["show", "-s", "--format=%P%n%s", sha], cwd=root).splitlines()
+    parents = out[0].split() if out else []
+    subject = out[1] if len(out) > 1 else ""
+    return parents, subject
+
+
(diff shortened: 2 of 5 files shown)


MinhHaDuong (commented):
## /review-pr — 5-agent panel (correctness, consistency, scope, red team, doc propagation)

**Panel verdict: request-changes.** 3 agents approved (consistency, scope, doc propagation); 2 requested changes (correctness, red team) on one bug they reproduced independently and I re-confirmed. Dissent preserved below — author decides.

### BLOCKER — `files_changed` over-counts in the exact scenario this PR targets
`enumerate-merges.py:96` computes `git diff --stat parent1..parent2` (two-dot = raw endpoint comparison), not the PR's own change set. When a branch was cut from an older `main` and other PRs merged in between — precisely the "batched interactive session merges several PRs one at a time" case ticket 0331 exists to serve, and a shape `rules/git.md` explicitly sanctions ("skip the rebase and let the merge land as-is") — the diff picks up the intervening PR's files as phantom changes.

Reproduced (correctness + red team + me), real git, real CLI entry point:
```
# PR-A (feat-a) adds a.txt, merges. PR-B (fix-widget) cut from PRE-PR-A base,
# adds b.txt + close commit, merges second.
$ enumerate-merges.py $BASE --project demo
{"branch":"fix-widget","commits":2,"files_changed":3,"ticket":42}
# PR-B actually changed 2 files; a.txt (PR-A's) leaks in as a phantom removal.
$ git diff --stat $BASE...feat-b   # three-dot, merge-base relative
 2 files changed, 2 insertions(+)   # correct value = 2
```
`commits` and `ticket` are correct (reachability-based); only `files_changed` is wrong, producing a plausible-looking but incorrect number with no crash and no fallback — which directly undermines exit criterion 1 ("telemetry-equivalent to a raid's per-PR roars"). Fix: use three-dot `parent1...parent2` in `files_changed`, and add a regression test with an interleaved-merge fixture (the current suite never triggers this — every `merge_pr` branches off the already-advanced `main`, so parent1 is always an ancestor of parent2). One-line fix.

### verifiable: a mid-batch git failure silently discards the whole batch
`build_records()` accumulates all records before `main()` prints anything, and SKILL.md's `ROWS="$(...)"` never checks the exit code. If any single git call raises `CalledProcessError` (missing object, shallow/partial-clone boundary), every already-computed record is lost, the script exits 1 to stderr, stdout is empty, and the empty `ROWS` is indistinguishable from "no merges" — so it silently routes to the single aggregate fallback. SKILL.md's `--is-ancestor` guard checks reachability, not that every downstream git call on the range succeeds. (Trigger is uncommon in a full clone; the cheap hardening is to print each record as it is computed, and/or check `$?` in SKILL.md before deciding per-PR vs aggregate.)

### consider: octopus merges drop parents beyond the second
`build_records` uses only `parents[0]`/`parents[1]`; a >2-parent merge silently loses the extra parents' tickets and counts (exit 0, record looks complete). This repo's 50 merges are all 2-parent, so no current trigger — worth a defensive skip-with-warning rather than silent truncation.

### consider: multi-ticket close records only the first id
`CLOSE_COMMIT_RE` captures `group(1)` only, so `ticket(0042, 0043): ...` records `ticket: 42` and drops 43 (the `ticket` field is a scalar `int|null`). Deliberate and tested (`test_multi_ticket_close_takes_first_id`), but undocumented as a limitation.

### nofollow: minor coverage gaps
Dead `len(parents) < 2` branch (unreachable under `--merges`), the `files_changed` no-summary-line path, and the `--project` basename default are all untested; harmless.

### Non-issues explicitly probed and cleared
Shell/command injection via branch name or merge subject (json.dumps-escaped, log-celebration parses via json, no eval) — did not fire. `files_changed` regex on rename-only / mode-only / binary diffs — git always emits the `N file(s) changed` summary; handled. Python style, JSON record shape vs `log-celebration`, the `erg-pr-merge ~line 228` / CLOSED_LIST comment, sentinel-path derivation, executable bits, `make check-skills-drift`, and the "nothing reads celebrations.jsonl programmatically / lair reads only the sentinel" claims — all verified accurate.

### Dissent (preserved)
Consistency, scope, and doc-propagation each returned **approve**; none probed the stale-based-branch merge scenario. Correctness and red team returned **request-changes** on the `files_changed` bug. Full suite green (840 passed; the 5 new tests pass) — but the suite does not cover the failing case, which is why it stays green.

*Panel: sonnet ×5, foreground. Tests run: 5 new pass, full suite 840 pass.*

MinhHaDuong (commented):
## Multi-perspective review — 5 agents (correctness, consistency, scope, red team, doc propagation)

**Scope, consistency, and doc-propagation reviewers: clean approve.** The Python conforms to the harness style rules (built-in generics, argparse, call-time repo-root resolution), the file footprint is exactly the three files the ticket needs, the `+x` bit is set, the field schema matches `log-celebration`, and every cross-reference (`erg-pr-merge:228`, the `ts`/`date` stamping claim, the sentinel read/write path) was verified accurate. New tests pass 6/6; CI pytest-guard is green on this SHA. Nice anti-tautology test on the branch-name-vs-close-commit case, and the three-dot `files_changed` fix for interleaved batched merges is well-targeted.

One correctness issue blocks, plus a few minor notes. Dissent preserved: the correctness reviewer voted request-changes on the blocker; the red-team reviewer rated the same defect a non-blocking comment (local, non-authoritative telemetry). **Author decides** — the fix is one line if you agree.

---

### Blocker — ticket recovery matches *any* `ticket(NNNN):` commit, not only erg-pr-merge's close commit

`skills/roar/enumerate-merges.py:36` — `CLOSE_COMMIT_RE = r"^ticket\((\d+)(?:,\s*\d+)*\):"` checks only the `ticket(NNNN):` prefix, never the `close and archive — PR #<n>` body that `erg-pr-merge:228` actually writes. In this repo `ticket(NNNN): <description>` is *also* the ordinary convention for commits that file or edit tickets (16 of the last 100 commits on `main` use the non-close form, e.g. `1529f59 ticket(0329): file seat-runner suite source-grep flake`). `ticket_from_range` returns the first match in `git log`'s newest-first order.

The normal path is safe: when erg-pr-merge writes its close commit it is the newest commit on the branch, so it wins even against a competing older `ticket(NNNN):` commit — verified, returns the right ticket. The misfire is bounded to PRs with **no close commit** (a `Ticket: none` PR, the `ERG_PR_MERGE_ALLOW_MISSING_TICKET` skip path, a non-erg merge) that happen to carry a ticket-prefixed dev commit. There, a raid's per-PR roar would log `ticket: null`; this script logs a wrong number — not telemetry-equivalent, which is the ticket's exit criterion.

Reproduced (PR branch script, no close commit present, one `ticket(0329):` dev commit):
```
$ python3 enumerate-merges.py <base> --project demo
{"project": "demo", "branch": "feat-0330", "commits": 2, "files_changed": 2, "ticket": 329}   # correct value is null
```
Independently reproduced by the red-team reviewer with an unrelated docs commit `ticket(1234): document…` merged as a PR that never touched 1234 → recorded `ticket: 1234`.

**Fix direction:** require the close template, e.g. anchor on `: close and archive` or `— PR #`, so only erg-pr-merge's own commit matches. Add a regression test for the inverse of the existing anti-tautology case: close commit *absent*, a foreign `ticket(NNNN):` dev commit *present* → expect `null`.

---

### Minor findings

- **verifiable:** `commits` (two-dot `p1..p2`, line 96) and `files_changed` (three-dot `p1...p2`, line 102) count different bases. On a branch that reconciled via `git merge origin/main` mid-development (the multi-PR-wave idiom in `rules/git.md`), three-dot `files_changed` correctly excludes the intervening PR's files, but two-dot `commits` counts the reconcile merge commit itself, giving a +1 skew — the two fields describe different things for the same record. No test covers `commits` in the interleaved/reconciled scenario. (correctness)

- **consider:** The SKILL.md step-2 snippet degrades gracefully on a script crash *only* under plain bash: `ROWS="$(enumerate-merges.py …)"` captures empty stdout → aggregate fallback. If a future edit hardens the snippet with `set -euo pipefail` (house style per `rules/coding-bash.md`), that assignment inside the `if…then` body is **not** `-e`-exempt and would abort roar step 2 instead of falling back — whereas the separate `merge-base --is-ancestor` guard *is* exempt (it's an if-condition). Asymmetric, latent landmine. Consider a comment pinning the no-`set -e` assumption, or make the fallback explicit. (red team)

- **consider:** `files_changed` parses the English summary line `r"(\d+) files? changed"` from `git diff --stat`. Under a git build with localized message catalogs this line translates and the regex silently returns 0. Cheap hardening: run the stat with `LC_ALL=C` (or `git -c`). Unconfirmed in the sandbox (no `.mo` catalogs), hence consider not verifiable. (red team)

- **nofollow:** the `~line 228` pointer to `erg-pr-merge` is brittle but verified accurate today and already self-flagged with an "update if the format changes" note. (consistency)

- **nofollow:** `len(parents) < 2` (line ~119) is unreachable under `--merges`; harmless documented defensive guard. (scope)

- **nofollow:** octopus merges (3+ parents) ignore parents beyond the second — not reachable for two-parent GitHub PR merges. (red team)

- **nofollow:** shell/JSON injection via branch names, em-dash / non-ASCII in close commits, and leading-zero / empty ticket ids were all attacked and are safe (`git --format=%s` folds newlines, `json.dumps` escapes, regex anchored at `:`). (red team)

MinhHaDuong (commented):
## `/review-pr` panel — 5 agents (correctness, consistency, scope, red team, doc propagation)

**Overall: approve with comments.** Four perspectives approve; the red team returns request-changes. I investigated its two HIGH findings and downgraded both (neither is a regression this PR introduces — see below), so I am not posting a blocking verdict. Dissent is preserved verbatim; the author decides. New test suite: 7/7 pass in a clean clone. `enumerate-merges.py` is mode 100755, shebang present. Close-commit regex matches `erg-pr-merge:228` byte-for-byte (em dash confirmed).

### Dissent — Red team verdict: **request-changes** (preserved verbatim, then adjudicated)

**RT-HIGH-1 — stale local HEAD could defeat the fix.** *"`enumerate-merges.py` enumerates `since_sha..HEAD` against whatever the current checkout's HEAD happens to be... If a supervisor session merges 3 sibling PRs via `gh pr merge` from a primary checkout parked on local main, local main's tip never moves — only origin/main does... the range is empty, ROWS="", and the code falls to the single aggregate blob — silently reproducing the exact attribution loss ticket 0331 exists to fix."*

> Adjudication (downgrade to `consider:`): this is a **pre-existing precondition of roar's telemetry step, not a regression**. The old aggregate code ran the identical `roar-last-sha..HEAD` range against the same HEAD; in a stale-HEAD checkout it too would have logged an empty/degenerate record. In the actual motivating incident (PRs #538–#548) HEAD *did* contain the merges — the old code enumerated 51 commits — so the new per-PR path works there. Roar's pre-check fetches and checks ancestry but never ff-syncs HEAD; that assumption predates this PR. Worth a follow-up (have step 2 `git merge --ff-only origin/main` before enumerating, or document the HEAD-currency requirement), but not blocking.

**RT-HIGH-2 — multi-ticket close commits drop all but the first ID.** *"`ticket(0042, 0043): close and archive — PR #123` → group(1) == "0042". Ticket 0043 is silently dropped. Multi-ticket-closing PRs are an explicitly supported pattern."*

> Adjudication (downgrade to `consider:`): the celebration record schema is scalar (`ticket: int|null`, consumed by the untouched `log-celebration`). The old aggregate logged `ticket null`, so first-ticket capture is **strictly better than status quo**, and the behavior is test-documented (`test_multi_ticket_close_takes_first_id`). Fully capturing both would require a schema change across consumers; ticket 0331's scope is cross-PR attribution (11 PRs → 11 records), not intra-PR multi-ticket. Not a correctness break.

### Findings worth acting on (non-blocking)

- `consider:` (red team, verifiable) — **CLOSE_COMMIT_RE is looser than the documented template.** `"ticket(99): close and archive stray note, ref PR #7"` matches and yields ticket 99, because `.*` bridges "archive" and the trailing `PR #\d+$`. The real format has nothing between (`close and archive — PR #N`). Tightening to `close and archive — PR #\d+$` removes the mis-attribution window. `skills/roar/enumerate-merges.py` CLOSE_COMMIT_RE.
- `consider:` (red team) — **one failing git call discards the whole batch.** `build_records` uses `check=True` with no per-record guard; a single transient git failure raises, `ROWS="$(...)" || ROWS=""` resets to empty, and the entire batch degrades to one aggregate blob. A per-record try/except would preserve partial attribution.
- `consider:` (red team) — **shared sentinel has no lock.** `roar-last-sha` lives under `--git-common-dir` (shared across worktrees). Two racing roars can enumerate overlapping ranges and double-log; last writer clobbers the other's sentinel advance. Pre-existing mechanism, but this PR is what turns it into duplicate `celebrations.jsonl` records. Follow-up ticket candidate.
- `consider:` (consistency) — **"PR" vs "merge request" wording.** New step-2 prose says "PR" 4×; the repo's tool-agnostic convention prefers "merge request" (rules/workflow.md § tool-agnostic language). The file already mixes both, so low stakes.
- `nofollow:` (red team, correctness, scope) — **octopus merges (>2 parents) silently skipped/truncated.** Not producible by the forge's PR-merge; dead path in practice.
- `nofollow:` (red team) — **greedy `(.+)$` keeps trailing whitespace in branch.** Cosmetic; GitHub `%s` subjects don't carry it.

### Merge-readiness nit
- `verifiable:` — branch is **31 commits behind `origin/main`** (`git rev-list --count t331..origin/main` = 31; mergeStateStatus BLOCKED). Rebase onto current origin/main at the merge gate before landing.

### What the panel confirmed solid
- Two-dot (`commits_in_range`, reachability) vs three-dot (`files_changed`, merge-base-relative tree diff) is deliberate and correct — the interleaved-batched-merge trap is exercised by `test_files_changed_excludes_interleaved_merges`.
- Anti-tautology core case holds: a no-digit branch (`fix-flaky-widget`) with a close commit still recovers ticket 42; a filing commit `ticket(0329): ...` is correctly *not* attributed.
- `repo_root()` resolved at call time, not import (worktree-safe). Scope is tight: exactly the 3 declared files; `log-celebration` and `erg-pr-merge` untouched. Python style matches `coding-python.md` (built-in generics, argparse, lean main).

MinhHaDuong:
## /gaze actions

round: 2
tier: full
adherence: PASS — 0 blocking
review: 0 blockers (octopus-merge minor, now fixed)
review-pr: 0 blockers (octopus-merge verifiable minor, now fixed)
simplify: 1 fix applied (merged ticket/commit-count lookups into one git call — commit 15625fc)
fix agent: 2 commits (round 2) — octopus-merge guard + regression test + ticket bump (e03f8f1, 5128a23)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Batched interactive roar produces per-PR celebration entries telemetry-equivalent to a raid's per-PR roars: ADDRESSED — `enumerate-merges.py` emits one record per first-parent merge; `files_changed` uses three-dot (merge-base-relative) so intervening batched PRs do not leak in, and `ticket` is recovered by scanning the second-parent range for the erg-pr-merge close template — matching a raid's per-PR roar record. Covered by `test_files_changed_excludes_interleaved_merges`, `test_ticket_from_close_commit_not_branch_name`, `test_filing_commit_not_mistaken_for_close`.
- Aggregate fallback preserved where merge commits are absent: ADDRESSED — `SKILL.md` step 2 keeps the byte-identical aggregate `log-celebration` echo in the `else` branch; guarded on sentinel-missing, non-ancestor, and empty enumeration. Covered by `test_empty_range_emits_nothing`.
- `make check` passes: ADDRESSED — CI 9/9 green on 5128a23 (pytest-guard SUCCESS); local adherence 17 pass, new module 8 pass.

Unresolved review comments: none. The three earlier review panels' blockers (two-dot `files_changed` over-count; `CLOSE_COMMIT_RE` matching any `ticket(NNNN):` filing commit) were fixed in commits e49f478 / 13b348f / 04851fe with dedicated regression tests, and re-verified holding by a fresh full battery.
Unresolved simplify: none.
Adherence: PASS (0 blocking).

Rationale:
Round 1 flagged one fresh `verifiable:` minor — `build_records` unpacked only the first two parents, silently undercounting a 3+-parent octopus merge (unreachable via the forge's 2-parent PR merges, but the script is a documented standalone/no-forge CLI where octopus merges occur). Round-2 fix e03f8f1 adds a symmetric `len(parents) > 2` skip-with-stderr-warning mirroring the existing `len(parents) < 2` guard, a docstring caveat, and a RED-proven regression test (`test_octopus_merge_skipped_not_undercounted`, confirmed failing pre-fix). The `ROWS` all-or-nothing degrade and multi-ticket-first-id notes are dispositioned `consider:` — both fail safe to prior behavior, carry no attached failing assertion, and were already adjudicated on the PR. Scope is exactly the three declared files plus the ticket bump; `log-celebration` and `erg-pr-merge` untouched.

telemetry: tier=full wall=~35min agents=5 tokens=na cost~=na

PR #594, opened 2026-07-14
fix(0332): replace SIGPIPE-class grep|grep pipe in rules-injection suite

MinhHaDuong:
## What

`tests/test_harness_rules_injection.sh:44` asserted that `scripts/on-start.sh` contains a `cat ... README.md` invocation via a two-stage pipe:

```bash
if ! grep -F 'cat' scripts/on-start.sh | grep -qF 'README.md'; then
```

This was a same-line co-occurrence check — `on-start.sh:33` pairs `cat` and `README.md` on one line. Under `set -o pipefail`, the trailing `grep -q` exits on first match and can close the pipe before the upstream `grep -F` finishes writing; the upstream then dies with SIGPIPE (141), which pipefail reports as the pipeline status — a nondeterministic false FAIL. Same defect class fixed suite-wide in PR #575 (printf|grep) and PR #584 (grep-here-string); this site greps a FILE through a filter pipe, so it was out of scope for both.

## Change

One anchored grep, no pipe, asserting the same invariant:

```bash
if ! grep -qE 'cat[^#]*README\.md' scripts/on-start.sh; then
```

## Sweep result

Line 44 was the **only** `grep | grep -q` pipe in `tests/test_harness_rules_injection.sh` (re-verified with a single anchored grep over the suite; no siblings).

## Mutation outcome

On a throwaway copy of `scripts/on-start.sh` with `README.md` dropped from the cat line, the converted check FAILs loudly (`FAIL: scripts/on-start.sh missing 'cat ... README.md' invocation`); against the pristine file it passes. Mutation never committed.

## Verification

- Suite green standalone (`bash tests/test_harness_rules_injection.sh`)
- `make check-fast` green
- `make check` green
- Fail-loud preserved; no retry loop, no `|| true` on the assertion.

**Ticket:** tickets/0332-sigpipe-class-pipe-in-rules-injection-suite.erg

### DIFF
--- a/tests/test_harness_rules_injection.sh
+++ b/tests/test_harness_rules_injection.sh
@@ -41,7 +41,7 @@ for needle in "${body_strings[@]}"; do
 done
 
 # 3. on-start.sh must literally contain a `cat ... README.md` invocation.
-if ! grep -F 'cat' scripts/on-start.sh | grep -qF 'README.md'; then
+if ! grep -qE 'cat[^#]*README\.md' scripts/on-start.sh; then
   echo "FAIL: scripts/on-start.sh missing 'cat ... README.md' invocation"
   fail=1
 fi
--- a/tickets/0332-sigpipe-class-pipe-in-rules-injection-suite.erg
+++ b/tickets/0332-sigpipe-class-pipe-in-rules-injection-suite.erg
@@ -7,6 +7,7 @@ Author: claude
 2026-07-14T12:40Z claude created
 2026-07-14T11:29Z claude note imagine: single grep|grep site confirmed (no siblings in suite); the pipe is a same-line cooccurrence check — on-start.sh:33 pairs cat and README.md on one line, so one anchored `grep -qE 'cat[^#]*README\.md'` against the file replaces it with no pipe; capture-then-[[ ]] dropped as ceremony; mutation check unchanged
 2026-07-14T11:40Z claude note plan+feasibility: imagine output is the plan (file:line-precise, phase-3 absorbed per raid-0252 precedent); anchors verified (suite :44, on-start.sh:33); PASS, no overlap with 0333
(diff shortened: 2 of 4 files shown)


MinhHaDuong:
## /gaze actions

round: 1
tier: small
adherence: PASS — 0 blocking
review: 0 comments (no issues; correct fix for the SIGPIPE-class defect)
review-pr: 0 blocking (correctness-only panel — approve; mutation soundness independently reproduced)
simplify: 0 fixes — nothing to simplify (already a minimal single anchored grep)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- No `grep | grep -q` pipeline remains in tests/test_harness_rules_injection.sh: ADDRESSED — commit 180b559, line 44 now `grep -qE 'cat[^#]*README\.md' scripts/on-start.sh`; anti-pattern grep over the suite returns nothing.
- Mutation check confirms the converted site still fails on needle removal: ADDRESSED — reproduced fresh in the gate (README.md→XXXX.md on a throwaway on-start.sh breaks the match, fail-loud; pristine line 33 is the only README occurrence, so no sibling masks the mutation).
- `make check` passes: ADDRESSED — suite green standalone and full `make check` exit 0.

Unresolved simplify: none
Adherence: PASS (0)

Scope note (report-only, non-blocking): the PR also appends log-only annotation lines to `tickets/0333-...erg` (imagine/plan notes from the shared 332/333 raid). This is provenance bookkeeping, not a second deliverable; the PR body correctly names only `tickets/0332` as the closing ticket.

Rationale:
All three exit criteria were independently verified against the diff and a live re-execution, not merely cited: the SIGPIPE-prone `grep|grep -q` pipe is removed, the replacement single-anchored grep was mutation-tested fresh in the gate, and the suite passes. The reviewer battery (adherence, review, correctness, simplify) surfaced only non-blocking nit/consider/nofollow items — none `verifiable:`-tagged, none requiring rework. CI is fully green and mergeStateStatus is not conflicted. Approved.

telemetry: tier=small wall=~9min agents=5 tokens=na cost~=na

PR #593, opened 2026-07-14
feat(0334): per-project least-privilege ~/.config/keys loading via KEYS=

MinhHaDuong:
## Mechanism

Adds an opt-in, default-deny least-privilege boundary to `scripts/bash-env.sh`. A project declares the credential providers it needs with `KEYS=prov1,prov2,...` in its project `$PWD/.env`. bash-env.sh then sources **only** `$HOME/.config/keys/<prov>.env` for each declared, validated provider — so a project no longer loads every secret on every bash subprocess.

- **Default-deny**: no `KEYS` line → no provider secrets loaded.
- **Name validation**: each provider name must match `^[a-z0-9-]+$` before any filesystem access. Path traversal (`../../etc/passwd`), slashes (`foo/bar`), uppercase, and empty names are skipped with a stderr warning and can never source an arbitrary path. Globbing is disabled during the comma split.
- **Trusted vs untrusted**: the per-provider `~/.config/keys/<name>.env` files are user-owned and TRUSTED, so they are sourced as shell code (full expansion) — in deliberate contrast to the untrusted project `.env`, which stays strict-parsed and never executed (builds on merged #588). `KEYS` itself arrives as a plain non-GUARD value through that strict parser.
- **Non-fatal**: a declared-but-missing provider warns and continues. The block is `set -euo pipefail`-safe (`${KEYS:-}` guarded, no `((...))`).

## Tests — `tests/test_bash_env_keys_selection.sh` (new, hermetic)

Fake `HOME` temp dir + fake `~/.config/keys/*.env` fixtures holding **non-secret sentinels** (`FAKE_OPENROUTER=orval`, etc.) + crafted project `.env`; bash-env.sh is sourced in an isolated subshell and the env is asserted. No real `~/.config/keys` file is ever read.

1. `KEYS=openrouter` → `FAKE_OPENROUTER` set, undeclared `FAKE_MISTRAL` **not** set.
2. No `KEYS` line → neither loaded (default-deny).
3. `KEYS=openrouter,zenodo` → both load (+ a whitespace-trimming variant `KEYS= openrouter , zenodo`).
4. `KEYS=../../etc/passwd` and `KEYS=foo/bar` → rejected by validation, nothing sourced, shell survives (rc 0).
5. `KEYS=nonexistent` → warns, shell survives, nothing spurious loaded.

[...]

PR #595, opened 2026-07-14
fix(0335): bash-env.sh robustness — CRLF strip, size cap, realpath set -e safety

MinhHaDuong:
Follow-ups from the #588 (strict-parse) and #593 (KEYS=) gaze reviews on `scripts/bash-env.sh`. Three non-blocking robustness nits, batched (all bash-env.sh-adjacent).

## Fixes
1. **CRLF strip** — the strict parser drops a trailing `\r` from each line, so a Windows-edited project `.env` (CRLF terminators) yields clean values. The `KEYS=` provider list was already covered by its existing whitespace trim (`[:space:]` includes `\r`); the line-level strip makes it belt-and-suspenders and cleans all other values.
2. **Size cap** — a project `.env` past 256 KiB (`wc -c`) is skipped whole with a stderr warning (`bash-env: project .env exceeds size cap …`) rather than parsed, so a pathological or adversarial file cannot tax every bash subprocess. This script is sourced on every Bash tool call, and the byte check is cheap. Skipped whole, never partially parsed (no desynced read).
3. **realpath `set -e` safety** — the dedup `realpath` command substitutions now use `|| true`, so an absent `~/.claude/.env` (its parent dir missing) cannot abort the script under an already-active `set -e` in the sourcing shell. Dedup behavior is unchanged when both files exist; an empty result still compares unequal, so the project file is still parsed when either path is absent.

## Tests
New `tests/test_bash_env_robustness.sh` (auto-discovered and run by `tests/test_bash_suites.py`), 8 assertions:
- CRLF value → `\r` stripped (**RED pre-fix**)
- CRLF `KEYS=` line resolves the provider (green pre-fix — already covered by the KEYS whitespace trim; kept as a regression guard)
- oversized `.env` → parse skipped, warning on stderr, shell survives (**RED pre-fix**: file was parsed, no warning)
- `set -e` + missing user `.env` → source reaches its end (**RED pre-fix**: aborted at the realpath dedup)
- `set -e` + no `.env` at all → source reaches its end (green pre-fix — block was skipped)

The oversize fixture builds ~600 KiB by string-doubling, with no `yes | head` pipe, to avoid the SIGPIPE-under-`pipefail` trap (ticket 0332 class).

## Memory
Refreshes `projects/…/memory/project_bash_env_secret_loading.md` to current reality: two-trust-level sources (user `~/.claude/.env` sourced; project `$PWD/.env` strict-parsed, never sourced, `GUARD_*`/`_GUARD_*` refused), `KEYS=` least-privilege provider loading from `~/.config/keys/`, and these robustness properties.

## Gates
`make lint`, `make check` (847 passed), all three bash-env shell suites, and `erg check` all green.

**Ticket:** tickets/0335-bash-env-robustness-crlf-sizecap-realpath.erg

### DIFF
--- a/projects/-home-haduong--claude/memory/project_bash_env_secret_loading.md
+++ b/projects/-home-haduong--claude/memory/project_bash_env_secret_loading.md
@@ -8,9 +8,16 @@ Secrets are loaded into Claude Code bash subprocesses via `BASH_ENV`, not `CLAUD
 
 **Why:** `CLAUDE_ENV_FILE` causes Claude Code to inline `KEY=VALUE` pairs as shell assignments before each Bash tool call (visible in `ps -ef` to all local users). Discovered 2026-05-03 when orphaned processes from April 30 were found exposing all API keys.
 
-**How to apply:** 
+**Two .env sources, DIFFERENT trust** (`scripts/bash-env.sh`, sourced on every subprocess):
+- `~/.claude/.env` — user-owned, TRUSTED: **sourced** as shell code (full expansion), inside `set -a`/`set +a`.
+- `$PWD/.env` — project-level, UNTRUSTED (an agent/clone/project write can place it): **strict-parsed** `KEY=VALUE`, never sourced. Values assigned LITERALLY (no eval/`$()`/backticks), one surrounding quote-pair stripped, non-identifier keys skipped. Any guard-namespaced key (substring `GUARD_`, covering `GUARD_*` and the `_GUARD_*` override pair) is REFUSED, so it cannot forge a per-process guard nonce or the worktree-path override.
+
+**Provider secrets load least-privilege via `KEYS=`** (#588/#593): the project `.env` opts in with `KEYS=name,name`; only those validated (`^[a-z0-9-]+$`) providers' user-owned `~/.config/keys/<name>.env` are sourced. Default-deny (no `KEYS` line → no provider secrets); name validation blocks path traversal out of `~/.config/keys/`. This is cooperative scoping (shrinks default exposure), NOT access control against a hostile `.env`.
+
+**Robustness (ticket 0335):** strict parse tolerates CRLF (trailing `\r` dropped), is bounded by a 256 KiB byte cap (oversized project `.env` skipped whole with a stderr warning), and the realpath dedup uses `|| true` so an absent file cannot abort the script under an active `set -e` in the sourcing shell.
+
+**How to apply:**
 - `settings.json` → `env.BASH_ENV` → `scripts/bash-env.sh`
-- `bash-env.sh` uses `set -a` / `set +a` to source `~/.claude/.env` and `$PWD/.env`
-- `bash-env.sh` must NOT have `set -euo pipefail` (it's sourced, not executed — flags would propagate into calling shell)
-- `bash-env.sh` is excluded from the `pipefail-guard` CI check alongside `shell-init.sh`
-- Never write secrets back to `CLAUDE_ENV_FILE` in hooks
+- `bash-env.sh` must NOT have `set -euo pipefail` (it's sourced, not executed — flags would propagate into the calling shell); excluded from the `pipefail-guard` CI check alongside `shell-init.sh`.
+- Never write secrets back to `CLAUDE_ENV_FILE` in hooks.
+- Tests: `tests/test_bash_env_project_env_parse.sh`, `tests/test_bash_env_keys_selection.sh`, `tests/test_bash_env_robustness.sh` (auto-run by `tests/test_bash_suites.py`).
--- a/scripts/bash-env.sh
+++ b/scripts/bash-env.sh
@@ -24,7 +24,11 @@
 #
 # The script runs on every bash subprocess, so it stays fast and never aborts on
 # a missing or malformed .env — bad lines in the project file are skipped, never
-# fatal.
+# fatal, and it stays safe under an already-active `set -e` in the sourcing shell
+# (the realpath dedup below is guarded with `|| true`). The strict parse tolerates
+# CRLF (a trailing \r is dropped, for Windows-edited files) and is bounded: a
+# project .env past a generous byte cap is skipped whole, so a pathological or
+# adversarial one cannot tax every subprocess.
 #
 # Provider-secret scoping (KEYS=). A project declares which credential providers
 # it needs by setting KEYS=<name,name,...> in its project .env (parsed by the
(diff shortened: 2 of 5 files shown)


MinhHaDuong:
## Multi-perspective review — PR #595

Five reviewers ran in parallel (correctness, consistency, scope, red-team, doc-propagation) against `scripts/bash-env.sh`'s CRLF/size-cap/realpath robustness fixes (ticket 0335). Verified independently: `make lint` (17 passed), `make check` (847 passed), and `tests/test_bash_env_robustness.sh` run directly (8/8 pass).

**Consensus: approve.** No blockers. Consistency, scope, and doc-propagation found zero defects — variable naming/cleanup, comment style, and test idiom all match the two sibling `test_bash_env_*.sh` suites; every diff hunk traces to a named exit criterion in tickets/0335; the header comment, memory-note refresh, ticket text, and PR body are all mutually consistent (256 KiB cap number, `|| true` guard, 8-assertion count, and the `tests/` file listing all check out).

Correctness and red-team independently converged on the same non-blocking edge case, so it's promoted above the rest:

### consider: realpath double-failure could silently skip a real project `.env`
`scripts/bash-env.sh:67-69`. If **both** `realpath` calls fail (e.g. `realpath` missing/non-functional in a minimal environment), `_be_proj` and `_be_user` both become `""` via `|| true`. Since `[ "" != "" ]` is false, the dedup check treats them as "same file" and **silently skips** parsing the project `.env` — even though `[ -f "$PWD/.env" ]` (line 61) already proved it exists and should be parsed. No warning is emitted. Correctness rated this medium confidence (mechanism certain, trigger-environment likelihood uncertain); red-team assessed it as non-exploitable for privilege escalation (worst case: the attacker's own directives, or a user-relevant warning, get silently dropped — never elevated access). Net: a real robustness gap, not a security hole. Suggest a follow-up ticket: warn to stderr when both realpaths are empty but `$PWD/.env` demonstrably exists.

### consider: size cap bounds read cost, not parse cost
`scripts/bash-env.sh:74-84`. The 256 KiB cap is enforced via `wc -c`, but a file *at* the cap with maximal line count (~262K one-byte lines) still runs the full per-line `case`/parameter-expansion loop uncapped — red-team estimated low-single-digit seconds of pure-bash parsing in the worst case. This is a **net improvement** over pre-PR (fully unbounded file size and line count), not a regression, but the header comment's claim that a pathological file "cannot tax every subprocess" is only partly true — it bounds I/O, not CPU. Suggest a follow-up ticket adding a line-count cap alongside the byte cap.

### nofollow: unquoted-value literal-trailing-CR data loss
`scripts/bash-env.sh:84`. An LF-only file whose unquoted value legitimately ends in a literal `\r` byte (not a CRLF line ending) has that CR silently stripped — the script can't distinguish "CRLF terminator" from "value ends in CR". Correctness-only finding, narrow and low-likelihood; not raised as an exit-criteria gap.

### nofollow: TOCTOU between size check and parse read
`scripts/bash-env.sh:75` vs `:122`. Two independent opens of `$PWD/.env` (one for `wc -c`, one for the parse loop) mean a file swapped in that window could bypass the cap. Requires an active racing writer at that exact path/window; not attacker-relevant beyond the existing untrusted-`.env` threat model, which already assumes the adversary can rewrite the file at will between any two subprocess invocations.

No findings from consistency, scope, or doc-propagation beyond what's summarized above (all three returned clean `approve` verdicts with no unresolved items).

**Panel verdicts:** correctness=comment, consistency=approve, scope=approve, red-team=comment, doc-propagation=approve. No `request-changes` from any perspective.

MinhHaDuong:
## /gaze actions

round: 1
tier: full
adherence: PASS — 0 blocking
review: 0 comments posted (clean)
review-pr: 1 review posted (approve, 3 informational `consider:` minors)
simplify: 0 fixes (code already clean)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- CRLF strip on strict-parsed values (incl. KEYS= line): ADDRESSED — `scripts/bash-env.sh:84` `_be_line="${_be_line%$'\r'}"`; tests 1,2 PASS (RED pre-fix on 1 verified)
- Bounded project-.env parse (256 KiB cap + stderr warning): ADDRESSED — `scripts/bash-env.sh:74-78`; test 3 PASS (RED pre-fix verified)
- realpath dedup safe under active `set -e` when files absent: ADDRESSED — `scripts/bash-env.sh:67-68` `realpath … || true`; tests 4a/4b PASS (RED pre-fix on 4a verified)
- Tests cover all three fixes: ADDRESSED — `tests/test_bash_env_robustness.sh`, 8 assertions, auto-discovered by `tests/test_bash_suites.py`
- Memory doc refreshed to current reality, one concept: ADDRESSED — `project_bash_env_secret_loading.md` (trust levels, KEYS= scoping, 0335 robustness), content verified against the diff
- `make check` and `make lint` pass: ADDRESSED — `make lint` 17/17, full `pytest tests/` 847/847 re-run in-gate (not trusting CI)

Unresolved review comments: none (the 3 Agent-C `consider:` minors are informational by tag — double-realpath-failure silent skip, literal-trailing-CR mangle, parse-loop iteration count not capped; each a valid follow-up, none blocking)
Unresolved simplify: none
Adherence: PASS (0 blocking)

Rationale:
Full-tier battery (adherence + /review + /review-pr + simplify + gate). All five ticket-0335 exit criteria ADDRESSED with file:line evidence and independently re-executed tests — RED-pre-fix confirmed for the CRLF, size-cap, and set-e assertions; GREEN-pre-fix confirmed for the two regression guards, matching the PR body. `git log origin/main..HEAD --stat` shows exactly the four intended files, no scope overflow. No blocking adherence, no verifiable: minors, no human request-changes, no must-fix simplify findings. The three `consider:` minors are worth a follow-up ticket but do not gate merge.

telemetry: tier=full wall=~900s agents=4 tokens=na cost~=na

MinhHaDuong:
Exit criteria: 6/6  Review: 0 unresolved  Simplify: 0 unresolved
Minors: verifiable:0 consider:2 nofollow:2 malformed:0
Rationale: Independent standalone re-gate. All six ticket-0335 exit criteria ADDRESSED with file:line evidence. RED-pre-fix re-confirmed in-gate by running the new suite against origin/main's bash-env.sh: CRLF value strip (test 1), size-cap parse-skip + stderr warning (test 3), and set -e realpath reaches-end (test 4a) all fail on old code; the two regression guards (tests 2, 4b) are GREEN pre-fix, matching the PR body. make lint 17/17, erg check PASS (330 tickets). git log origin/main..HEAD --stat shows exactly the 4 intended files (bash-env.sh, test_bash_env_robustness.sh, memory note, ticket) — no scope overflow. The 2 consider: minors (double-realpath silent skip, per-line parse cost uncapped) and 2 nofollow: minors are informational by tag; none are verifiable: or human request-changes. Consistent with the prior /gaze APPROVED verdict.

PR #596, opened 2026-07-14
fix(0333): output-rewrite-tolerant git parses in erg-pr-merge + sync-local-main

MinhHaDuong:
## Summary

Makes the parsed-git call sites in the merge helper `erg-pr-merge` and in
`scripts/sync-local-main.sh` tolerant of output-rewriting/framing hooks, so the
guards no longer misfire when a hook mangles the stdout of a git call they parse.

The #584 incident (an `erg-pr-merge` run that spuriously aborted "close-claimed
ticket(s) 0329 are absent from tickets/ at branch tip" while the file was
present) is the motivating symptom. Its root cause is **unconfirmed** — rtk
0.34.3 rewrites only the outer Bash-tool command text, never a script's internal
subprocess git calls — so this is framed as general format-tolerance against
output-rewriting/framing, not an rtk-specific fix. The scripts carry no rtk
reference and work where rtk is absent.

## Per-site disposition

Probe evidence (`rtk hook check`, rtk 0.34.3): `git branch --show-current` and
`git worktree list --porcelain` ARE rewrite targets; `git symbolic-ref`,
`git ls-tree`, `git cat-file -e` are not.

1. **`erg-pr-merge:76` branch guard** — replaced the rewrite-target porcelain
   `git branch --show-current` with plumbing `git symbolic-ref --quiet --short
   HEAD || true`. Probe-verified not a rewrite target; `|| true` preserves the
   friendly "must run from PR branch" die on a detached HEAD (where
   `symbolic-ref --quiet` exits non-zero and would abort under `set -e`). Same
   idiom as `worktree-salvage.sh`.
2. **`erg-pr-merge:164` existence guard** — replaced the `git ls-tree | grep`
   presence check with an exit-code test per ticket: resolve the slugged
   filename via a shell glob (`compgen -G`, a filesystem read, not git stdout),
   then confirm it is committed at HEAD with `git cat-file -e HEAD:<path>`. The
   verdict is the exit code alone, immune to any stdout framing; the ID-prefix
   match of the old check is preserved. The genuinely-absent-ticket abort
   invariant is unchanged (case 22, unframed). Accepted limitation (documented
   at the site): `compgen` resolves the slug from the working tree, so a working
   tree drifted from HEAD could report a present ticket MISSING — harmless
   because merge always runs on a clean PR-branch tip (working tree == HEAD),
   and resolving from HEAD instead would reintroduce the git-stdout parse this
   block removes.
3. **`sync-local-main.sh:56` checkout-location parse** — filter the porcelain
   stream to well-formed record keys (drops injected banners) and require
   `NF == 2` on the `branch` record (rejects malformed extra-field summary
   lines). The `worktree` line keeps `substr` and no arity check (porcelain
   paths may contain spaces). Empirically, a bare `--- Changes ---` banner does
   not break the old awk; the `NF == 2` guard is the discriminator against a
   summarising rewrite.
4. **`worktree-salvage.sh:22`** (sibling sweep) — already reads the branch via
   the same plumbing idiom `git symbolic-ref --quiet --short HEAD`. Probe-verified
   not a rewrite target. **Cleared, no change.**
5. **`worktree-gc.sh:64` worktree-list parse** (sibling sweep) — the same
   `git worktree list --porcelain` idiom as site 3, and invoked via the Bash
   tool from `skills/molt` and `skills/roar`. Its `branch refs/heads/*` arm
   accepted a multi-token remainder as the branch name, so a framed record like
   `branch refs/heads/main is stale` set `branch="main is stale"`. The arm now
   rejects a whitespace-bearing remainder (fail closed: no branch set → the
   worktree is skipped, never wrongly removed); an injected banner line still
   matches no arm and is a no-op. Parser probe RED-before
   (`branch="main is stale"`) / GREEN-after (`branch=""`). **Fixed.** No
   `tests/test_worktree_gc.sh` suite exists; per scope discipline none was
   created — the hardening is defended by inspection and the recorded probe.
6. **`guard-commit-on-main.sh:52` `git branch --show-current`** (sibling sweep)
   — a confirmed rewrite target, but this script is a PreToolUse **hook**,
   spawned by the hook system and never via the Bash tool, so its internal git
   call is not subject to the outer Bash-tool rewrite. **Cleared, hook-spawned,
   out of scope, no change.**

## RED-before / GREEN-after

Regression cases use a transparent-by-default git shim (inert unless a
`SHIM_FRAME_*` flag is set) that prepends rtk-style framing to the targeted
subcommand only, leaving the existing 25 + 10 cases untouched.

| Case | Suite | Pre-fix (RED) |
|------|-------|---------------|
| 26 | `test_erg_pr_merge.sh` | framed `branch --show-current` → dies `must run from PR branch 'pr-framebranch'` |
| 27 | `test_erg_pr_merge.sh` | framed `ls-tree` → `close-claimed ticket(s) 0331 are absent from tickets/ at the branch tip` (the #584 symptom, reproduced) |
| 28 | `test_erg_pr_merge.sh` | source ratchet flags the surviving `git branch --show-current` |
| 11 | `test_sync_local_main.sh` | framed `worktree list` → `co_path` corrupted, main not fast-forwarded |

All four are GREEN after the fix. The `worktree-gc.sh` parser hardening is
covered by inspection + a standalone RED/GREEN probe (recorded in the reroll
commit); no dedicated suite exists to extend.

## Verification

- `make check` green (846 passed) — includes both bash suites via the
  `test_bash_suites.py` integration wrapper (run with plain `bash`, no rtk).
- `make lint` green (adherence tier, 17 passed); `erg check` PASS.
- A real end-to-end `erg-pr-merge` run against a live forge is not feasible
  pre-merge; the framing-shim cases carry the verification.

**Ticket:** tickets/0333-erg-pr-merge-guard-breaks-under-rtk-rewrite.erg


MinhHaDuong:
## Multi-perspective review — PR 596 (ticket 0333)

Ran correctness, consistency, scope, red-team, and doc-propagation passes in parallel, plus independent verification: both new bash-suite cases (26/27/28 in `test_erg_pr_merge.sh`, 11 in `test_sync_local_main.sh`) genuinely reproduce RED on the pre-fix commit (`5357bf6`) and GREEN on the fix — confirmed by checking out the pre-fix commit directly and re-running the suites, not just trusting the ticket log. `make check` also reproduces green (846 passed) on this branch.

### Blocker

- **Sibling-sweep exit criterion is checked off (`[x] Sibling helper sweep done`) but a live, same-idiom, same-context sibling was missed: `scripts/worktree-gc.sh:68`.** It parses `git worktree list --porcelain` in an unfiltered `while read` loop — the exact idiom this PR calls "a confirmed live rewrite target" when hardening `sync-local-main.sh:56` — and it's invoked directly via the Bash tool from `skills/molt/SKILL.md:73` and `skills/roar/SKILL.md:124`, the identical interactive/autonomous execution context as `erg-pr-merge`/`sync-local-main.sh`. The ticket's own Actions §3 says "Sweep sibling helpers ... at minimum" naming `worktree-salvage.sh` and `sync-local-main.sh`, implying a broader sweep was intended; the execute log only verified `worktree-salvage.sh` (cleared, already safe) and never mentions `worktree-gc.sh`. Two other candidate siblings we checked (`scripts/guard-commit-on-main.sh:52,54` — a directly-spawned PreToolUse hook, never a Bash-tool command itself; `scripts/check-cross-pr-ticket-collision.sh:78` — CI-only, `pull_request` events, no rtk-hooked shell ever present) are genuinely out of scope and correctly left untouched. `worktree-gc.sh` is not — please either fix it in this PR or explicitly clear it in the ticket log with the reasoning for deferral (and uncheck/annotate the exit criterion accordingly rather than leaving it silently overclaimed).

### Minors

- **verifiable:** the existence-guard rewrite (`compgen -G "tickets/${tid}-*.erg" | head -1` + `git cat-file -e HEAD:<path>`) trades the old check's "committed-tree-only, filesystem-independent" guarantee for one that depends on the working tree matching HEAD. Concrete failure: ticket 0333 is committed at HEAD as `tickets/0333-old-slug.erg`, but the working tree has an uncommitted `git mv` to `tickets/0333-new-slug.erg`. `compgen -G` finds only `new-slug.erg` (disk state); `git cat-file -e HEAD:tickets/0333-new-slug.erg` fails (that path isn't at HEAD); the ticket is falsely reported MISSING even though it genuinely exists at the branch tip under its old name. A second variant: if two files match the ID-prefix glob (a duplicate/colliding ticket ID — a scenario this harness's own ticket rules acknowledge can occur), `head -1` picks one by sort order rather than "whichever is at HEAD," and can falsely report MISSING if it picks the wrong one. Both fail closed (block the merge, recoverable via the existing `ERG_PR_MERGE_ALLOW_MISSING_TICKET` override) — not a wrong-merge risk, just a false-abort annoyance — but worth a one-line note or switching to `git ls-tree HEAD -- "tickets/${tid}-*.erg"` (plumbing, no stdout grep, immune to both output-rewriting and this regression).

- **consider:** memory `feedback_rtk_rewrites_git_output.md` still prescribes `rtk proxy git ...` as the standing workaround for machine-parsed git output, and ticket 0333's own Actions section proposed it as a fallback ("plumbing ... `rtk proxy git` otherwise"). The fix that actually landed uses only plumbing/defensive parsing and never `rtk proxy git` anywhere — consistent with the ticket's own empirical correction that rtk fires once on the outer Bash-tool command text and never touches a script's internal subprocess calls (so `rtk proxy git` inside a script wouldn't engage the hook it's meant to bypass). The memory entry doesn't reflect this and isn't touched by this PR; a follow-up correcting it (scope `rtk proxy git` to direct Bash-tool invocations, not internal script calls) would close the loop.

- **consider:** the three-part hardening idiom (plumbing `symbolic-ref` over `branch --show-current`; exit-code `cat-file -e` over `ls-tree | grep`; porcelain-key-filter + `NF` check over raw parsing) is documented only as inline comments in two scripts, echoing a third (`worktree-salvage.sh:22`) that had already independently reached for part of it before this ticket. Nothing generalizes it (e.g. in `rules/coding-bash.md`), so a fourth script's author has no rule file to consult and will likely reach for `git branch --show-current` again. The ticket's "No generic wrapper layer (YAGNI)" note is a defensible call against a code abstraction but doesn't preclude writing the pattern down in prose.

- **consider:** `tickets/0332-sigpipe-class-pipe-in-rules-injection-suite.erg` picks up two planning-log lines in this PR with zero reference anywhere in the PR body or commit trailers (`Ticket:`/`Ticket-ref:` lines name only 0333). This is harmless for merge safety — `erg-pr-merge` only closes tickets explicitly named in those lines, so 0332 cannot be accidentally closed — but it's a traceability gap: an auditor scanning `Ticket:`/`Ticket-ref:` lines has no way to discover 0332 was touched; only the raw diff stat shows it. 0332 itself remains genuinely open and unfixed (its target file `tests/test_harness_rules_injection.sh` has zero changes in this diff), which is fine — just flagging the untracked touch.

- **nofollow:** the ticket's diagnosis log calls `branch --show-current` and `worktree list --porcelain` "confirmed live rewrite targets" in the same breath as stating rtk "fires ONCE on the outer Bash-tool command text, never on a script's internal subprocess calls" — which would apply to these internal call sites too, and undercuts calling them "confirmed" exposures rather than defensive hardening. The code's own comments and commit message state the more careful framing ("not a claim that rtk corrupts these specific in-script calls"), so the shipped artifact is more honest than the ticket log — noted for rigor, not pursued further.

- **nofollow:** a `git` on PATH capable of injecting well-formed fake porcelain records (not just a banner) could in principle redirect `sync-local-main.sh`'s `co_path` resolution to an attacker-chosen directory. Exploiting this already requires PATH-level control over `git`, at which point arbitrary code execution is trivially available through other means — no incremental exposure, not pursued.

### Verification performed
- Confirmed RED-before (checked out pre-fix commit `5357bf6`, ran both suites — cases 26/27/28 and 11 fail as claimed) and GREEN-after (HEAD — all pass) directly, not just from the ticket log.
- `make check` — 846 passed.
- Confirmed the two out-of-scope sibling candidates (`guard-commit-on-main.sh`, `check-cross-pr-ticket-collision.sh`) are genuinely not exposed to the threat class this PR addresses.

MinhHaDuong:
## /gaze actions

round: 1
tier: full
adherence: PASS — 0 blocking
review: 1 comment posted (correctness sound; flagged incomplete sibling sweep)
review-pr: 1 comment posted (1 blocker: sibling-sweep exit criterion overclaimed)
simplify: 0 fixes applied

## /verify-gate verdict

verdict: REROLL

Exit criteria:
- erg-pr-merge's parsed-git call sites are hook-rewrite-proof: ADDRESSED — branch guard → `git symbolic-ref --quiet --short HEAD || true`; existence guard → `compgen -G` + `git cat-file -e HEAD:<path>` (exit-code verdict). Both correct.
- Regression test RED-before/GREEN-after: ADDRESSED — verified independently: checked out RED commit 5357bf6, reproduced FAIL on cases 26/27/28 (test_erg_pr_merge.sh) and case 11 (test_sync_local_main.sh); case 27 reproduces the exact #584 symptom ("close-claimed ticket(s) 0331 are absent from tickets/ at the branch tip"); all GREEN at HEAD.
- Sibling helper sweep done (fixed or explicitly cleared): MISSING — the sweep cleared `worktree-salvage.sh` and fixed `sync-local-main.sh`, but `scripts/worktree-gc.sh:68` parses `git worktree list --porcelain` with the identical unfiltered idiom this PR calls "a confirmed live rewrite target," is invoked via the Bash tool from `skills/molt/SKILL.md:73` and `skills/roar/SKILL.md:124`, and was neither fixed nor explicitly cleared. Action item 3 named the two files "at minimum." `scripts/guard-commit-on-main.sh:52,54` also uses `git branch --show-current` but is a PreToolUse hook (never Bash-tool-invoked, so out of scope for the rtk mechanism) — this warrants an explicit clearance line to resolve it in the PR record, not a fix.
- `make check` passes: ADDRESSED — Agent C re-ran `make check` (846 passed).

Unresolved review comments:
- BLOCKER: sibling-sweep exit criterion overclaimed — `worktree-gc.sh:68` unswept (fix its porcelain parse or clear it explicitly with a fail-closed rationale).
- verifiable: existence-guard `compgen -G` reads the working tree to resolve the filename before the HEAD check; a working tree drifted from HEAD (dirty rename/delete) yields a false MISSING. Fails closed, override-recoverable — low real-world risk at merge time (branch guard ensures a clean PR-branch tip) but worth a hardening or an explicit accepted-limitation note.

Unresolved simplify: none (skipped — nothing to simplify)
Adherence: PASS (0 blocking; 2 nits on the compgen resolution, non-blocking)

Rationale:
The core deliverable — hardening the two #584 call sites in `erg-pr-merge` and `sync-local-main.sh` — is correct, well-commented, and the RED-before/GREEN-after evidence reproduces exactly (including the #584 symptom). The block is narrow and bounded: the "Sibling helper sweep done" exit criterion is checked, but a real sibling (`worktree-gc.sh`) using the identical `git worktree list --porcelain` idiom, invoked via the Bash tool from two skills, was neither fixed nor explicitly cleared. Two independent reviewers caught this. Discharging it is cheap: harden `worktree-gc.sh`'s parse (reject a malformed multi-token `branch` record; the banner line is already ignored by its `case` statement) or clear it explicitly with its fail-closed rationale, and add an explicit clearance line for the hook-spawned `guard-commit-on-main.sh`. Approving over a live blocker on a checked-but-unmet criterion is exactly the rubber-stamp this gate exists to prevent. Round-1 fix agent dispatched.

telemetry: tier=full wall=1520s agents=3 tokens=na cost~=na

MinhHaDuong:
## /gaze actions

round: 2
tier: full
adherence: PASS — 0 blocking
review: resolved (incomplete-sweep finding addressed)
review-pr: resolved (blocker addressed)
simplify: 0 fixes applied
fix agent: 1 commit (44bbf28)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- erg-pr-merge's parsed-git call sites are hook-rewrite-proof: ADDRESSED — branch guard → `git symbolic-ref --quiet --short HEAD || true`; existence guard → `compgen -G` + `git cat-file -e HEAD:<path>` (exit-code verdict). Correct; compgen working-tree assumption now documented as a deliberate trade-off (resolving from HEAD would reintroduce the ls-tree parse exposure the block removes).
- Regression test RED-before/GREEN-after: ADDRESSED — independently reproduced RED at commit 5357bf6 (cases 26/27/28 in test_erg_pr_merge.sh, case 11 in test_sync_local_main.sh); case 27 reproduces the exact #584 symptom; GREEN at HEAD.
- Sibling helper sweep done (fixed or explicitly cleared): ADDRESSED (round 2) — full disposition now in the ticket body and PR body. `worktree-gc.sh` FIXED: its `branch refs/heads/*` arm rejects a multi-token remainder (fail-closed), verified by parser probe (malformed→empty, valid→"main", banner→ignored). `guard-commit-on-main.sh` CLEARED with rationale (PreToolUse hook, never Bash-tool-invoked). `worktree-salvage.sh` cleared (already uses the symbolic-ref idiom); `sync-local-main.sh` and `erg-pr-merge` fixed.
- `make check` passes: ADDRESSED — re-ran on the final tip 44bbf28: 846 passed (1 pre-existing benign async-mock warning in test_beat.py, unrelated).

Unresolved simplify: none
Adherence: PASS (0 blocking)

Scope note (report-only, not blocking): `tickets/0332-*.erg` carries two raid-332/333 imagine/plan log-note lines; 0332 is not close-claimed by this PR so it cannot be accidentally closed. Traceability nit only.

Rationale:
Round 1 REROLLed on one narrow, checkable defect: the "Sibling helper sweep done" exit criterion was checked while `worktree-gc.sh` (same `git worktree list --porcelain` idiom, invoked via the Bash tool from molt/roar) was neither fixed nor cleared. The round-1 fix hardened `worktree-gc.sh`'s parse (fail-closed against a framed multi-token branch record), explicitly cleared the hook-spawned `guard-commit-on-main.sh`, and recorded a complete per-site disposition in both the ticket and the PR body. The core deliverable — the two #584 call sites in `erg-pr-merge` and `sync-local-main.sh` — was correct from round 1 with reproducible RED-before/GREEN-after evidence. All exit criteria are now met with concrete evidence; the full suite is green on the final tip. verify: approved.

telemetry: tier=full wall=2820s agents=4 tokens=na cost~=na

PR #603, opened 2026-07-14
seat(0207): scrub injected credential from reviewer output (criterion-2 mitigation)

MinhHaDuong:
## Summary

Delivers **criterion 2 mitigation** of ticket 0207 — the red-team fix that must land *before* a real OpenRouter key is trusted inside the seat sandbox (PR #575 gaze dissent). Scope is the mitigation only; the advisory trial (criteria 3–4) stays open.

**Mitigation (a) — credential scrub (the real work).** A reviewer seat given a real key reads attacker-controllable diff text, and its output flows back out; a prompt-injected diff could coax the key into the findings or stderr channel. Fix: after the `run_seat` capture, redact the *exact runtime `OPENAI_API_KEY` value* in `raw.out` **and** `raw.err` in place, **before** any path that echoes them — the failure tail, the contract-grep that writes `$OUT`, and the WARN `cat`. The value is read from the process env via a `python3` heredoc (never argv; only file paths are argv), provider-agnostic (substitutes the value, not a vendor key pattern), and a no-op for the local dummy key. The capture is restructured to record rc explicitly (`|| review_rc=$?`) so the scrub sits between capture and every echo path.

**Mitigation (b) — egress pinning: already satisfied.** The seat runs under `--network=none` and reaches exactly the `ENDPOINT` host:port through the single-destination relay; no re-implementation. A tier-1 source test now pins that invariant (exactly one host-side `--connect` target = the ENDPOINT pair; no wildcard/second destination), so a regression goes red.

## Tests
- **RED-first exfiltration-sim** in `tests/test_seat_runner.sh`: a stub podman models a compromised seat that leaks the injected key into the findings line **and** to stderr; asserts the dummy secret is absent and `[REDACTED-CREDENTIAL]` present on both the success (findings-file) and failure (stderr-tail) paths. Proven RED on pre-fix code (4 fails), GREEN after.
- Egress single-target source assertion (added, passes on current code).
- `tests/test_seat_runner.sh` **49/49**, bash-suites `seat_runner` **1/1**, `make check-fast` **PASS**, `verify-adherence` **PASS**.

## OpenRouter real smoke (criterion 2) — transport proven, auth blocked
Ran `scripts/seat-runner.sh` against `https://openrouter.ai/api/v1` (`openai/gpt-4o-mini`, `--health-path /api/v1/models`, `--credential-env`). Transport was proven **end-to-end** (~8 s): health probe reached OpenRouter, the TLS relay carried bytes through the `--network=none` container, aider connected and issued the completion, and a structured provider response returned through the relay. However the provisioned key authenticates as **"User not found"** (deactivated account) — an **environmental/provisioning failure, not a code failure**. Per raid discipline this is recorded honestly and **not faked as a pass**: exit-criterion 2 is left **unticked** (no judgeable findings without a valid key). `LEAK-CHECK: PASS` (the scrub had nothing to leak). A valid key will complete the finding-quality judgment.

## Ticket


MinhHaDuong (commented):
## Multi-perspective panel review — PR #603

Five reviewers (Correctness, Red-team, Consistency, Scope, Doc-propagation), each `sonnet`, fresh context. **No blockers found.** Consolidated verdict: **comment** (approve-with-notes). Tests re-run locally on this branch: `tests/test_seat_runner.sh` **49/49**, `make check-fast` exit 0.

The scrub does exactly what the PR claims: it redacts the exact runtime `OPENAI_API_KEY` value from `raw.out` and `raw.err`, unconditionally, before all three echo paths (failure-tail `:351`, contract-grep→`$OUT` `:356`, WARN `cat` `:358`), on every exit path including timeout/SIGKILL. Credential fidelity (`export OPENAI_API_KEY="${!CREDENTIAL_ENV}"` at `:144` → same value scrubbed) and the `review_rc=0 … || review_rc=$?` `set -e`-safe refactor were both verified sound. Scope is clean — every hunk serves mitigation (a)+(b), STATE.md is a proportionate stale-bullet correction, and `Ticket-ref:` (non-closing) is the right call since criterion 2 stays unticked on the deactivated key.

### Findings

**[Correctness] `verifiable:` (medium) — WARN-`cat` leak path is functionally protected but untested.**
The scrub runs unconditionally on both files before all three reads, so the WARN path (`grep … || { cat raw.out >&2; }`, `seat-runner.sh:356-358`) is covered *by construction* — the correctness reviewer reproduced RED-on-base / GREEN-on-head for this exact scenario (stub podman, exit 0, non-contract-shaped stdout leaking the key). But the two added tests only exercise the findings-file and failure-tail paths; no test drives the WARN-`cat` path the ticket's own feasibility note called out as "a THIRD leak path." Recommend a third `assert_absent`/`assert_contains` pair so the "all three paths" claim is test-backed.

**[Red-team] `verifiable:` (low, latent) — `CRED_ARGS` is injected into the containment self-test invocation that has no need for it.**
`CRED_ARGS` is baked unconditionally into `run_seat` (`:265`), so the pre-review self-test container (`:276`) also receives the real key. Not exploitable today — the self-test body is a fixed literal string and `$PROBE` is only grep-checked for markers, never echoed — but `$PROBE` is never passed through the scrub, so a future diagnostic `echo "$PROBE"` or a credential-adjacent probe would leak silently and untested. Worth a small follow-up ticket to scope `CRED_ARGS` out of the self-test call.

**[Correctness + Red-team] `consider:` — literal `str.replace` cannot catch a transformed/encoded leak; document the limit.**
`data.replace(secret, …)` (`:342`) is exact-substring only; base64/hex/reversed/char-split re-emissions defeat it. Both reviewers agree this is defense-in-depth, not a containment boundary, and the PR is honest in calling it a "mitigation." The red-team reviewer adds the load-bearing nuance, preserved verbatim: *"the model is never given `OPENAI_API_KEY` as context, and this invocation wires no tool/shell-execution path back to the model's generated text… the realistic literal-leak vector is non-adversarial: aider/litellm's own auth-error or debug-trace output printing the key verbatim — which the scrub does catch."* Recommendation: add a comment stating the scrub is literal-value-only and its safety rests on the reviewing LLM having no channel to read+transform the true secret, so a future maintainer enabling aider tool/shell use knows they would invalidate the argument.

**[Red-team] `consider:` — confirm the one unverified assumption.** The whole "model can't read the secret" argument rests on `aider --chat-mode ask … --yes-always` never executing model-suggested shell commands. That flag's scope was not verified in this review; worth a one-line confirmation before a real key is trusted.

**[Doc-propagation] `consider:` — the script header does not mention the scrub.** The top-of-file "Containment:" block documents every other security property exhaustively but is silent on the new credential-output scrub; the `--credential-env` usage text (`~:61`) could cross-reference that the value is scrubbed from output, not just kept off argv. Non-blocking.

**[Consistency] `consider:` — commit type `seat(0207):` is a one-off.** The repo's conventional-commit vocabulary is a closed set (`fix(` ×15, `ticket(` ×20, `feat(` ×4, …); `seat(` appears only on this commit. A credential-leak fix fits the established `fix(0207):`. Cosmetic; the commit is already made.

**`nofollow:`** — short-secret substring collision (real keys are high-entropy); `review_rc` lowercase top-level var (common repo-wide); inline `python3` heredoc has no precedent but `python3` is already a hard dep (`net-relay.py`) and the unit is small; STATE.md bold-span colon placement; a stale unscrubbed copy in `scripts/seat-runner-prototype.sh` lives in a *different* worktree/branch, out of scope here.

### Disposition
Nothing blocks merge at the PR's stated scope (criterion 2, mitigation a+b). Before calling ticket 0207 fully closed, two items merit follow-up tickets: (1) the WARN-path test, (2) scoping `CRED_ARGS` out of the self-test. The doc/limit comments are optional. Verify the `--chat-mode ask` no-shell-exec assumption before trusting a live key.

🤖 review-pr panel (correctness · red-team · consistency · scope · doc-propagation)

MinhHaDuong:
## Review response (panel was approve-with-notes, no blockers)

Addressed in the follow-up commit:

- **Correctness `verifiable:` — WARN-cat path untested despite the "all three paths" claim.** Added a third exfiltration-sim case: a seat that exits zero but emits no contract-shaped line makes the contract-grep fail, driving the `cat raw.out >&2` WARN path; the test asserts the injected key is scrubbed there too. `tests/test_seat_runner.sh` is now **51/51** (all three echo paths — findings-file, failure-tail, WARN-cat — have coverage).
- **Red-team `verifiable:` — `CRED_ARGS` injected into the containment self-test (`$PROBE` never scrubbed).** Filed as follow-up **ticket 0337** (self-test needlessly receives the real credential; a non-issue today under `--network=none` with line-by-line probe assertions, but the key should not be in that container's env). Left as a separate ticket rather than widening this PR's scope.
- **`consider:` — document the literal-value limit.** Script header now states the scrub is a literal runtime-value redaction (defense-in-depth against the client verbatim-echoing the key, e.g. an auth-error line), not a semantic filter.

The `seat(` vs `fix(` commit-type nit is noted; not rewriting merged history over it.

MinhHaDuong:
## /gaze round-1 panel — PR #603 (post-fix commit 6edbc69)

Five reviewers (Correctness, Consistency, Scope, Red-team, Doc-propagation), each `sonnet`, fresh context, scoped to: (a) did follow-up commit `6edbc69` correctly and fully address the prior panel's verifiable findings from commit `0dd91c6`, and (b) any NEW defects in the full diff. Prior panel's `consider:`/`nofollow:` points were not re-litigated. **No blockers found.** Consolidated verdict: **comment** (approve-with-notes). Tests: `make check-fast` **765 passed, 82 deselected**; `tests/test_seat_runner.sh` independently run **51/51**.

### Prior findings — disposition confirmed

- **[Correctness] `verifiable:` WARN-cat leak path untested → FIXED, fully.** Isolating the `0dd91c6...6edbc69` delta: 5 comment lines (no production-code change) plus a 25-line test case (`EXFILBIN3` stub) in `tests/test_seat_runner.sh`. The scrub logic itself was already shipped at `0dd91c6`; this commit adds coverage only. Traced the new test's mechanics: the stub returns no `FINDING|`/`SUMMARY|` line on the review call, driving `grep` failure → the WARN `cat "$WORK/raw.out" >&2; exit 1` path (`seat-runner.sh:358-360`); stderr capture order (`2>&1 >/dev/null`) is correct; asserts the injected secret absent and `[REDACTED-CREDENTIAL]` present in that exact stream. Ran directly — all three exfiltration-sim cases (findings-file, failure-tail, WARN-cat) pass. All three claimed echo paths now have coverage.
- **[Red-team] `verifiable:` `CRED_ARGS` unnecessarily injected into the containment self-test → deferred to new ticket 0337, not fixed here.** Confirmed a legitimate scope call, not a blocker: `run_seat`'s `--network=none` (unconditional) covers the self-test invocation too, and `$PROBE` is only ever `grep -q`-checked line-by-line — no wholesale-echo path exists today, so the real key sitting in that container's env cannot exfiltrate. Ticket `tickets/0337-seat-self-test-cred-args-scoping.erg` is spec-compliant (`erg check tickets/` → PASS, 332 tickets) and well-formed against the handoff-document template (Context/Relevant files/Actions/Test/Invariants/Exit criteria all present, both exit criteria mechanically verifiable). Per the ticket-decomposition doctrine, a sub-task with its own test/exit-criteria surface that doesn't gate the parent PR's stated scope (criterion-2 mitigation a+b) belongs in a child ticket — `Ticket-ref:` (non-closing) is the right call.
- **[Correctness+Red-team] `consider:` literal-value scrub can't catch transformed/encoded leaks → documented, not code-fixed (as intended).** New header bullet ("CREDENTIAL SCRUB (0207)...literal value substitution...NOT a semantic filter") is accurately scoped and correctly placed (adjacent to the existing credential-injection bullet).
- **[Red-team] `consider:` `--chat-mode ask`/`--yes-always` shell-exec-prevention assumption never independently verified → now independently VERIFIED, high confidence.** Traced the installed aider (0.86.2) source directly: `AskCoder` never overrides `get_edits()`/`apply_edits()` (base class no-ops), and never populates `self.shell_commands` (only `editblock_coder.py` does) — shell execution and file writes are structurally unreachable in ask mode, not merely policy-discouraged. Bonus: `--yes-always` with `explicit_yes_required=True` actually *forces a "no"* on any shell-exec confirmation, so even a hypothetical future wiring would be refused. This is defense-in-depth on top of, not instead of, the read-only-rootfs/`--network=none` containment.
- **[Consistency] `consider:` `seat(0207):` commit-type is a one-off → confirmed legitimately left as won't-fix**, consistent with this repo's git discipline (avoid rewriting merged history for a cosmetic nit); the type is now used twice within this ticket's commits, so it's an outlier only against the repo-wide vocabulary, not internally inconsistent.
- **[Doc-propagation] `consider:` script header silent on the scrub → header block fixed; usage-text half NOT done.** See new finding below.

### New findings (commit 6edbc69 / full-diff delta)

**`consider:` (red-team) — the ask-mode/no-shell-exec structural guarantee is unpinned and untested against aider upgrades.** `aider` is installed as an unpinned uv-tool; nothing in this repo pins its version or gates the `AskCoder`-stays-inert behavior in CI. The new header doc now leans on this guarantee more explicitly than before. Worth a follow-up ticket: a self-test asserting `get_edits`/`shell_commands` stay inert for `edit_format=ask`, or a version pin.

**`consider:` (doc-propagation) — the `--credential-env` usage text (`scripts/seat-runner.sh:68`) still doesn't cross-reference the scrub.** Only the Containment header block got the addition the prior finding suggested; the flag's own usage line is unchanged. Low stakes — the Containment block sits immediately above and does cover it — but a reader skimming just the usage line won't learn the value is also scrubbed from output.

**`consider:` (doc-propagation) — ticket 0337 is not cross-referenced from ticket 0207's body/log**, only the reverse. Soft: 0337 is a sibling follow-up spawned by red-team dissent, not a strict child under the tracking-ticket convention, so this is a nicety rather than a gap.

**`nofollow:`** — the new `EXFILBIN3` test case is structurally consistent with its two predecessors (same stub/assert shape, fresh `RANDOM`-seeded secret per run, correct ordering verified: scrub runs before the WARN-cat `grep`/`cat`); no issue found. `nofollow:` — no other doc surface (`skills/gaze/SKILL.md`, `skills/reviewers/SKILL.md`, `skills/raid/SKILL.md`) needs the scrub detail; none touched, none stale. `nofollow:` — STATE.md's updated bullet is accurate against the ticket log and exit-criteria checkboxes (criterion 2 mechanism delivered, exit criterion correctly left unticked pending a valid key).

### Disposition

Nothing blocks merge. Commit `6edbc69` correctly and fully closes the one `verifiable:` gap it targeted (WARN-cat coverage), and its choice to defer the other `verifiable:` finding (CRED_ARGS self-test scoping) to ticket 0337 rather than widen this PR is sound — independently re-confirmed non-exploitable today. The `--chat-mode ask` shell-exec-prevention assumption, previously an open `consider:`, is now independently verified true at the source level. Two small `consider:` doc gaps and one new `consider:` (aider-version pin/guard) are optional follow-ups, not blockers.

🤖 review-pr panel round 2 (correctness · consistency · scope · red-team · doc-propagation), post-fix verification of commit 6edbc69

MinhHaDuong:
## /gaze actions

round: 2
tier: full (219 lines, 5 files)
adherence: PASS — 0 blocking (test_seat_runner.sh 51/51; pytest -m adherence 17/17; bash-rules review clean, `review_rc=0 … || review_rc=$?` idiom confirmed set-e-safe)
review: 0 blockers, 2 nits (literal-substitution limit — disclosed by design; CRED_ARGS self-test gap — tracked by follow-up ticket)
review-pr: 1 panel comment posted (post-fix commit, 5 perspectives, verdict: comment, 0 blockers)
simplify: 0 fixes applied — 3 findings triaged as skips: test-invocation helper (matches file's explicit-stub idiom), inner `if secret:` guard (kept as defense-in-depth), altitude (scrub keyed to call-site files rather than `run_seat` — reinforces the follow-up ticket's scope; consider folding into it)
fix agent: 1 commit (d497f21) — renumber follow-up ticket 0337→0339, cross-PR ID collision with #599; plus reroll-bump log line on 0207

## /verify-gate verdict

verdict: APPROVED

Exit criteria (judged against the PR's stated scope — criterion-2 mitigation, `Ticket-ref:` non-closing):
- 1 (seat via sandbox-runner, BASH_ENV credentials): ADDRESSED — pre-existing, unaffected by this diff
- 2 (both endpoint configs smoked): ADDRESSED for PR scope — scrub at scripts/seat-runner.sh:317-354 (commits 2f8678f + 6edbc69) covers all three echo paths, proven by three exfil-sim tests (51/51 re-run in the review worktree); OpenRouter real smoke honestly logged as blocked on a deactivated key, checkbox correctly left unticked
- 3 (advisory trial): OUT-OF-SCOPE — explicitly deferred, ticket stays open
- 4 (promote/drop recommendation): OUT-OF-SCOPE — tied to criterion 3

Unresolved review comments: none — prior panel's two `verifiable:` findings verified fixed (WARN-cat exfil test EXFILBIN3) or deferred with a well-formed follow-up ticket (now 0339); round-1 gaze panel raised 0 blockers, 3 `consider:` + 2 `nofollow:` minors, all correctly non-blocking
Unresolved simplify: none (0 must-fix)
Adherence: PASS

Round-1 REROLL blocker, resolved: CI `cross-pr-ticket-collision` failed — ticket ID 0337 also added by earlier open PR #599; later claimant renumbers per tickets/AGENTS.md. Fix d497f21 is a pure rename to 0339 (0 content changes; 0338 taken by #602, no open PR claims 0339) + one bump log line; scripts and tests byte-unchanged since the reviewed head. All 9 CI checks now pass.

Note for the author: the review-response comment above and prior panel text refer to the follow-up as "ticket 0337" — it is now `tickets/0339-seat-self-test-cred-args-scoping.erg`. The `consider:` minors worth follow-up attention: pin/verify the aider `--chat-mode ask` no-shell-exec assumption before trusting a live key, and the altitude suggestion to move the scrub into `run_seat` itself (would close 0339 by construction).

Rationale:
The delivered mitigation is sound and fully evidenced: the credential scrub sits at the capture chokepoint before all three echo paths, with RED-first tests for each; egress pinning is verified already satisfied and pinned by a source test; the OpenRouter smoke result is recorded honestly rather than faked. The only blocker was the mechanical ticket-ID collision, fixed in one clean commit. The merge decision belongs to the caller.

telemetry: tier=full wall~=1200s agents=15 tokens=791k (subagents, warn threshold 500k exceeded — full battery on a security PR plus one reroll round) cost~=na

PR #617, opened 2026-07-14
feat(0344): erg-pr-merge -C PATH flag for bare rule-matched invocation

MinhHaDuong:
## What

Adds a `-C PATH` flag to `skills/merge/erg-pr-merge`. The flag `cd`s into `PATH` before any git/gh/erg call, so the **bare** invocation
`~/.claude/skills/merge/erg-pr-merge -C WORKTREE N` targets an arbitrary checkout from any cwd.

## Why

The standing allow rule `Bash(~/.claude/skills/merge/erg-pr-merge:*)` prefix-matches only the bare invocation. A `cd X && …` prefix no longer starts with the script path, so it falls through to the stochastic auto-mode classifier, which denies self-merges unpredictably — the 2026-07-14 raid wave (#601-#603) stalled on exactly this. `-C` keeps the invocation bare while pointing at the target checkout.

## Corrected skip-trigger finding

The prior write-up blamed "no erg available" for the skipped ticket-closes on #601/#602. The real trigger is the ticket-close guard `[[ -d tickets ]]` (script line ~216): with a husk launch cwd outside any checkout, `tickets/` is absent so the whole close block is skipped — `ERG=${ERG:-erg}` resolution is not the cause. The `-C` cd lands *before* that guard (and before the `git symbolic-ref` branch check), so it fixes the close transitively. ERG resolution is left untouched.

## Safety

- The parse block is the very first thing after `die()`, before `in_worktree()` and Step 0 — no git/gh/erg call can precede the cd, so there is no half-cd state.
- The whole block is gated on `$1 == "-C"`; without the flag it is a no-op. Cases 1-28 of the regression suite are byte-behavior-identical (the backward-compat proof).
- `-C` with a missing or non-directory path fails loudly before any other work.

**HIGH-STAKES NOTE:** this script is the raid pipeline's own merge tool. The diff is deliberately additive and gated on `$1 == "-C"`.

## Test (RED → GREEN)

New Case 29 in `tests/test_erg_pr_merge.sh`, invoked from `$WORK` (deliberately not a git repo):
- happy path `-C <repo> 42` closes the seeded ticket — **RED** on today's script (dies "must run from PR branch"), **GREEN** after.
- `-C` with no path → non-zero + "requires a PATH" — RED→GREEN.
- `-C /nonexistent 42` → non-zero + "is not a directory" — RED→GREEN.

`make check` passes (27 shell suites + full pytest, exit 0). `/verify-adherence` clean — `/gaze` may skip the mechanical phase.

**Ticket:** tickets/0344-erg-pr-merge-chdir-flag-for-bare-invocation.erg

### DIFF
--- a/skills/merge/SKILL.md
+++ b/skills/merge/SKILL.md
@@ -2,7 +2,7 @@
 name: merge
 description: Atomically close the linked ticket(s) and merge a PR. Must be run from the PR head branch. Works in git worktrees and on VMs. GitHub-only (requires the GitHub CLI).
 user-invocable: true
-argument-hint: [pr-number]
+argument-hint: [-C path] [pr-number]
 ---
 
 # Merge $ARGUMENTS
@@ -12,11 +12,21 @@ Run:
 ~/.claude/skills/merge/erg-pr-merge $ARGUMENTS
 ```
 
-**Cross-repo prerequisite**: the caller must ensure cwd is the target
-project and the PR branch is checked out before invoking `/merge`. For
-cross-repo flows, this means `cd <project-path> && git fetch origin`
-and checking out the PR branch before the call. The script itself is
-cwd-based — it never takes a repo or path argument.
+**Cross-repo prerequisite**: the script operates on the checkout it runs in,
+with the PR branch checked out. Point it at a checkout in one of two ways:
+
+- **`-C PATH` (preferred for agents)** — the script cds into `PATH` before any
+  git/gh/erg call. The canonical agent invocation is the bare form
+  `~/.claude/skills/merge/erg-pr-merge -C WORKTREE N`, which prefix-matches the
+  standing allow rule `Bash(~/.claude/skills/merge/erg-pr-merge:*)` from any cwd.
+  A `cd WORKTREE && …` prefix does **not** match that rule (the command text no
+  longer starts with the script path), so it falls through to the stochastic
+  auto-mode classifier — use `-C` instead of a `cd` prefix.
+- **Implicit cwd** — with no `-C`, the script uses the current directory, so the
+  caller must `cd <project-path> && git fetch origin` and check out the PR branch
+  before the call.
+
+`-C` with a missing or non-directory path fails loudly before any other work.
 
 ## Ticket lines in the PR body
 
--- a/skills/merge/erg-pr-merge
+++ b/skills/merge/erg-pr-merge
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # erg-pr-merge — close ticket and merge a PR atomically
 #
-# Usage: erg-pr-merge [PR_NUMBER]
+# Usage: erg-pr-merge [-C PATH] [PR_NUMBER]
 #
 # Must be run from the PR's head branch. Works in git worktrees and on VMs.
 # GitHub-only: requires the GitHub CLI authenticated to the repo's GitHub remote.
@@ -24,6 +24,23 @@
 set -euo pipefail
 
 die() { echo "erg-pr-merge: $*" >&2; exit 1; }
+
+# ── Step -1: optional -C PATH (must land before any git/gh/erg call) ──
+# The standing allow rule `Bash(~/.claude/skills/merge/erg-pr-merge:*)` matches
+# only the BARE invocation; a `cd X && …` prefix falls through to the stochastic
+# auto-mode classifier. -C lets the bare form target an arbitrary checkout from
+# any cwd. The cd MUST precede every git/gh/erg call so no half-cd state exists —
+# in particular the ticket-close step's `[[ -d tickets ]]` guard and the
+# `git symbolic-ref` branch check must read the target checkout, not the launch
+# cwd (a husk cwd otherwise skipped the close, ticket 0344). Plain `cd`
+# semantics for a relative path, matching `git -C`; no realpath. (ticket 0344)
+if [[ "${1:-}" == "-C" ]]; then
+    [[ $# -ge 2 ]] || die "-C requires a PATH argument"
+    TARGET_DIR="$2"
+    [[ -d "$TARGET_DIR" ]] || die "-C path '${TARGET_DIR}' is not a directory"
+    cd "$TARGET_DIR" || die "-C: could not cd into '${TARGET_DIR}'"
+    shift 2
+fi
 
 # Identity predicate (tickets 0301/0296; incidents I1/I2). Do NOT regress to
 # `[ -f .git ]`: that matched ANY dir with a `.git` FILE — a submodule, or a
--- a/skills/merge/erg-pr-merge
+++ b/skills/merge/erg-pr-merge
@@ -6,9 +6,12 @@
 # Must be run from the PR's head branch. Works in git worktrees and on VMs.
 # GitHub-only: requires the GitHub CLI authenticated to the repo's GitHub remote.
 #
-# For cross-repo merges, the caller must cd into the target project and
-# checkout the PR branch before invoking this script. See raid Phase 7
-# and nightbeat-supervisor Phase 2 for the caller contracts.
+# For cross-repo merges, point the script at the target checkout (with the PR
+# branch checked out) either with `-C PATH` (preferred — keeps the invocation
+# bare so it matches the standing allow rule from any cwd) or by cd-ing there
+# first. Prefer `-C PATH` over a `cd PATH && …` prefix: the prefix defeats the
+# allow rule (see SKILL.md). See raid Phase 7 and nightbeat-supervisor Phase 2
+# for the caller contracts.
 #
 # Happy path:
 #   0. Verify: on PR head branch, CI green, no conflicts
--- a/tests/test_erg_pr_merge.sh
+++ b/tests/test_erg_pr_merge.sh
@@ -205,8 +205,10 @@ ERG
     git -C "$REPO" switch -q -c "$BRANCH"
 }
 
-run_merge() {  # $1 body, $2 title  — runs the script with cwd in the repo
-    ( cd "$REPO"
+run_merge() {  # $1 body, $2 title, $3+ script args (default: 42)
+    # cwd is $REPO unless RUN_MERGE_CWD overrides it (Case 29 runs from $WORK).
+    ( cd "${RUN_MERGE_CWD:-$REPO}"
+      (( $# >= 3 )) || set -- "$1" "$2" 42
       PATH="$STUBDIR:$PATH" \
       ERG="$ERG_LOCAL" \
       STUB_PR="42" STUB_BRANCH="$BRANCH" STUB_BASE="$BASE" \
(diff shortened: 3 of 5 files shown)


MinhHaDuong (commented):
## Multi-perspective review — PR #617 (`erg-pr-merge -C PATH`)

Five parallel reviewers (correctness, consistency, scope, red team, doc-propagation). Verdict spread: **2 approve, 2 comment, 1 request-changes.** Gate: `make check` = **856 passed, exit 0**; the shell suite is 34/34 incl. new Case 29. Case 29 was confirmed **genuinely RED on the pre-fix script and GREEN after** (correctness reviewer overlaid the test on merge-base `eb7be4e` and reproduced the failure).

The feature is correct and in-scope: the `-C` block lands after `die()` and before `in_worktree()` and Step 0, so the `cd` precedes every git/gh/erg call (no half-cd state); arg parsing is safe under `set -u`; the ticket is correctly left open (no `Closed:` header / `closed` verb) for merge-time close. Exit criteria are met.

Two things to weigh before merge — one design dissent, one high-consensus doc nit.

---

### Preserved dissent — Red Team recommends request-changes (confidence: high, live repro)

**`-C <non-worktree-path> N` reaches the destructive base-checkout path and is invisible to `guard-cd-primary-repo.sh`.** If `-C` points at a non-worktree checkout (e.g. the primary repo), `in_worktree()` returns false → `IN_WORKTREE=0` → `MERGE_FLAGS` gains `--delete-branch`, and Step 4 runs `git checkout "$BASE_BRANCH"` + `fetch` + ff-merge on that target (script lines 410-414, 457+). `scripts/guard-cd-primary-repo.sh` — written for exactly the "mutate the primary checkout from a worktree session" class — never fires, because its fast path is a textual `grep -qwP 'cd'` of the Bash command string and the `cd` here happens *inside* the script. Combined with the standing allow rule `Bash(~/.claude/skills/merge/erg-pr-merge:*)`, this is reachable unprompted. Live repro attached by the reviewer (guard exits 0 on `erg-pr-merge -C <primary> 42`).

Synthesizer triage (for the author's decision, not to overrule): the canonical documented usage `-C WORKTREE N` is **safe** — a worktree target keeps `IN_WORKTREE=1` and skips the destructive path entirely. The gap is real but narrows to *misuse* pointing `-C` at a non-worktree. And `guard-cd-primary-repo.sh`'s own header declares it "a narrow advisory guard against one reflex, **not a security boundary**" (fail-open when jq is absent) — so this is a widening of an advisory-guarded reflex surface, not a breach of a security boundary. That is why I did not unilaterally set the review to request-changes. **Author decides**: accept the tradeoff and merge, or harden. Two proposed fixes, either small: (a) reject a `-C` target that fails `in_worktree()`, matching the trust boundary the destructive-path skip already assumes; or (b) teach `guard-cd-primary-repo.sh` to apply its primary-root check to the `-C <path>` argument. I'd lean (a) — it closes the gap at the source and the ticket's real need (husk-cwd → worktree) is satisfied by worktree targets anyway.

---

### High-consensus finding — stale header comment (flagged independently by consistency, scope, doc-propagation)

`verifiable:` `skills/merge/erg-pr-merge:9-11` (unchanged by this PR; confirmed identical in `origin/main`) still reads:
```
# For cross-repo merges, the caller must cd into the target project and
# checkout the PR branch before invoking this script. See raid Phase 7
# and nightbeat-supervisor Phase 2 for the caller contracts.
```
Line 4's `# Usage:` synopsis was updated to show `[-C PATH]`, but this explanatory paragraph 5 lines down still teaches *only* the `cd`-first contract — the exact anti-pattern this PR's own SKILL.md now warns defeats the allow rule. A reader of just the script header gets stale advice. One-line-block fix; SKILL.md (the primary doc surface) is already correct, so not a blocker on its own.

---

### Doc-propagation follow-ups (not this PR's job — recommend a fast-follow ticket)

- `consider:` `skills/raid/SKILL.md` Phase 7 and `skills/nightbeat-supervisor/SKILL.md` §2 — the harness's two heaviest merge callers — still use the `EnterWorktree`/`cd`-then-bare-`/merge` dance that ticket 0344's own Context cites as the cost (`raid wave #601-#603`). They aren't broken (raid uses `EnterWorktree`, not a bash `cd`, so no classifier-fallthrough), but the practical payoff of `-C` isn't wired into them yet.
- `nofollow:` `projects/…AEDIST…/memory/feedback_bg_merge_anchoring.md` still prescribes the `cd <wt> && … erg-pr-merge N` compound (which does *not* match the allow rule). The harness's own `feedback_allow_rule_bare_invocation.md` already cross-references `-C`/ticket 0344; the AEDIST entry does not. Cross-project memory — out of this PR's mandate.

---

### Minor / noted

- `consider:` (correctness) `-C` is recognized only as literal `$1`; `erg-pr-merge 42 -C dir` silently no-ops the flag and dies with a generic "must run from PR branch" — reproduced live. Matches the documented "first argument" contract, so not a spec violation; a one-line die-on-misplaced-flag would remove the footgun.
- `consider:` (red team) relative-`$0` breaks `SYNC_LOCAL_MAIN` resolution after the `-C` cd — but the canonical invocation is an absolute path (required for the allow rule to match) and the failure is non-fatal (`|| true`, `[[ -x ]]` guard). Off the documented path.
- `nofollow:` (consistency) the "See nightbeat-supervisor Phase 2" cross-ref in lines 9-11 is dangling (that file has no such heading) — pre-existing, better swept with the header-comment fix above.

**Scope, backward-compat, style, prose (LLMism sweep) all clean.** The `git -C` "no realpath" semantics claim was verified empirically. Cases 1-28 are byte-behavior-identical.

MinhHaDuong:
## Review round 1 — disposition

**Fixed (a55cdb8):**
- Stale header comment (lines 9-11) now leads with `-C` and warns against the `cd PATH && …` prefix, matching SKILL.md.

**Deferred to author, tracked in ticket 0350:**
- **Red-team guard blind spot** — `guard-cd-primary-repo.sh` never sees the in-script `cd` of `-C <path>`. Not fixed here because the two proposed fixes are both design calls the author owns: (a) restricting `-C` to `in_worktree()` targets breaks the documented VM / non-worktree flow; (b) teaching the guard about `-C` contradicts that guard's explicit YAGNI "one reflex only" / "not a security boundary" stance. Practical risk is low: the destructive Step 4 is only reachable when the target already has the PR branch checked out (a primary checkout on main dies at the branch check), so `-C` cannot silently clobber primary.
- **Doc adoption** (raid Phase 7 / nightbeat §2 / AEDIST memory still teach the pre-`-C` dance) — filed in 0350 as documentation follow-up.

Keeping this PR tightly scoped: it is the raid pipeline's own merge tool, so the diff stays additive and gated on `$1 == "-C"`.

MinhHaDuong:
## /gaze actions

round: 1
tier: small (145 lines, 5 files)
adherence: PASS — 0 blocking (shell suite 35/35, pytest adherence 17 passed, erg check 342 tickets)
review: 0 new blockers — two minors, both already tracked in ticket 0350 ("not new information")
review-pr: correctness-only (small tier) — verdict approve, confidence 90; 1 consider: applied, 1 nofollow: muted
simplify: 3 findings applied in 685daba — run_C now delegates to run_merge (RUN_MERGE_CWD + passthrough args; kills the third stub-env near-copy), Case 29 (b)/(c) single-capture idiom (matches Case 7); suite re-run green 35/35

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- `erg-pr-merge -C <path> <N>` merges from an arbitrary cwd: ADDRESSED — Step -1 parse block (erg-pr-merge lines 31-46) precedes the branch check (line 104) and every git/gh/erg call; Case 29(a) closes ticket 0344 from a non-git cwd; loud-failure sub-cases (b)/(c) pass at tip 685daba.
- SKILL.md documents the flag and the bare-invocation rule synergy: ADDRESSED — c7a5694: argument-hint `[-C path] [pr-number]`, new section naming the allow rule `Bash(~/.claude/skills/merge/erg-pr-merge:*)` and warning off the `cd &&` prefix.
- Ticket-close step finds tickets/erg via the -C checkout: ADDRESSED — the cd lands before the `[[ -d tickets ]]` guard (line 156); Case 29(a) `closed_has 0344` proves the close end-to-end from a husk cwd.
- `make check` passes: ADDRESSED — re-run by the gate at tip 685daba, not taken on faith: 856 pytest passed, shell suites green, exit 0.

Unresolved review comments: none — round-1 verifiable: (stale header) fixed in a55cdb8; red-team guard-blind-spot dissent author-triaged and tracked in ticket 0350 (in this diff); consider:/nofollow: items informational/muted per tag.
Unresolved simplify: none — all applied (685daba).
Adherence: PASS

Rationale:
All four exit criteria carry direct diff- and run-level evidence; every review comment is fixed, tracked in 0350, or muted by its tag; the two-dot diff (5 files) is fully traceable to the exit criteria plus the review-disposition artifact (ticket 0350 — a legitimate deferral record, not scope creep). No blocking adherence findings, no unresolved verifiable: minors, no scope overflow. The merge decision belongs to the caller; note ticket 0350's guard-blind-spot design call awaits the author.

telemetry: tier=small wall=990s agents=8 tokens=~601k(subagents) cost~=na
note: warn thresholds touched (wall 16.5min > 15min; tokens > 500k) — run completed normally, no escalate threshold reached.

MinhHaDuong:
**Advisory reviewer seats — 0207 trial round 1 (non-blocking)**

verifiable: skills/merge/erg-pr-merge:38 — The `cd "$TARGET_DIR"` on line 38 uses plain `cd` without `cd -` or subshell, so the cwd is permanently changed for the rest of the script, which is intentional but means the script can no longer be safely called from a subshell that expects its cwd preserved — the caller must be aware of this side effect.  [openrouter-budget]
consider: skills/merge/SKILL.md:18 — The SKILL.md says `-C PATH` is "preferred for agents" but the ticket 0350 follow-up notes that raid/nightbeat/AEDIST callers still teach the old `cd` dance — the SKILL.md should either update those callers or note the inconsistency.  [openrouter-budget]
consider: skills/merge/erg-pr-merge:38 — The `cd` on line 38 does not use `realpath` or resolve symlinks, which is a deliberate choice documented in the comment, but relative `-C` paths may behave unexpectedly if the script later changes directory again (it doesn't currently, but the design is fragile).  [openrouter-budget]
verifiable: tests/test_erg_pr_merge.sh:922 — The test `run_C` function on line 922 uses `( cd "$WORK" ... )` which is a subshell, but the script itself does not use a subshell for its `cd` — the test correctly isolates the cwd change, but the script's own `cd` is permanent, which is a design asymmetry worth noting.  [openrouter-budget]
consider: tickets/0350-erg-pr-merge-c-guard-blind-spot-and-adop.erg:1 — The new ticket 0350 documents a guard blind spot but leaves the decision unresolved — the diff should either resolve it or explicitly defer it to a follow-up, not leave it as an open question in the same PR.  [openrouter-budget]

Seats: gpt-5.6-luna (frontier), deepseek-v4-flash (budget), sandboxed read-only; findings are advisory — disposition at your gate.

PR #618, opened 2026-07-14
fix(0347): fail loud on reasoning-field models in seat-runner

MinhHaDuong:
## Summary

`z-ai/glm-5.2` and `moonshotai/kimi-k2.7-code` never completed a seat review during the 2026-07-14 audits: the reviewer inside the sandbox produced NO stderr and idled until `SEAT_TIMEOUT` SIGKILLed it (empty stderr — a silent timeout). Both models answer a trivial prompt in 1-3s over the same endpoint/key, and both return a `reasoning` field alongside `content`. The seat's pinned aider/litellm venv hangs on that shape. The venv is a host-level uv-tool install, unpinnable from this repo, so the durable fix lives in `scripts/seat-runner.sh`.

## Action 1 — reasoning-shape pre-flight probe (core fix)

In `scripts/seat-runner.sh`, right after the `--health-path` probe and only when `--credential-env` is set, POST a one-token chat completion (`max_tokens=1`, target `$MODEL` with any litellm `openai/` route prefix stripped) to `${ENDPOINT}/chat/completions` via `curl -sf --max-time 15`, parse in-memory with `python3`, and if the response carries a non-empty `reasoning`/`reasoning_content` field, exit non-zero with a diagnostic naming the hang class — BEFORE any container launch. A probe that cannot connect or parse WARNs and falls through (a working model is never blocked on probe fragility).

Credential discipline: the key never touches curl argv — it goes in a mode-600 `curl -K` config file under `$WORK` (`umask 077` write, `rm -f`'d on every branch, backstopped by the EXIT trap reaping `$WORK`). The response body is parsed by piping to `python3` stdin, never written to a file the credential-scrub pass skips. `WORK`/`cleanup` trap setup moved earlier so the curlrc has a reaped home; the move is safe on early exits (empty `RELAY_PID` short-circuits, empty `SEAT_CONTAINERS` no-ops).

## Action 2 — aider `--timeout` backstop

Added `--timeout "${AIDER_API_TIMEOUT:-120}"` to the seat's aider invocation and documented `AIDER_API_TIMEOUT` next to `SEAT_TIMEOUT`. Converts any future silent litellm hang into a stderr exception the existing `tail -20 raw.err` failure tail surfaces, instead of a SIGKILL with empty stderr.

## Action 3 — hermetic tests (RED + negative-space)

In `tests/test_seat_runner.sh` (reusing the stub-podman / stub-curl idioms):
- **RED case:** a canned reasoning-shaped body must fail loud (`reasoning-field response` diagnostic) before the podman stub's `run` is reached (no `GUARD-BYPASS`). Confirmed RED on the pre-fix script (2 assertions failed: no diagnostic, `GUARD-BYPASS` present).
- **Negative-space case:** an ordinary content-only body must reach `podman run` — guards against a tautological always-block.

Both green after the fix: `tests/test_seat_runner.sh` → 55 passed, 0 failed.

## Action 4 — dropped

Not shipped. Action 1's fail-loud path already satisfies the exit criterion ("fails LOUD with a diagnostic, not a silent timeout"). A model-settings reasoning-suppression shim was descoped; no live OpenRouter probe was run, so no spend and no key handling.

## Verification

- `make check` green (exit 0); `make lint` green (17 passed).
- `/verify-adherence` → PASS (bash rules, credential-sink audit by a sonnet reviewer, `check-agnostic.sh` on `scripts` and `tickets` both clean).
- `bash -n` + `shellcheck -S warning` clean.

**Ticket:** tickets/0347-seat-stack-hangs-on-reasoning-field-models.erg

### DIFF
--- a/scripts/seat-runner.sh
+++ b/scripts/seat-runner.sh
@@ -72,6 +72,10 @@
 #                          /health); empty string skips the reachability probe.
 #
 # Env: SEAT_TIMEOUT (per-seat wall-clock cap, seconds; default 600)
+#      AIDER_API_TIMEOUT (aider/litellm per-request API timeout, seconds; default
+#      120) — a backstop that converts a silent litellm hang into a stderr
+#      exception the failure tail surfaces, instead of a SEAT_TIMEOUT SIGKILL with
+#      empty stderr (ticket 0347).
 set -euo pipefail
 
 SELF_DIR="$(cd "$(dirname "$0")" && pwd)"
@@ -150,6 +154,22 @@ if [[ -n "$CREDENTIAL_ENV" ]]; then
     CRED_ARGS=(-e OPENAI_API_KEY)
 fi
 
+# Scratch dir + cleanup trap are set up EARLY: the reasoning-shape pre-flight
+# probe below writes a mode-600 curl config file here, and the self-test, relay,
+# and clone all reuse it. cleanup reaps the relay, any orphaned seat container,
+# and $WORK on exit (an early probe/health FATAL is reaped too).
+WORK=$(mktemp -d)
+RELAY_PID=""
+SEAT_CONTAINERS=()
+cleanup() {
+    [[ -n "$RELAY_PID" ]] && kill "$RELAY_PID" 2>/dev/null || true
+    for _c in ${SEAT_CONTAINERS[@]+"${SEAT_CONTAINERS[@]}"}; do
+        podman rm -f "$_c" >/dev/null 2>&1 || true
+    done
+    rm -rf "$WORK"
+}
+trap cleanup EXIT
+
 if [[ "$SELF_TEST_ONLY" -eq 0 ]]; then
     [[ -n "$BRANCH" ]] || { echo "seat-runner: --branch required" >&2; exit 2; }
     # Fail loud if the endpoint is down (0217: unreachable endpoint = non-zero).
@@ -159,6 +179,55 @@ if [[ "$SELF_TEST_ONLY" -eq 0 ]]; then
     if [[ -n "$HEALTH_PATH" ]]; then
         curl -sf --max-time 5 "${SCHEME}://${EP}${HEALTH_PATH}" >/dev/null \
             || { echo "seat-runner: endpoint ${ENDPOINT} unreachable" >&2; exit 1; }
+    fi
+
+    # ── Reasoning-shape pre-flight probe (ticket 0347) ────────────────────────
+    # z-ai/glm-5.2 and moonshotai/kimi-k2.7-code return a `reasoning` /
+    # `reasoning_content` field alongside `content`; the seat's pinned
+    # aider/litellm venv hangs forever waiting for content it never surfaces and
+    # the outer SEAT_TIMEOUT SIGKILL leaves EMPTY stderr — a silent timeout that
+    # burns an audition's wall-clock. Catch that class HERE, host-side, before any
+    # container launch: POST a 1-token completion and fail LOUD if the response
+    # carries a non-empty reasoning field. Only runs when a real credential is
+    # injected (--credential-env). The key never touches curl argv — it goes in a
+    # mode-600 curl config file under $WORK (trap-reaped), and the response body is
+    # parsed IN-MEMORY (piped to python3 stdin), never written to a file the
+    # credential-scrub pass skips. Probe fragility (connect or parse failure) WARNs
+    # and falls through: never block a working model on a flaky probe.
+    if [[ -n "$CREDENTIAL_ENV" ]]; then
+        _probe_model="${MODEL#openai/}"          # the endpoint knows the raw provider id
+        _probe_rc="$WORK/probe.curlrc"
+        ( umask 077; printf 'header = "Authorization: Bearer %s"\n' "$OPENAI_API_KEY" > "$_probe_rc" )
+        _probe_body="$(printf '{"model":"%s","max_tokens":1,"messages":[{"role":"user","content":"ping"}]}' "$_probe_model")"
+        _probe_resp=""
+        if _probe_resp="$(curl -sf --max-time 15 -K "$_probe_rc" \
+                -H 'Content-Type: application/json' -X POST -d "$_probe_body" \
+                "${ENDPOINT}/chat/completions" 2>/dev/null)"; then
+            _probe_verdict=0
+            printf '%s' "$_probe_resp" | python3 -c '
+import json, sys
+try:
+    d = json.load(sys.stdin)
+    msg = (d.get("choices") or [{}])[0].get("message") or {}
+    r = msg.get("reasoning")
+    if r is None:
+        r = msg.get("reasoning_content")
+    present = bool(r.strip()) if isinstance(r, str) else bool(r)
+    sys.exit(0 if present else 1)
+except Exception:
+    sys.exit(2)
+' || _probe_verdict=$?
+            rm -f "$_probe_rc"
+            if [[ "$_probe_verdict" -eq 0 ]]; then
+                echo "seat-runner: FATAL model '${MODEL}' returns a reasoning-field response — known aider/litellm hang class (ticket 0347)" >&2
+                exit 1
+            elif [[ "$_probe_verdict" -eq 2 ]]; then
+                echo "seat-runner: WARN reasoning-shape probe could not parse the response from ${MODEL} — proceeding (probe is advisory)" >&2
+            fi
+        else
+            rm -f "$_probe_rc"
+            echo "seat-runner: WARN reasoning-shape probe could not reach ${ENDPOINT}/chat/completions — proceeding (probe is advisory)" >&2
+        fi
     fi
 fi
 
@@ -200,18 +269,6 @@ fi
 if [[ "$SELF_TEST_ONLY" -eq 0 && -z "$AIDER_REAL" ]]; then
     echo "seat-runner: FATAL aider not found on PATH" >&2; exit 1
 fi
-
-WORK=$(mktemp -d)
-RELAY_PID=""
-SEAT_CONTAINERS=()
-cleanup() {
-    [[ -n "$RELAY_PID" ]] && kill "$RELAY_PID" 2>/dev/null || true
-    for _c in ${SEAT_CONTAINERS[@]+"${SEAT_CONTAINERS[@]}"}; do
-        podman rm -f "$_c" >/dev/null 2>&1 || true
-    done
-    rm -rf "$WORK"
-}
-trap cleanup EXIT
 
 # ── Host-side relay: /relay.sock ⇄ the real endpoint. The container binds the
 #    socket file; under --network=none it has no other route out. ──────────────
@@ -314,7 +371,7 @@ exec ${AIDER_REAL} \\
     --message "\$(cat /prompt.txt)" \\
     --read /review.diff \\
     --yes-always --no-check-update --no-show-model-warnings --no-pretty \\
-    --no-stream --map-tokens 0
+    --no-stream --map-tokens 0 --timeout "${AIDER_API_TIMEOUT:-120}"
 EOF
 )
 review_rc=0
--- a/scripts/seat-runner.sh
+++ b/scripts/seat-runner.sh
@@ -211,16 +211,18 @@ if [[ "$SELF_TEST_ONLY" -eq 0 ]]; then
     if [[ -n "$CREDENTIAL_ENV" ]] && command -v python3 >/dev/null 2>&1; then
         _probe_model="${MODEL#openai/}"          # the endpoint knows the raw provider id
         _probe_rc="$WORK/probe.curlrc"
-        _probe_body="$(python3 -c 'import json,sys; print(json.dumps({"model": sys.argv[1], "max_tokens": 1, "messages": [{"role": "user", "content": "ping"}]}))' "$_probe_model")"
-        _probe_resp=""
-        if ! ( umask 077; printf 'header = "Authorization: Bearer %s"\n' "$OPENAI_API_KEY" > "$_probe_rc" ); then
-            echo "seat-runner: WARN reasoning-shape probe could not write its curl config — proceeding (probe is advisory)" >&2
-        elif _probe_resp="$(curl -sf --max-time 15 -K "$_probe_rc" \
-                -H 'Content-Type: application/json' -X POST -d "$_probe_body" \
-                "${ENDPOINT}/chat/completions" 2>/dev/null)"; then
-            rm -f "$_probe_rc"                    # key file gone before the body is parsed
-            _probe_verdict=0
-            printf '%s' "$_probe_resp" | python3 -c '
+        if ! _probe_body="$(python3 -c 'import json,sys; print(json.dumps({"model": sys.argv[1], "max_tokens": 1, "messages": [{"role": "user", "content": "ping"}]}))' "$_probe_model")"; then
+            echo "seat-runner: WARN reasoning-shape probe could not build its request body — proceeding (probe is advisory)" >&2
+        else
+            _probe_resp=""
+            if ! ( umask 077; printf 'header = "Authorization: Bearer %s"\n' "$OPENAI_API_KEY" > "$_probe_rc" ); then
+                echo "seat-runner: WARN reasoning-shape probe could not write its curl config — proceeding (probe is advisory)" >&2
+            elif _probe_resp="$(curl -sf --max-time 15 -K "$_probe_rc" \
+                    -H 'Content-Type: application/json' -X POST -d "$_probe_body" \
+                    "${ENDPOINT}/chat/completions" 2>/dev/null)"; then
+                rm -f "$_probe_rc"                    # key file gone before the body is parsed
+                _probe_verdict=0
+                printf '%s' "$_probe_resp" | python3 -c '
 import json, sys
 try:
     d = json.load(sys.stdin)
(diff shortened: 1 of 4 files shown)


MinhHaDuong (commented):
## Multi-perspective review — PR #618 (ticket 0347)

**Panel verdict: approve, no blockers.** Correctness → comment; Consistency, Scope, Red-team, Doc-propagation → approve. `make check` green (exit 0); seat suite 55/55 across three reviewers; scope reviewer confirmed the full 856-item pytest run. Branch is current with `origin/main`.

The core security question — can the new host-side probe leak the API key? — was **live-attacked and cleared** by the red-team reviewer: curlrc is mode-600 before the key is written (umask 077 in a subshell, `stat` confirmed 600 mid-run), the key never reaches curl argv (only `-K`), `curl -sf` discards the body on HTTP failure so a 401 echoing a key-shaped marker surfaced nowhere in stdout/stderr, and the body is only ever piped to python3 stdin. RED test confirmed genuinely RED on the pre-fix script.

### Findings (all non-blocking)

**Correctness**
- `verifiable:` **`reasoning_content` shape has zero test coverage.** The probe checks `reasoning` then falls back to `reasoning_content` (`scripts/seat-runner.sh`), and the ticket names `moonshotai/kimi-k2.7-code` — a Moonshot/DeepSeek-R1-lineage API, the family that commonly returns `reasoning_content`. But `grep -n reasoning_content tests/test_seat_runner.sh` → 0 matches; the RED/negative pair only exercises `reasoning`. Exit criterion "canary pins the reasoning-response shape" is half-pinned. A third stub case with `{"message":{"content":"ok","reasoning_content":"..."}}` closes it cheaply.
- `verifiable:` **Non-{0,1,2} probe verdict silently proceeds** — the `if -eq 0 … elif -eq 2` has no `else`. Reproduced with a stub `python3` exiting 42: no WARN emitted. Low impact (a broken host python3 fails the relay startup moments later anyway), but a defensive `else → WARN` closes it.
- `consider:` **`max_tokens=1` may not elicit the reasoning field** on the real API — if a provider only populates `reasoning` after enough content, the 1-token probe could come back reasoning-empty for a model that hangs on a full call (false "proceed"). Unverified — needs a live glm-5.2/kimi probe with a real key. Worth confirming before trusting in production. (Red-team flagged the same false-negative class.)
- `consider:` **`curl -sf` treats HTTP 4xx/5xx identically to unreachable** — an expired key WARNs-and-proceeds rather than surfacing a distinct diagnostic. Matches the stated "never block on a flaky probe" design; a bad key still fails loud downstream.

**Red-team** (all reproduced, none leak the key)
- `consider:` **JSON injection via unescaped `$MODEL`.** `printf '{"model":"%s",...}' "$_probe_model"` does no escaping; `--model 'openai/foo","stream":true,...'` produces attacker-extended JSON that can neuter the probe (degrade it to its WARN fallback — silently disabling the very hang-detection this PR adds). No key leak. Low exploitability today (no caller threads untrusted input into `--model`). Build the body with `python3 … json.dumps` instead of `printf %s`.
- `consider:` **Unguarded curlrc write breaks the "probe fragility never blocks a working model" contract.** Reproduced with an unwritable temp dir: the bare `( umask 077; printf … > "$_probe_rc" )` fails under `set -e` and kills the script with a raw bash error instead of the documented WARN-and-proceed. Narrow trigger (EACCES/disk-full on a fresh mktemp dir).
- `nofollow:` No `--max-filesize` on the probe curl — a compromised endpoint could return an unbounded body buffered into a shell var then `json.load`. Mirrors the existing accepted trust model for the relay path; not a new exposure class.

**Consistency**
- `consider:` Tier label `2a⁶` (`tests/test_seat_runner.sh:382`) breaks the file's Unicode prime-mark sequence (`2a′ 2a″ 2a‴ 2a⁗`). Cosmetic; likely forced by Unicode lacking a quintuple-prime glyph.
- `nofollow:` Ticket 0347's "Relevant files" still points at `tickets/0346-…` which is now archived under `tickets/closed/`. Pre-existing, not introduced here.

**Doc propagation**
- `consider:` The `--credential-env` entry in the `Usage:` block doesn't note that setting the flag now also activates a host-side network call (the probe) before container launch. Not false (the `Env:` block and inline comment cover it), but a Usage-only skimmer would miss the second behavior. Low blast radius.

### Scope
Every changed line traces to a ticket exit criterion or a logged, revised Action; only `seat-runner.sh`, `test_seat_runner.sh`, and the ticket file are touched. The WORK/cleanup-trap relocation is a byte-identical move justified by the probe's need for a reaped curlrc home — no double-definition (verified). Note (`consider:`): the exit criterion is satisfied via the **"fail loud"** half of its OR — glm-5.2 and kimi-k2.7-code still cannot be audition-reviewed, they now fail with a diagnostic instead of hanging. The ticket itself narrows to this (aider/litellm venv unpinnable from the repo), so it's a legitimate reading, but downstream ticket 0346's audition board should know reasoning-emitting candidates (~half the leaderboard) remain out of reach.

### Recommendation
Merge-ready. The one finding I'd suggest landing before or right after merge is the `reasoning_content` test case (the ticket's own canary criterion is only half-covered, and it's the shape the named kimi model most plausibly returns). The JSON-escape and curlrc-write-guard are worth a quick follow-up if `--model` provenance ever loosens or the probe is reused — a short ticket would capture them well. Everything else is nit/cosmetic or pre-existing.

MinhHaDuong (commented):
## Multi-perspective panel review — 5 agents (correctness, consistency, scope, red-team, doc-propagation)

**Verdict: comment** — no blockers. Credential handling is sound, scope is clean, all 58 seat tests + `make check` pass. Findings below are hardening/hygiene; the author arbitrates. Two of them converge across three agents and are worth a small follow-up.

### Confirmed clean (positive findings)
- **Red-team, high confidence — no credential leak.** `$WORK` is `mktemp -d` (0700); `probe.curlrc` is written mode-600 via `( umask 077; printf … )` (umask set in the subshell before the redirection opens the file — no world-readable window); the key never touches curl argv (`-K "$_probe_rc"`, not `-H`), never appears in any WARN/FATAL message (those interpolate only `${MODEL}`/`${ENDPOINT}`), and `_probe_resp` is piped to python3 stdin, never echoed. Dual `rm -f` + the relocated `trap cleanup EXIT` (now installed *before* the probe) reap the key file even on a FATAL exit between write and rm. `curl -sf` has no `-L`, so a malicious 3xx from the endpoint is not followed to a second host. The python parser builds the request via `json.dumps` and consumes the response under a bare `except`, so a hostile `$MODEL` id or adversarial response body is not a shell/format-string injection vector.
- **Scope, high confidence — clean.** Exactly the three files the ticket names; no drive-by edits. The WORK/cleanup-trap relocation is load-bearing (the probe writes `$WORK/probe.curlrc`), not an unrelated refactor. Dropping Action 4 is legitimate — exit criterion 1's "*or* fails loud with a diagnostic" disjunct is satisfied by the probe without a reasoning-suppression shim or live OpenRouter spend.
- **Correctness, high confidence — verdict logic sound.** `|| _probe_verdict=$?` correctly maps the parser's exit codes (0=present/FATAL, 1=absent/proceed, 2=unparseable/WARN) under `set -euo pipefail`; RED reproduced on the pre-fix script, GREEN after.
- **Doc-propagation, high confidence — callers handle the new FATAL correctly.** Both `reviewers.sh` callers (`request` fail-open, `audition` fail-loud) treat any non-zero exit generically; neither string-matches the reason, so the new "reasoning-field response" FATAL is handled correctly today.

### Findings (all non-blocking)

1. **`consider:` (3-agent convergence — consistency, red-team, doc-prop) — the header Containment section doesn't document the new host-side call.** `scripts/seat-runner.sh` header enumerates every trust-boundary mechanism (relay socket, per-seat timeout, CREDENTIAL SCRUB), but the new reasoning probe adds the **first host-side outbound POST carrying the real credential**, bypassing the `--network=none` + relay egress path the header describes, and writes the key to host disk (`$WORK/probe.curlrc`). The container-egress claims stay true, but the header reads as the file's security contract and is now incomplete. Recommend a short Containment paragraph mirroring how PER-SEAT TIMEOUT / CREDENTIAL SCRUB are documented.

2. **`consider:` (red-team) — trust-surface widening worth an explicit author nod.** Item 1 is not a bug — the host already holds the key, and the probe targets the same `$ENDPOINT` the seat would reach via the relay anyway (no new destination or party). But it is the first time the *host* makes the model call directly. Flagging for a "yes, host-side pre-flight calls are acceptable" sign-off.

3. **`consider:` (correctness, tempered) — the reasoning-probe canary is environment-gated and does not run in CI.** The RED/negative-space tier is gated on `command -v aider` (test_seat_runner.sh:282-ish), and CI installs no aider, so `test_bash_suites.py` runs the suite but silently SKIPs this tier — exit criterion 2's canary fires only on a dev machine with aider. This follows the suite's **pre-existing** gating pattern (the containment tier is likewise podman-gated), so it is not a regression this PR introduces; noted so the durability of the canary is understood. A future improvement could split the pure-parse assertion (no aider needed) from the podman-reaching path.

4. **`consider:` (correctness + consistency) — no test asserts the aider `--timeout`/`AIDER_API_TIMEOUT` wiring.** The suite's tier-1b idiom adds an `assert_contains` for every new flag, but the new `--timeout "${AIDER_API_TIMEOUT:-120}"` on the aider invocation has none. A refactor could silently drop it. Cheap source-grep assertion.

5. **`consider:` (correctness) — the WARN/fragility branches are untested.** None of connect-fail, unparseable-body, curlrc-write-fail, or no-python3 has a test; a regression flipping WARN-and-proceed into a hard block (false-blocking every model) would pass unnoticed.

6. **`consider:` (correctness) — probe only runs with `--credential-env`.** A reasoning-emitting model behind a local unauthenticated endpoint still hits the original silent hang. The two ticket-named models are cloud/credentialed, so the literal exit criteria hold; flag in case broader coverage was intended.

7. **`nofollow:` (red-team) — curlrc `header = "…%s…"` doesn't escape an embedded `"` in the key.** Fails safe (parse error → WARN → proceed, no leak); key is operator-provisioned. Robustness only.

8. **`consider:` (red-team) — no response-size cap on the probe fetch** (`--max-time 15` bounds wall-clock, not bytes). A compromised third-party OpenRouter backend could stream a large body into a bash var → python3. Minor DoS, not a leak.

9. **`nofollow:` (correctness) — an empty `choices: []` is treated as "no reasoning / proceed"** (`d.get("choices") or [{}]` masks the empty-list case before any exception), rather than routed to the unparseable/WARN branch. Low severity.

_Panel: correctness=comment, consistency=comment, scope=approve, red-team=approve, doc-propagation=comment._

MinhHaDuong:
## Multi-perspective panel review — round 3, 5 agents (correctness, consistency, scope, red-team, doc-propagation)

**Panel verdict: request-changes** (one blocker, independently reproduced by two agents and confirmed by me on read).

`make check` → 856 passed (54.7s). `bash tests/test_seat_runner.sh` → 60 passed, 0 failed (real podman + image available in the review worktree, so tier-2b containment checks ran, not skipped).

### Follow-up commits since round 2 — resolution check

Verified against current code, not just commit messages:
- `f8142a6`: JSON-injection via unescaped `$MODEL` → **fixed**, request body now built via `python3 json.dumps` (`scripts/seat-runner.sh:214`). curlrc-write-failure abort under `set -e` → **fixed**, the write now sits as an `if !(...)` condition that WARNs instead of aborting (`:216-217`, reproduced against an unwritable `$WORK`). `reasoning_content` shape coverage → **fixed**, second RED case added (`tests/test_seat_runner.sh:438-440`).
- `2589d25`: Containment-header doc of the host-side probe call → **fixed** (`scripts/seat-runner.sh:58-67`). RED tests moved to a python3-only gate so they run in the CI merge gate instead of being skipped without aider/git → **fixed**, verified by reading the gate condition. `AIDER_API_TIMEOUT`/`--timeout` wiring assertions → **fixed**, added to tier 1.
- **Not fixed, still open from round 1**: "non-{0,1,2} probe verdict silently proceeds" (see Blocker below) — untouched by either follow-up commit.
- **Partially fixed**: round 1's Usage-block gap (`--credential-env` doesn't note it now triggers a host-side call) — round 2 only added the Containment paragraph, which documents the mechanism but the `Usage:` bullet itself (`:76-78`) still doesn't cross-reference it.

### Blocker

**`scripts/seat-runner.sh:236-242`** — the reasoning-probe verdict switch (`if -eq 0 … elif -eq 2 … fi`) has no `else`. Any verdict code other than `0` or `2` is silently treated as `1` (safe) and the script proceeds straight to `podman run`. Reproduced independently by two review agents: stub `curl` returning a genuine non-empty `reasoning` field, stub the verdict-parsing `python3` invocation to exit `42` (simulating a crashed/killed interpreter on that one call — plausible under resource pressure in a raid/nightbeat batch) → no FATAL, no WARN, `podman run` reached with a known-bad reasoning-emitting model. This reproduces the exact silent-proceed failure class ticket 0347 exists to close, under a narrower (but real) trigger than the original bug. Fix is a one-line `else` that WARNs loudly on an unrecognized verdict rather than silently treating it as "safe."

This same gap was raised (as a non-blocking `verifiable:`) in round 1 and survived two hardening commits untouched — elevating it now because two independent agents reproduced it against the *current* code and it directly undermines the exit criterion ("fails LOUD with a diagnostic, not a silent timeout").

### Tagged minors (non-blocking)

- `verifiable:` `scripts/seat-runner.sh:214` — `_probe_body="$(python3 -c ...)"` is an unguarded command substitution outside any `if`/`elif` test (unlike every other fallible substitution in this block). If `python3` resolves via `command -v` but crashes on this specific invocation, `set -e` aborts the whole run with a raw exit code and no WARN — for *any* model, not just reasoning-emitting ones. Contradicts the documented "every fragility WARNs and falls through" contract. Reproduced by stubbing `python3` to `exit 42` unconditionally: seat-runner died at this line, before reaching the curlrc-write guard.
- `consider:` `scripts/seat-runner.sh:216` (red-team, reproduced) — curlrc's `header = "Authorization: Bearer %s"` doesn't escape an embedded `"` in the key; fails safe (curl errors → WARN-and-proceed, no leak, no crash) but worth a quick fix.
- `consider:` `scripts/seat-runner.sh:218-220,244` (red-team, reproduced) — `curl -sf` treats HTTP 4xx/5xx identically to "unreachable": a stubbed 401 lands in the same WARN-and-proceed branch as a connection refusal, so an expired/revoked key is indistinguishable from a network blip at the probe layer.
- `consider:` `scripts/seat-runner.sh:227` (red-team, reproduced) — `(d.get("choices") or [{}])[0]` treats `{"choices": []}` as "no reasoning field" (verdict 1, silent proceed) rather than routing to the unparseable/WARN branch (verdict 2); piping that body through the parser one-liner confirmed exit 1, not 2.
- `consider:` `scripts/seat-runner.sh:218` (red-team) — no `--max-filesize` cap on the probe fetch; only wall-clock (`--max-time 15`) is bounded.
- `consider:` `scripts/seat-runner.sh:237` (consistency) — the FATAL message embeds `(ticket 0347)` inline; none of the file's other 26 runtime diagnostics cite a ticket number (the convention keeps ticket refs in `#`-comments, not operator-facing text).
- `consider:` `scripts/seat-runner.sh:76-78` (doc-propagation) — the `--credential-env VAR` `Usage:` bullet still doesn't cross-reference the host-side probe side effect the new Containment paragraph documents; a reader who skips straight to `Usage:` still gets an incomplete picture.
- `consider:` `tickets/0347-seat-stack-hangs-on-reasoning-field-models.erg` (doc-propagation) — the "Relevant files" section still points at `tickets/0346-...`, archived to `tickets/closed/` before this ticket was even created; pre-existing per round 1, but this PR edited this same file's body and left the stale path in place.
- `nofollow:` `tests/test_seat_runner.sh:391` (consistency) — the `2a⁶` tier label still breaks the file's prime-mark sequence (`2a′ 2a″ 2a‴ 2a⁗`); cosmetic, unfixed since round 1, no test-behavior impact.
- `nofollow:` (red-team) — `OPENAI_API_KEY` is visible via `/proc/<pid>/environ` to same-UID processes (inherited by every child of seat-runner, curl included); this is the pre-existing `--credential-env` design, not a new exposure introduced by this PR.
- `nofollow:` (scope, carried from round 1) — exit criterion (a) is satisfied only via its "fails LOUD" disjunct: glm-5.2/kimi-k2.7-code still cannot complete a seat review, they now fail cleanly pre-container instead of hanging. Ticket-sanctioned boundary; flagged so ticket 0346's audition board (which scans ~half the OpenRouter leaderboard) knows reasoning-emitting candidates are rejected at pre-flight, not reviewed.

### Scope and credential safety — clean

Scope: exactly the three files the ticket names (`scripts/seat-runner.sh`, `tests/test_seat_runner.sh`, the ticket file); the `WORK`/`cleanup`-trap relocation is byte-identical (confirmed via diff) and load-bearing (the probe needs a reaped curlrc home); no drive-by changes; `tests/test_seat_runner.sh` is auto-wired into `make check` via the existing `test_bash_suites.py` glob, so exit criterion (c) is mechanically satisfied.

Red-team, high confidence: the credential does **not** leak via argv (checked `/proc/<pid>/cmdline` and `ps -ef` mid-request — only the `-K` path appears), via stderr/error paths (`2>/dev/null` on the probe curl call), or via a scrub-skipped file (curlrc is `rm -f`'d before the response is parsed, response only ever piped to `python3` stdin). The probe reaches no new network destination — same `${ENDPOINT}` the relay already dials — it's a new host-side *capability* (making the call directly with the credential the host already holds), not new *reach*.

### Recommendation

Fix the missing `else` at `scripts/seat-runner.sh:236-242` (and, while touching that block, guard the line-214 substitution the same way its neighbors already are) — both are one-line-scale changes. Everything else is hardening/hygiene the author can triage; none of it blocks merge once the verdict-switch gap is closed.

MinhHaDuong:
**Advisory reviewer seats — 0207 trial round 1 (non-blocking)**

verifiable: scripts/seat-runner.sh:234 — When `reasoning` is present but empty while `reasoning_content` is non-empty, the fallback is skipped and the probe incorrectly proceeds despite the documented requirement to reject either non-empty reasoning field.  [openrouter-frontier]

Seats: gpt-5.6-luna (frontier), deepseek-v4-flash (budget), sandboxed read-only; findings are advisory — disposition at your gate.

MinhHaDuong:
## /gaze actions

round: 2
tier: full
adherence: PASS — 0 blocking (bash -n clean, shellcheck -S warning clean, seat suite green, make lint 17/17)
review: 2 minor findings (both folded into the panel blocker/verifiable set)
review-pr: 1 synthesized panel comment posted (5 perspectives; request-changes → 1 blocker, 1 verifiable, 7 consider, 3 nofollow)
simplify: 0 fixes applied — 6 findings noted-skipped (deliberate credential hygiene on the double `rm -f`; header comment is a round-2 review deliverable; verdict-capture idiom is the documented safe pattern; test fixture/stub dedup requires refactoring pre-existing EXFIL* blocks outside the diff; audition-loop probe hoist + python3-spawn merge live outside the diff / contradict the ticket-specified curl -K path)
fix agent: 1 commit (459ed95) — total verdict switch + guarded probe-body build, 2 new RED-confirmed tests (60+2fail → 62/0)

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- glm-5.2 / kimi-k2.7-code complete a seat review OR fail LOUD with a diagnostic: ADDRESSED — probe at scripts/seat-runner.sh:211-252 (459ed95); FATAL on verdict 0, WARN on 2, WARN on any unexpected code (new catch-all :243-244); RED canaries for both model shapes block before podman run; satisfied via the ticket-sanctioned fails-LOUD disjunct.
- Canary test pins the reasoning-response shape: ADDRESSED — `reasoning` (glm-5.2) and `reasoning_content` (kimi-k2.7-code) RED cases, plus round-2-fix canaries for parser-crash (exit 42 → WARN) and body-builder-crash (→ WARN, no set -e abort). All pass: tests/test_seat_runner.sh 62/62.
- make check passes: ADDRESSED — 856 passed at HEAD 459ed95 (independently re-run by the gate); all 9 CI checks green.

Unresolved review comments: none — round-3 blocker (verdict-switch fallthrough) and verifiable (unguarded `_probe_body` substitution) both RESOLVED by 459ed95, verified by diff read + the reproducing stubs now passing. Remaining consider:/nofollow: items (curlrc quote-escaping, 4xx-vs-unreachable conflation, empty-choices masking, --max-filesize, Usage cross-ref, stale 0346 path in ticket, 2a⁶ label, /proc environ pre-existing) are informational — author triage, none gate.
Unresolved simplify: none (must-fix: none; skips documented above; the audition-loop probe-hoist efficiency note is follow-up ticket material, not a gate item)
Adherence: PASS

Rationale:
Both round-1 REROLL items were fixed and independently re-verified at the new HEAD: the reasoning-probe verdict switch is now total (0 FATAL / 1 silent proceed / 2 WARN / else WARN naming the code), and the probe-body command substitution is guarded so a python3 crash WARNs-and-falls-through instead of aborting the seat run under set -euo pipefail. The gate read the fix diff directly, traced the verdict-1 and unexpected-code control flow by hand, and re-ran the full evidence (62/62 seat tests, 856/856 pytest, shellcheck clean) rather than trusting the fix agent's claims. Scope of the fix commit is confined to the three ticket-named files. No new blocker introduced — only failure branches gained diagnostics; working-model paths are behaviorally unchanged.

telemetry: tier=full wall~=2100s agents=10(+5 panel-inner) tokens~=933k cost~=na
note: wall exceeded the 15-min warn threshold (full battery + 1 REROLL round); flagged per telemetry policy.

PR #630, opened 2026-07-14
fix(0339): scope endpoint credential out of the containment self-test

MinhHaDuong:
## What

`run_seat` in `scripts/seat-runner.sh` unconditionally spliced `CRED_ARGS` — the real `-e OPENAI_API_KEY` when `--credential-env` is set — into **every** podman run, including the pre-review containment self-test (`PROBE=$(run_seat …)`). That container only probes write/secret/network isolation and never needs the endpoint key. Latent least-privilege gap (raised by the PR #603 red-team panel; `$PROBE` also bypassed the 0207 scrub, moot once the key never enters that container).

Fix: `run_seat` gains a `--no-cred` first-arg flag that empties a **local** `cred_args` copy for that call; the self-test call site passes it. The global `CRED_ARGS` assignment, the export, the scrub chokepoint, and the review-run call site are untouched — only the local var seen by the self-test changes, so the sandbox posture is unchanged and the review run still receives the key. Chosen over a wrapper/global toggle to match the file's flag-parse + optional-array-splice idioms and keep the change to one function.

## Tests (`tests/test_seat_runner.sh`)

- **tier 2a″** — FLIP the self-test-argv assertion from "carries bare `-e OPENAI_API_KEY`" to "omits it". The `ENV_CAP` assertion is left intact.
- **tier 2a⁗ (exfil block)** — add per-branch argv capture (`PODMAN_SELFTEST_ARGV_CAP` / `PODMAN_REVIEW_ARGV_CAP`). Assert the self-test container **omits** the key (RED before the fix) and the review run still **carries** it (invariant guard, green before and after — catches a blanket-strip wrong fix).

### env-vs-argv precision note

Argv is the **only** discriminating channel. The runner exports `OPENAI_API_KEY` into its own process env for the review run, and the stub podman inherits that env on **both** branches regardless of the container's `-e` flags — so an env-dump assertion would be a tautology (passes before AND after the fix). The `-e OPENAI_API_KEY` passthrough puts the bare var name on its own argv line; that is what the assertions grep.

### RED → GREEN evidence

Against the **unmodified** `seat-runner.sh`, the two new/flipped assertions fail (RED); the invariant guard passes:

```
FAIL: credential-env: self-test container carries the endpoint key (--no-cred not applied)
FAIL: no-cred: self-test container carries the endpoint key (should be stripped)
PASS: no-cred: review run still carries the endpoint key (invariant)
Results: 62 passed, 2 failed
```

After the fix:

```
PASS: credential-env: self-test container omits the endpoint key (--no-cred, ticket 0339)
PASS: no-cred: self-test container omits the endpoint key (ticket 0339)
PASS: no-cred: review run still carries the endpoint key (invariant)
Results: 64 passed, 0 failed
```

`make check`: 858 passed. `/verify-adherence`: PASS (adherence suite 18/18) — `/gaze` may skip the mechanical phase.

## Invariants held

- Review run still receives the real credential (proven by the review-argv guard).
- Scrub chokepoint and all three echo paths untouched; sandbox posture unchanged.
- Tier-1 source assertion `CRED_ARGS=(-e OPENAI_API_KEY)` still matches (global assignment untouched).
- 0347's reasoning-shape probe and `AIDER_API_TIMEOUT` regions untouched.

## Rider commit (no close claim)

This branch also carries a separate chore-close commit for **ticket 0344** (`e509b47`): PR #617 merged with a `**Ticket:**` line but erg-pr-merge ran from a checkout-less husk cwd, so the `[[ -d tickets ]]` guard skipped the close — the exact defect #617 fixes. Closed and archived manually per the rules/git.md recipe (staged the `Closed:` edit before the `git mv`; commit carries 2 insertions, not a bare 100%-rename). No `**Ticket:**` close line for 0344 here — it is already-merged cleanup, not a claim.

**Ticket:** tickets/0339-seat-self-test-cred-args-scoping.erg

### DIFF
--- a/scripts/seat-runner.sh
+++ b/scripts/seat-runner.sh
@@ -152,7 +152,9 @@ fi
 #    (NOT a secret), passed inline. With --credential-env NAME: read the secret
 #    from that env var HERE, export it, and pass the BARE `-e OPENAI_API_KEY`
 #    form so podman reads the value from the runner's env — never argv (argv
-#    leaks to `ps -ef`). Fail loud if the named var is empty; never echo it. ────
+#    leaks to `ps -ef`). Fail loud if the named var is empty; never echo it.
+#    The containment self-test omits this credential entirely (run_seat --no-cred,
+#    ticket 0339): it probes only isolation and never needs the endpoint key. ────
 CRED_ARGS=(-e OPENAI_API_KEY=local-dummy)
 if [[ -n "$CREDENTIAL_ENV" ]]; then
     if [[ -z "${!CREDENTIAL_ENV:-}" ]]; then
@@ -324,6 +326,14 @@ fi
 
 # ── The sandboxed seat launcher (timeout-wrapped, reaped; network-denied) ─────
 run_seat() {
+    # --no-cred (first arg) runs the container with NO endpoint credential: the
+    # containment self-test only probes write/secret/network isolation and never
+    # needs the real key (ticket 0339, least-privilege). Only the review run gets it.
+    local cred_args=("${CRED_ARGS[@]}")
+    if [[ "${1:-}" == "--no-cred" ]]; then
+        cred_args=()
+        shift
+    fi
     local cname="seatrun-$$-${RANDOM}"
     local rc=0
     SEAT_CONTAINERS+=("$cname")
@@ -345,7 +355,7 @@ run_seat() {
         -e PATH="/usr/bin:/bin" \
         -e TERM=dumb \
         -e COLUMNS=500 \
-        ${CRED_ARGS[@]+"${CRED_ARGS[@]}"} \
+        ${cred_args[@]+"${cred_args[@]}"} \
         -e OPENAI_API_BASE="$CONTAINER_BASE" \
         -w /repo \
         "$IMAGE" "$@" || rc=$?
@@ -356,7 +366,7 @@ run_seat() {
 
 # ── Containment self-test (0217: proven, not assumed; alive-then-blocked) ─────
 echo "seat-runner: containment self-test..." >&2
-PROBE=$(run_seat bash -c '
+PROBE=$(run_seat --no-cred bash -c '
     echo SANDBOX-ALIVE
     touch /repo/PWNED 2>/dev/null && echo WRITE-ALLOWED || echo WRITE-BLOCKED
     cat '"$HOMEDIR"'/.ssh/id_* '"$HOMEDIR"'/.claude/scripts/bash-env.sh \
--- a/tests/test_seat_runner.sh
+++ b/tests/test_seat_runner.sh
@@ -217,11 +217,16 @@ STUB
         # The secret reaches the container through the process env podman inherits...
         assert_contains "credential-env: secret reaches podman via process env" \
             "OPENAI_API_KEY=${PROBE_VAL}" "$(cat "$ENV_CAP")"
-        # ...as a BARE -e passthrough in argv (a line that is exactly the var name)...
+        # ...but NOT into the containment self-test container: that run is invoked
+        # with `run_seat --no-cred` (ticket 0339, least-privilege), so the endpoint
+        # key must be ABSENT from the self-test argv (this is the only podman run
+        # under --self-test-only). The ENV_CAP assertion above is unaffected: the
+        # runner exports OPENAI_API_KEY into its OWN process env for the review run,
+        # which the stub inherits regardless of the -e argv the container receives.
         if grep -qx 'OPENAI_API_KEY' "$ARGV_CAP"; then
-            pass "credential-env: argv carries bare -e OPENAI_API_KEY passthrough"
+            fail "credential-env: self-test container carries the endpoint key (--no-cred not applied)"
         else
-            fail "credential-env: argv carries bare -e OPENAI_API_KEY passthrough"
+            pass "credential-env: self-test container omits the endpoint key (--no-cred, ticket 0339)"
         fi
         # ...and the VALUE must NEVER appear in argv (that is the ps -ef leak).
         if grep -qF "OPENAI_API_KEY=${PROBE_VAL}" "$ARGV_CAP"; then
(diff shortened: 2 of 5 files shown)


MinhHaDuong (commented):
## `/review-pr` — 5-agent panel (correctness, consistency, scope, red team, doc propagation)

**Consensus: APPROVE.** Four approve, one comment; no blockers. Two agents independently reproduced the RED→GREEN evidence (revert the `--no-cred` at the self-test call site → `62 passed, 2 failed` with the exact `FAIL:` text quoted in the PR body → restore → `64 passed`), and `make check` (858 passed) was reproduced. The fix is a genuine least-privilege reduction, not cosmetic.

### Red team — the security claim holds (high confidence)
Empirically verified against this host's podman 5.7.0: podman does **not** auto-forward the invoking process's env; forwarding is strictly `-e`-gated. So the runner's unconditional `export OPENAI_API_KEY` has zero effect on a container launched without `-e OPENAI_API_KEY` in argv — stripping the flag for the self-test (`cred_args=()`) completely removes the credential from that container's environment. `--no-cred` is not on the top-level CLI parser and both `run_seat` call sites pass a hardcoded literal first arg, so no attacker/diff-controlled data can reach the flag position (no accidental leak, no accidental strip of the review run). `set -euo pipefail` edges are safe: `${cred_args[@]+"${cred_args[@]}"}` with an empty array and `shift` on a lone `--no-cred` both verified non-aborting. Scrub chokepoint and SEAT_CONTAINERS cleanup untouched.

### Correctness / Consistency / Scope — clean
- Flag-parse, local-copy semantics, and call-site wiring correct; the `${arr[@]+"${arr[@]}"}` splice matches 3 existing uses in the file; complies with `rules/coding-bash.md`.
- The argv-based test (rather than the ticket's originally-prescribed env-dump) is the *correct*, non-tautological channel — independently confirmed that an env-dump assertion passes identically before and after the fix because the stub podman inherits ambient env regardless of `-e`. The review-run invariant guard correctly catches a blanket-strip wrong-fix.
- Diff touches only the 4 expected files; no drive-by refactors, no test bloat. Rider commit `e509b47` (ticket 0344 close) is a correct non-bare rename (`+2 -0`, `Closed:` header + log line staged before `git mv`), correctly carries **no** `**Ticket:**` close line, and is scope-isolated to ticket files — acceptable already-merged-cleanup rider.

### Findings (non-blocking)

- **`verifiable:` — comment inaccuracy, `tests/test_seat_runner.sh:224`** (doc propagation, high confidence). The tier-2a″ comment states the runner "exports OPENAI_API_KEY into its OWN process env **for the review run**, which the stub inherits." But this block invokes `bash "$SR" --self-test-only …`, and the review run is gated behind `SELF_TEST_ONLY -eq 0` (`scripts/seat-runner.sh:184`), so no review run executes in this path; the `export` runs unconditionally before the gate, not "for the review run." The near-identical comment in the tier-2b exfil block (line ~348) correctly says "which BOTH stub runs inherit." Suggest: change "for the review run" → "unconditionally once `--credential-env` is set" (or similar). One-line fix, worth it before merge but not gate-worthy.

- **`consider:` — implementation is stricter than the ticket wording** (correctness, scope, doc propagation all noted this). Exit criterion 2 reads "the self-test env carries only the dummy key," but `cred_args=()` strips `CRED_ARGS` entirely, so the self-test container gets **no** `OPENAI_API_KEY` at all — a superset (more least-privilege) fix. No functional risk (the probe script never references the var). A one-line ticket-log note would reconcile the delta; `erg-pr-merge` closes on merge regardless.

- **`consider:` — rider 0344 cited in prose, not via `Ticket-ref:`** (consistency). Functionally harmless (no line for `erg-pr-merge` to misparse) — the workflow rules suggest a `Ticket-ref: tickets/0344-…` tag for cite-without-close, but the prose "Rider commit (no close claim)" heading conveys the same intent and is unambiguous.

- **`nofollow:` — ticket 0339 exit-criteria checkboxes left unchecked.** Sampled 5 recently-closed tickets; unchecked-at-close is the prevailing repo norm here. No action.

MinhHaDuong:
## /gaze actions

round: 1
tier: small
adherence: PASS — 0 blocking (adherence suite 18/18; seat-runner suite 64/64 live, real podman/aider tier; bash -n + shellcheck clean)
review: 3 findings (1 minor, 2 nits) — approve
review-pr: correctness-only panel (small tier) — approve, confidence 0.9; RED→GREEN independently reproduced (62/2 pre-fix, 64/0 on HEAD)
simplify: skipped (PR merged mid-run — no live branch to commit fixes to)

**Note:** PR 630 was merged at 2026-07-14T22:22Z (merge commit 15ac353, ticket 0339 closed via 307c4bf) while this gaze run was in flight. This verdict is post-hoc, for the record; it gated no merge.

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Self-test container never receives the real endpoint key: ADDRESSED — `run_seat --no-cred` empties a local `cred_args` copy (scripts/seat-runner.sh, commit 2367fec); self-test call site `PROBE=$(run_seat --no-cred bash -c …)`; review-run call site and global CRED_ARGS untouched (only 2 call sites repo-wide, verified).
- A test proves the self-test env carries only the dummy key: ADDRESSED (over-satisfied: no key at all) — flipped tier-2a assertion + new per-branch argv captures (PODMAN_SELFTEST_ARGV_CAP / PODMAN_REVIEW_ARGV_CAP) at tests/test_seat_runner.sh:227-231, 337-368, incl. the review-run invariant guard against a blanket-strip wrong fix. RED reproduced against pre-fix script by two independent agents.

Unresolved review comments: none — prior /review-pr panel's two actionable findings fixed by 42ef804 (verified by isolated diff); this run's fresh findings are all tagged non-blocking:
- consider: SELFTEST_ARGV assertions pass silently on a missing capture file (asymmetric with REVIEW_ARGV's fail-on-missing); latent robustness gap, confirmed not currently triggered.
- nofollow: rider commits e509b47 (0344 manual close — verified well-formed per rules/git.md: rename + 2 insertions, disclosed in PR body, no close claim) and 42ef804 bundled in the PR.
- nits: run_seat consumes literal `--no-cred` from argv (both call sites hardcoded); pre-existing PROBE command-substitution subshell hides the self-test cname from the parent SEAT_CONTAINERS reap array (predates this PR).

Unresolved simplify: none (phase skipped)
Adherence: PASS

Rationale:
Both exit criteria carry file:line mechanism evidence, live-reproduced RED→GREEN, and the scope from the true merge-base (21bbc27) maps exactly to the two criteria plus the declared rider — nothing else. All open panel findings are consider/nofollow/nit class; no verifiable: minor is unresolved.

telemetry: tier=small wall=~840s agents=4 tokens=350646 cost~=na

PR #632, opened 2026-07-14
chore(0348,0353): close wontfix — harness cool-down, no second-order tooling

MinhHaDuong:
Closes tickets 0348 (scores read-back / help verb / stale pointer) and 0353 (latency tracking / SLOW elimination) as wontfix, per the author's cool-down doctrine: close-one-open-two follow-up minting is exactly the second-order tooling pattern to stop. Neither ticket blocks a merge, corrupts state, or affects a science project. Tickets pre-closed in the diff (Closed: header + closed log line + archived to tickets/closed/), so this PR itself claims no close.

Ticket-ref: tickets/closed/0348-reviewers-scores-readback-help-verb-stal.erg
Ticket-ref: tickets/closed/0353-audition-latency-tracking-relative-slow.erg
Ticket: none

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #634, opened 2026-07-14
skill(reviewers): scores read-back + help verb, refresh stale trial pointer (0348)

MinhHaDuong:
## Summary

Closes three /reviewers discoverability gaps (child of 0205), surfaced while presenting the CLI.

- **`scores [seat-or-candidate]`** — read back the fixed-schema trial cards (`scorecard` lines and `audition` blocks) as one sortable comparison table. Corpus-wide (every `*.erg` under `tickets/`, including `tickets/closed/`) and **read-only**: never edits a roster, never writes an `erg log` line. Scans only each ticket's `--- log ---` section, so a card quoted in a body as documentation is never mistaken for a result. A malformed line WARNs on stderr (harvest convention). Adds `_card_field`, a sibling of `_contract_field` for the space-delimited cards.
- **`help`** — prints the usage block to stdout and exits 0; the no-arg / unknown-verb fallback still writes usage to stderr and exits 1.
- **Stale roster pointer** — `panel.yml`'s copilot `trial-ticket` refreshed to `tickets/closed/0206-...`. `scorecard`/`scores` resolve the trial ticket by 4-digit ID via `erg log <ID>`, so resolution never depended on the path — **only `list`'s display did**. Roster comment records this; seat retirement stays the author's 0205 call.
- **SKILL.md** — documents both verbs plus a "Candidate scouting" section (endpoint inventory via `~/.config/keys/*.env`, `GET <base>/v1/models`, website-only OpenRouter rankings, audition-vs-live-trial privacy asymmetry). Mechanics only.

## Review round 1 — two blockers fixed

1. **Greedy count parse** — VERIF/CONS now parsed from the FIRST keyword occurrence (parameter expansion), so freeform verdict prose ("flagged 3 as verifiable") no longer poisons the count that feeds 0205's promote/drop decision. Regression fixture: a verdict with decoy `9 verifiable` / `7 consider` prose still reports the true `1`/`2`.
2. **Whole-file scan** — `scores` now reads only the `--- log ---` section, not the body. A body-quoted card can no longer spawn a phantom row or WARN spam. Regression fixture: a `doc-example` card in a ticket body must not appear.

Also realigned the SKILL.md example table to the actual `printf` output (non-blocker).

## Report-only — ticket 0348 reopened

[...]

PR #636, opened 2026-07-15
feat(reviewers): audition/seat latency tracking + peer-relative SLOW gate

MinhHaDuong:
## Summary

Ticket 0353 (reopened under the 2026-07-15 raid directive, superseding the PR #632 cool-down close). Tracks per-review response time on both reviewer paths and eliminates a candidate that is slow *relative to its peers* on the same frozen benchmark board.

**Audition path** (`skills/reviewers/reviewers.sh`):
- Per-PR elapsed collected into an array; nearest-rank `latency-p50` / `latency-p95` appended after `cost=`. The existing `latency=` field stays a running **SUM** across board PRs — deliberately unchanged (it is not the "mean" the ticket prose implied).
- **Peer-relative SLOW gate**: this candidate's p50 is compared against OTHER candidates that replayed the *same board size*. Flagged with a bare trailing ` SLOW` token (verdict → eliminate-slow, no advisory trial) when `p50 > REVIEWERS_SLOW_FACTOR × cross-candidate median p50` (default factor 3, STRICT `>`, self included in the median sample). Fires only with ≥1 peer; never edits a prior card (append-only); never touches the roster.

**Request path**: each `cli-agent`/`local-model` seat is timed (whatever the outcome) into a `<seat>.latency` sidecar; `forge-bot` seats run async server-side and get no sidecar. `scorecard` folds the sidecar seconds into the trial line at END — absent a sidecar the line is byte-identical to the old schema.

**scores**: `LATENCY` column now shows p50 (fallback to the sum for pre-0353 cards); new `P95` and `FLAG` (SLOW) columns; scorecard rows show per-seat latency. The scores/audition corpus log-scan is factored into a shared `_corpus_log_lines` helper, and `_percentile` serves both the per-run stats and the peer median.

**Tests**: real-time-independent via a `REVIEWERS_AUDITION_ELAPSED` injection seam. Hand-written peer cards (p50 10/10/35 → median 10 ≠ mean 18.3) prove median-not-mean; boundary (30 = 3×10) not flagged; lone candidate never flagged; factor override in both directions; scorecard sidecar byte-identity + validity. `make check` green (858 pytest items + agnostic/skills-drift/bash-suite gates).

## Adherence

Self-gated: `make check` (includes the adherence/agnostic gates) exits 0.

## Report-only notes (author's raid directive: NO follow-up tickets)

- **Board identity is inferred from PR count only.** `board=<N>MR` is the only board discriminator, so two different N-PR boards would be treated as peers. One board ships today, so this is inert — noted, not fixed.
- **Request-path latency has no live non-forge-bot seat to exercise it** until one is enrolled in `panel.yml`; the sidecar + scorecard fold are covered by tests, not by a live run.

**Ticket:** tickets/0353-audition-latency-tracking-relative-slow.erg

### DIFF
--- a/skills/reviewers/SKILL.md
+++ b/skills/reviewers/SKILL.md
@@ -43,6 +43,12 @@ diff (the branch is resolved from the PR, or passed explicitly):
 proceed — one seat never blocks the verdict (0205). Empty roster → no-op,
 exits 0. No secrets in config; seat credentials load via the BASH_ENV path.
 
+**Per-seat latency** (ticket 0353): each `cli-agent`/`local-model` seat is timed
+and its wall-clock seconds written to a `<seat>.latency` sidecar beside the
+findings, whatever the outcome (a slow-then-failed seat is still evidence).
+`scorecard` folds that figure into the seat's trial line. `forge-bot` seats run
+async server-side and get **no** sidecar — there is nothing local to time.
+
 ### `/reviewers harvest <pr>`
 
 Collect every seat's findings and normalize the seat-runner's
@@ -64,8 +70,14 @@ Append a fixed-schema trial line to the seat's trial ticket via `erg log`
 evidence-based:
 
 ```
-MR #42 seat=local-qwen verdict: PASS — 0 verifiable, 2 consider
+MR #42 seat=local-qwen verdict: PASS — 0 verifiable, 2 consider latency=48.7s
 ```
+
+When `request` left a `.latency` sidecar for the seat (a `cli-agent`/
+`local-model` seat — see below), its wall-clock seconds fold into the line as a
+trailing `latency=<s>s` field (ticket 0353). Absent a sidecar the line is
+byte-identical to the pre-latency schema — the field is appended at the end, so
+every existing parser is unaffected.
 
 ### `/reviewers scores [seat-or-candidate]`
 
@@ -84,18 +96,21 @@ card to an archived trial ticket fails "no ticket found". `scores` reads by
 grep, so it is not bound by that.)
 
 ```
-KIND      NAME                 MR/BOARD                   VERIF  CONS  FIND  DUP  UVER  UHAL  OVERLAP   LATENCY     COST
-audition  hy3-free             10MR                           -     -    59   23     0    36      38%   3419.1s      n/a
-scorecard copilot              ImperialDragonHarness#537      0     0     -    -     -     -        -         -        -
+KIND      NAME                 MR/BOARD                   VERIF  CONS  FIND  DUP  UVER  UHAL  OVERLAP   LATENCY      P95     COST  FLAG
+audition  hy3-free             10MR                           -     -    59   23     0    36      38%      9.4s    41.0s      n/a  SLOW
+scorecard copilot              ImperialDragonHarness#537      0     0     -    -     -     -        -     48.7s        -        -     -
 ```
 
 No argument → all seats and candidates. An argument filters to one seat or
 candidate name. A malformed trial line WARNs on stderr, never silently dropped
 (the harvest convention). Only each ticket's `--- log ---` section is read, so
 a card quoted in a ticket body as documentation is never mistaken for a result.
-`scorecard`'s per-MR columns (`VERIF`/`CONS`) and
-`audition`'s per-board columns (`FIND`/`DUP`/`UVER`/`UHAL`/`OVERLAP`/`LATENCY`/
-`COST`) share one table; a `-` marks a column that does not apply to that row.
+`scorecard`'s per-MR columns (`VERIF`/`CONS`) and `audition`'s per-board columns
+(`FIND`/`DUP`/`UVER`/`UHAL`/`OVERLAP`) share one table; a `-` marks a column that
+does not apply to that row. `LATENCY` shows a candidate's `latency-p50` (the
+gate-relevant stat, falling back to the running-sum `latency=` for older cards)
+and a seat's per-MR latency; `P95` is the candidate's tail latency; `FLAG`
+surfaces the peer-relative `SLOW` verdict (ticket 0353).
 
 ### `/reviewers help`
 
@@ -132,7 +147,8 @@ trial ticket via `erg log` (verb `note`):
 ```
 audition candidate=<label> model=<id> board=<N>MR findings=<F> \
   duplicate=<d> unique-verified=<uv> unique-hallucinated=<uh> \
-  overlap=<pct>% latency=<s>s cost=<$x|n/a>
+  overlap=<pct>% latency=<s>s cost=<$x|n/a> \
+  latency-p50=<s>s latency-p95=<s>s [SLOW]
 ```
 
 `overlap%` is the share of the candidate's findings that merely duplicate the
@@ -140,6 +156,22 @@ panel. `cost` is `$ per review` from the token counts the seat reports on its
 `SUMMARY` line, priced via `REVIEWERS_PRICE_IN_PER_M` / `REVIEWERS_PRICE_OUT_PER_M`
 (USD per 1M tokens); it is `n/a` when the seat reports no tokens or no price is
 configured — an honest blank, never a fabricated `$0`.
+
+`latency=` is a running **sum** of wall-clock across the board PRs.
+`latency-p50` / `latency-p95` are the per-PR latency distribution (nearest-rank,
+appended after `cost=` so existing parsers are unaffected; ticket 0353).
+
+**Peer-relative SLOW gate** (ticket 0353): a candidate whose `latency-p50`
+exceeds `REVIEWERS_SLOW_FACTOR` × the cross-candidate median p50 (default factor
+**3**) earns a bare trailing ` SLOW` token, and its verdict becomes
+**eliminate-slow** — it does not proceed to the live advisory trial. The
+comparison is against **other candidates that replayed the same board size**:
+identical work, so the statistic is host- and diff-size-independent. It fires
+only with **≥1 peer** (a lone candidate has no basis for comparison) and is a
+**strict** `>`, so a candidate at exactly `factor × median` is kept. The gate
+only reads peers' logged cards and appends to the new candidate's own card — it
+**never edits a prior candidate's line** (append-only) and never touches the
+roster; roster promotion stays the manual call at 0205's integration review.
 
 **Fail-loud**: unlike `request`'s per-seat fail-open, audition aborts non-zero
 if the seat-runner cannot replay a board PR (unreachable endpoint, sandbox
@@ -162,6 +194,9 @@ tickets/NNNN-...` (where the scorecard is logged; default the 0207 trial ticket)
 `--credential-env NAME` (for an authenticated endpoint; threaded to the
 seat-runner, never written to config), `--name LABEL` (candidate label in the
 scorecard; default the model id).
+
+Environment: `REVIEWERS_SLOW_FACTOR` (default `3`) sets the SLOW threshold as a
+multiple of the cross-candidate median p50.
 
 ## Candidate scouting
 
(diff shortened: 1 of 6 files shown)


MinhHaDuong (commented):
## /review-pr — 5-agent panel (Correctness · Consistency · Scope · Red team · Doc propagation)

`make check` green on the branch (858 pytest passed, 60s) + `test_reviewers.sh` 55/55 + `test_reviewers_audition.sh` 28/28. The latency capture, `_percentile` nearest-rank math, the strict-`>` peer-relative SLOW gate (self-included median, ≥1-peer, boundary-not-flagged), and the request-path sidecar byte-identity are all correct and well-tested — the median-not-mean fixture (10/10/35) genuinely discriminates the two implementations. Scope maps cleanly to the five ticket Actions with no creep; the `_corpus_log_lines` extraction is justified by a real second caller. Docs, table columns (14/14), and the 0348 dependency all check out.

**Panel split:** Consistency / Scope / Doc → approve. Correctness → comment. Red team → request-changes. The split is over one convergent finding, below.

---

### Blocker (Correctness + Red team, high confidence) — `--name`/`model` values with a space or `=` silently corrupt the card and break self-exclusion

`audition` validates `model`/`label` against newlines/CRs only (`skills/reviewers/reviewers.sh:386-387`), but the trial card is space-delimited `key=value` and `_card_field` is first-match to the next space (`:129-132`). A space or `=` in `candidate=${label}` — the first field on the line — truncates or hijacks every later reader.

Reproduced against the shipped `_card_field`:
```
label="Local Qwen 30B"  →  _card_field candidate  ⇒  "Local"   (truncated)
label="x latency-p50=0.1s"  →  _card_field latency-p50  ⇒  "0.1s"  (hijacked; real value lost)
```
`--name` is documented as a free-text "candidate label," so `"Local Qwen 30B"` is ordinary operator input, not only an adversarial one. Consequences introduced by this PR's new `_audition_peer_p50s` self-exclusion (`:181-182`): the candidate's own prior card is never recognized as self (`"Local" != "Local Qwen 30B"`), so on a re-audition it leaks into its own peer median; and the `scores` NAME column truncates. The existing guard just extended the validation net for newlines but missed the two characters the card format actually depends on.

**verifiable:** `skills/reviewers/reviewers.sh:129-132,181-182,386-387` (SHA b5c347f), repro above.
**Fix (one line):** extend the existing `case` guards to also reject spaces and `=` (or restrict `label`/`model` to a slug charset). Cheap, and it closes the exact tamper-/typo-surface the peer-relative gate relies on.

*Note on framing:* `--name` is operator input, not an untrusted boundary — the red-team "adversarial poisoning of the peer median" scenario overstates the threat model (the operator running an audition is not attacking their own gate). The load-bearing reason to fix is the plain-typo correctness path (a labelled candidate silently mis-scored), not security.

---

### Non-blockers

**consider:** (Correctness/Red team) self-exclusion keys on the exact label string, and `label` defaults to the raw `model` when `--name` is omitted (`:382`). Re-auditing the same model against two endpoints without distinct `--name`s makes each run exclude the other from its peer sample, silently thinning the median. Either key self-match on `label+model`, or document that `--name` must be unique per distinct candidate.

**consider:** (Correctness) with exactly one peer (N=2 total), nearest-rank "median" returns the lower value, so the gate is more aggressive than the word "median" implies at the ticket's stated `≥1 peer` boundary. All SLOW tests use 2 peers (N=3, odd); the 1-peer path is untested. Add a lone-peer case or note the nearest-rank-at-even-N behavior.

**consider:** (Red team) `_audition_peer_p50s` checks only that `latency-p50` is non-empty (`:180`); a non-numeric value coerces to `0.0` through `sort -g | awk` with no warning, dragging the median toward zero. Validate numeric before feeding `_percentile`.

**consider:** (Red team, low) `_percentile` uses `sort -g`/awk without `LC_ALL=C`. Not reproducible on the target French-locale system (glibc `sort -g`/gawk parsed `.`-decimals correctly), but a cheap `LC_ALL=C` prefix is defensive insurance.

**consider:** (Consistency) ticket 0353 Action 1 still says the new fields go "alongside the existing **mean**," but the shipped `latency=` is a running **sum** (code comment `:429-431` and SKILL.md agree). Stale planning prose — worth a one-line ticket note, not blocking; code/docs are internally consistent.

**nofollow:** FLAG (`*" SLOW"`) spoofing via label/model — ruled out; injected content can never land at the true end of the card (`latency-p95=…s${slow}` is always appended last). `date +%s.%N` macOS portability — out of scope (Linux-only harness). Integer-truncating `overlap=` — pre-existing, untouched.

---

**Verdict: request-changes** — one reproduced correctness defect (space/`=` in a documented free-text label breaks the peer-relative self-exclusion this PR introduces) with a one-line fix. Everything else is green and the design is sound; the three `consider:` gate-robustness items (numeric validation, label-identity keying, 1-peer coverage) are worth folding into the same fix commit but the author may defer them.

MinhHaDuong:
## /gaze review-pr — PR 636 (ticket 0353)

**Risk band:** +scripts (touches `skills/reviewers/reviewers.sh`, a script `/gaze`'s own external-reviewer-panel delegation depends on; extends a scorecard schema other code parses; updates `SKILL.md` docs). Ran correctness + consistency + doc-propagation perspectives in parallel.

**Suite:** `make check` exits 0 (858 pytest items + agnostic/skills-drift/bash-suite gates). `tests/test_reviewers.sh` 55/0, `tests/test_reviewers_audition.sh` 31/0.

### Exit-criteria spot-checks

- **Latency on both paths**: confirmed. `request` times every seat invocation outside the success/failure branch (`t0`/`t1`/`.latency` sidecar written unconditionally, keyed off `seat_ok`) — a slow-then-failed seat is still timed. `audition` collects per-PR elapsed into `lat_samples[]` and computes `_percentile 50/95` (nearest-rank, nothing off-by-one: verified `N=1` and `N=3` cases by hand against the awk `ceil` formula).
- **SLOW flag peer-relative, gates only, never touches the roster**: confirmed. `_audition_peer_p50s` reads only OTHER candidates' `latency-p50` on the same `${bsize}MR` board size from the ticket-corpus log sections (`_corpus_log_lines`, which naturally recurses into `tickets/closed/` — no `-maxdepth`); the gate is append-only to the new candidate's own card and never edits a prior card or any roster/panel file. Strict `>` verified via the awk `exit !(x > f*m)` idiom; boundary case (`30 = 3×10`) is test-confirmed NOT flagged.
- **`scores` shows latency stats and flags**: confirmed present, and confirmed layered on a real base — the `scores` verb from ticket 0348 (PR #634) is already on `origin/main`; this PR only extends its printf/`fmt` and column set (`LATENCY`/`P95`/`FLAG`), which was byte-matched against the `SKILL.md` example table by one of the review agents.
- **Backward compatibility**: confirmed. `_card_field` returns empty string on an absent field; `scores`'s required-field check doesn't include the new p50/p95 fields, so an old scorecard line WARNs-and-continues rather than crashing; `latshow="${p50:-$lat}"` falls back to the pre-existing running-sum field.
- **Board-identity limitation accurately reported**: the PR body's "Report-only notes" section states board identity is inferred from PR count only, and that this is inert while only one board ships. Per the author's raid directive this is explicitly not being fixed here — correctly represented as a known limitation, not silently wrong for the current single-board case.

### Findings

**consider:** `skills/reviewers/reviewers.sh:503` — the peer-relative median (`_percentile 50 "${peer_p50s[@]}" "$p50"`) includes the candidate's own p50 in the sample, not just its peers. This is deliberately commented ("self included in the median sample, so a mean-based bug on 10/10/35 is caught") and is self-consistent with the test fixtures, but it means self-inclusion always raises the threshold when the candidate is the outlier — it can only make the gate more lenient toward a slow candidate, never produce a false positive. The ticket body's "cross-candidate median p50" wording doesn't pin down whether self is included; both readings satisfy the exit criterion as worded. No assertion of incorrect behavior — this is a hypothesis about intended strictness, not a bug.

**consider:** `skills/reviewers/reviewers.sh:501` — peer samples are per audition *card*, not per candidate identity: a candidate re-run N times contributes N p50 samples to the peer pool, weighting the baseline toward frequently-rerun candidates. Same caveat as above — plausible either way, not specified by the ticket.

**nofollow:** `skills/reviewers/SKILL.md` documents the self-inclusion-in-median behavior only implicitly (via the gate description, not spelled out as "including its own p50"); a reader computing the threshold by hand from the doc alone could reconstruct a peer-only median and get a different number. Doesn't cause wrong tool usage since the tool itself is correct — noted, not pursued given the author's no-follow-up-tickets directive for this ticket.

**nofollow:** `REVIEWERS_TICKETS` (used inline at reviewers.sh:497 and :533) is not declared alongside the other `REVIEWERS_*` overrides in the top-of-file block (`REVIEWERS_REPO`/`REVIEWERS_PANEL`/`REVIEWERS_FINDINGS_DIR`/`REVIEWERS_BOARD`, lines 15-22) and isn't listed in `SKILL.md`'s "Environment" note alongside `REVIEWERS_SLOW_FACTOR`. Self-verified directly (not merely agent-reported): confirmed present at both call sites and absent from both locations. Cosmetic — the var works correctly either way — but is a real, unexplained inconsistency in an otherwise carefully-documented env-var surface.

One perspective flagged a possible ticket-log discrepancy (a reopen note citing "PR #632" with no earlier log entry mentioning that PR). Verified directly and it's not a discrepancy: `gh pr view 632` confirms PR #632 is the real, merged wontfix-close PR for this ticket ("chore(0348,0353): close wontfix — harness cool-down"); dropped from the findings.

### Verdict

**APPROVE.** All three perspectives (correctness, consistency, doc-propagation) independently converged on approve/comment with high confidence; no blockers. `make check` and both targeted test suites are green. The two `consider:` items are judgment calls on unspecified median semantics, not defects — worth a nod from the author but not merge-blocking, consistent with "NO follow-up tickets" for this ticket.

MinhHaDuong:
Exit criteria: 4/4  Review: 0 unresolved  Simplify: 0 unresolved
Adherence: 0 blocking  Scope overflow: 0 files
Minors: verifiable:0 consider:2 nofollow:2 malformed:0

Per-criterion evidence (tip ab541a6):
- Per-review latency, both paths — ADDRESSED. Audition: `_elapsed` (reviewers.sh:139), per-PR samples → `_percentile 50/95` (reviewers.sh:492-493), `latency-p50=/latency-p95=` on the card (reviewers.sh:517). Request: unconditional timing → `<seat>.latency` sidecar (reviewers.sh:303), folded by `scorecard` (reviewers.sh:370); sidecar-absent lines byte-identical (test-covered).
- SLOW flag peer-relative, gates trial only, never edits roster — ADDRESSED. `_audition_peer_p50s` same-board peers only (reviewers.sh:188,508), strict `>` factor×median (reviewers.sh:511), ≥1-peer guard, append-only card; grep confirms zero panel.yml writes; pipeline gate documented SKILL.md:164-173. Tests: slow flagged, boundary 30=3×10 kept, lone candidate never flagged, factor override both directions (test_reviewers_audition.sh:291-333).
- scores shows latency stats + flags — ADDRESSED. LATENCY/P95/FLAG columns (reviewers.sh:543,557-568), p50-with-sum-fallback for pre-0353 cards; byte-matched against SKILL.md example table by review-pr.
- SKILL.md documents gate; tests green; make check passes — ADDRESSED. SKILL.md:164-190 (factor, strict >, peer semantics, roster invariant); CI 9/9 SUCCESS on ab541a6; gate independently re-ran test_reviewers.sh (55/0) and test_reviewers_audition.sh (31/0) in a clean worktree at ab541a6, post-simplify.

Review comments: /review-pr APPROVE (2026-07-15T00:56Z), no blockers; 2 consider: (self-inclusion in median, per-card peer weighting — informational, lenient-direction only), 2 nofollow: (muted). Simplify findings applied as commit ab541a6; suites green after. Reopen of 0353 (6445c1c) is authorized by the author's 2026-07-15 raid directive (PR #634 precedent). Two report-only WARNs in the PR body (board identity by PR count; no live non-forge-bot seat) accepted per the same directive.

Rationale: strongest remaining attack is the unspecified median semantics (self-inclusion + per-card weighting) — both readings satisfy the ticket as worded, both bias only toward leniency (no false SLOW positive), and both are comment- and test-pinned; not merge-blocking. No follow-up tickets filed per author constraint.

PR #680, opened 2026-07-27
docs(update-publist): credentials come from the keystore, not .env

MinhHaDuong:
The HAL deposit step of `update-publist` said its credentials come "from the
project `.env`". That was true before the keystore migration. It is now false
in the worst direction: a reader who believes it edits `.env` — the one file
whose invariant is that it holds no credential value.

This names the real source (`~/.config/keys/hal.env`), the real mechanism
(a default-deny `KEYS=` selection), and the part that actually bites: the
loader reads the `.env` of the directory the shell *starts* in, so a run
launched from an unselected cwd fails as an ordinary auth error. `~/.claude/.env`
now carries `hal:HAL_ID,hal:HAL_PASSWORD` (machine-local, no value written), which
covers every startup directory that does not declare a `KEYS=` line of its own.
A directory whose `.env` *does* carry one replaces the harness selection wholesale —
measured, and tracked as ticket 0360 in #681. The original wording here said "a run
from any cwd resolves", which is false in exactly the case 0360 documents.

Verified by hash from three startup cwds (`$HOME`, the project root, a project
worktree): both names present and matching `~/.config/keys/hal.env`. No value
was printed at any point.

Found while fixing the same stale claim in the climate-finance project, which
turned out to carry a stale *fork* of this skill that shadowed this one — two
documents disagreeing on fact. The fork is deleted there in favour of this
document (project ticket 0364).

Ticket: none

### DIFF
--- a/skills/update-publist/SKILL.md
+++ b/skills/update-publist/SKILL.md
@@ -111,9 +111,18 @@ curl -s 'https://api.archives-ouvertes.fr/search/hal/?q=authIdHal_s:minh-ha-duon
 ```
 
 Standard values for this author:
-- Structure: `#struct-1380080` (CIRED)
+- Structure: `#struct-1380080` (CIRED). **Verified against the HAL ref API,
+  2026-07-27** — a fork of this skill carried `struct-1002424`, which
+  resolves to ECOSYS, an unrelated laboratory. Do not "correct" this id
+  from another document; re-verify it instead:
+  `curl -s 'https://api.archives-ouvertes.fr/ref/structure/?q=docid:1380080&fl=docid,name_s,acronym_s&wt=json'`
+- ORCID: `0000-0001-9988-2100`
 - idhal: `minh-ha-duong`
 - Typology: `COMM` for a conference talk (adjust per doctype)
+- Stamps: CIRED, CNRS
+- Domains: `shs.eco`, `shs.hisphilso` (adjust per paper)
+- File reference in the TEI:
+  `<ref type="file" target="paper.pdf" subtype="author" n="1"/>`
 
 For a journal article, resolve the journal's HAL id rather than typing
 the title into the TEI:
@@ -154,6 +163,15 @@ so it is deleted even on error or interruption. Contents:
 ```
 user = "HAL_ID_VALUE:HAL_PASSWORD_VALUE"
 ```
+
+**Dry run first.** The same request with `-H "X-test: 1"` validates the
+package without creating a record. Run it, read the response, and only
+then repeat without the header.
+
+**Updating an existing deposit** is the same request as a `PUT` to
+`https://api.archives-ouvertes.fr/sword/hal/{hal-id}`. Use it rather
+than a second POST — a second POST creates a duplicate record, and
+duplicates need moderator action to merge.
 
 The SWORD acknowledgement means "stored in workspace" — moderation
 status is confirmed by HAL's email, not the API response.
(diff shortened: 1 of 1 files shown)


MinhHaDuong:
## Two contradictions this PR deliberately does not resolve

Review of the companion project PR (climate-finance-het#1190) found that the deleted fork and this document disagree on two facts, beyond the credential-source line this PR fixes:

1. **Is `~/CNRS/html` a git repo?** This doc says "NOT a git repo. It deploys over FTP." The fork said "IS a git repo: commit directly on its main, one commit per logical change; `index.html` is gitignored."
2. **Where does the bib live?** This doc says `src/Ha-Duong.bib`. The fork said `~/CNRS/html/Ha-Duong.bib` (repo root, *not* `src/`) — and climate-finance-het's `docs/style-anchor-v205.md:270` independently records the same root path, i.e. it corroborates the *deleted* document.

`~/CNRS` does not exist on padme, so neither claim is checkable from here and I have left this doc's version standing rather than guess. Worth a look before the next deposit run: point 2 in particular was used as evidence that this document was the better one, and the independent corroboration points the other way.

MinhHaDuong:
## Correct — and the trigger is narrower than "a run from anywhere" being false

Confirmed, my defect. Before anyone rewords it, I measured the three cases, because the accurate sentence turns on a detail neither of us had checked — whether the override needs a `.env` or a `KEYS=` line:

| startup directory | harness `KEYS=` survives? |
|---|---|
| no `.env` at all | **yes** — `HAL_ID` set |
| `.env` present, no `KEYS=` line | **yes** — `HAL_ID` set |
| `.env` with a `KEYS=` line not naming `hal` | **no** — `HAL_ID` unset, **and `OPENAI_API_KEY` unset** |

Probe: `env -i HOME=$HOME BASH_ENV=~/.claude/scripts/bash-env.sh bash -c '…'` from each, reporting set/unset by name only.

Two things follow that the current draft of 0360 does not say sharply enough. The override fires on the **presence of a `KEYS=` line**, not on the presence of `.env` — so a project with a `.env` of plain settings keeps the full harness selection, and adding one `KEYS=` line to it silently drops every harness credential at once, not just the one being discussed. That is a worse failure than "a project cannot narrow": it is a cliff, and it is invisible until an unrelated tool fails to authenticate.

Suggested replacement for the Step 2 credentials paragraph — yours to take or rewrite, I am not touching the file since you have it:

> Credentials: `HAL_ID` and `HAL_PASSWORD` live in `~/.config/keys/hal.env` and reach the environment because a `KEYS=` line selects them — no `.env` holds a credential value. Selection is default-deny and does **not** compose: if the directory the shell starts in has an `.env` containing a `KEYS=` line, that line is the entire selection and must itself name `hal:HAL_ID,hal:HAL_PASSWORD`. Where the startup directory declares no `KEYS=` of its own, `~/.claude/.env` supplies it. Ticket 0360 tracks making these compose.

**Two more instances of the same false claim**, since it propagated further than the doc:

1. **This PR's body**, "a run from any cwd resolves" — I have corrected it in place.
2. **climate-finance-het `tickets/0364`, lines 61-63**, "so the skill authenticates from any cwd" — already merged to that repo's main, and it contradicts line 55 of the same ticket, which states the override correctly. That is in my lane; PR incoming there. Flagging it so you do not spend a round on it.

Commit `b6756fe`'s message carries the claim too. Immutable, and superseded by the text above rather than worth a rewrite.

PR #692, opened 2026-07-28
skills(0376): give hunt and gaze a test-run budget

MinhHaDuong:
**Ticket:** tickets/0376-hunt-executors-re-run-the-full-test-suit.erg

## Why

The 2026-07-28 trace analysis of 22 hunt/raid kills measured full `make check` re-runs as the largest active-time bucket: 348 of 1,250 agent-minutes (27.8%), median 18 full-suite runs per kill. The hunt contract already prescribed `check-fast` during development, but step 8 said "implement until `make check` passes" and the review-fix loop (step 12) set no budget at all.

## What

- hunt step 8: `make check-fast` (or the affected test file alone) is the loop gate.
- hunt step 9 mechanical-fix bullet: same swap.
- hunt step 10: the one mandatory full `make check`, moved into the PR-opening step so it reads as a gate, not a loop exit.
- hunt steps 12–13: per-cycle `check-fast` plus the implicated tests; one additional full check only when a fix leaves the fast tier.
- gaze Fix-agent contract: mirrors the same budget.
- New guard `tests/test_hunt_test_budget.py`: negative guard on the old unbounded step-8 wording, loose positive markers scoped to each step's text block (prose-test polarity rule).

Rejected alternative: a numeric cap ("at most 3 full runs") — a failing diagnostic run is never waste, and a cap invites gaming the counter; the budget names *which gate applies where*, which is checkable from the text.

## Verification

- `tests/test_hunt_test_budget.py`, `tests/test_hunt_invokes_review_pr.py`, `tests/test_skill_descriptions.py`: 9/9 green.
- Full `make check`: 872 passed; 4 local failures shown pre-existing/environmental (2 rtk output-rewrite artifacts that pass under `rtk proxy`, 2 bash suites failing identically on a worktree without this change — "KEYS provider not found", local env only).

## Notes

- Wave 1 of raid 376/377; ticket 0377 (round scoping) lands next and rewrites hunt steps 12–13 on top of this wording — sequential by design.
- Report-only finding (severity floor): `review-pr:88` and `gaze:260` run a full `make check` per synthesis invocation; candidate for the same budget treatment, not ticketed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #694, opened 2026-07-28
skills(0378): hunt step 2b — triage needs-human, else detach execution

MinhHaDuong:
**Ticket:** tickets/0378-hunt-pre-step-triage-needs-human-else-de.erg
Ticket-ref: tickets/0390-screen-needs-human-before-autonomous-ex.erg

## What this does

Adds step 2b to `skills/hunt/SKILL.md`, between the exit-criteria read and the
worktree step, so a hunt routes a ticket before any code is written:

- **Needs-human triage** — `Label: needs-human`, decision verbs in the exit
  criteria, manuscript prose as the deliverable, or a premise resting on
  conflicting sources. On a hit the hunt writes no code and returns ONE batched
  decision list, which the contract states is a success outcome, not a failure.
- **Detach by default** — otherwise the interactive hunt hands the ticket to ONE
  background worktree-isolated agent whose first action is the mechanical
  `Skill(skill: "hunt", args: "<id>")`, contract unparaphrased. One agent, not a
  raid: a raid for a single ticket buys the orchestrator and review-panel
  overhead the trace analysis measured as the top cost.
- **Inline override** — an explicit `inline` argument keeps the old co-working
  behaviour in the author's own session.
- **Recursion guard** — step 3's worktree ownership check is the sole
  discriminator, forward-referenced rather than restated, so the harness keeps
  one definition of ownership instead of two that can drift.

Steps 3–13 are not renumbered; the literal `2b.` keeps every existing
cross-reference to a numbered hunt step valid.

The motivation is measured, from the 2026-07-28 trace analysis of 22
climate-finance-het kills: an interactive hunt accumulates the executor's test
dumps, diffs, and review rounds in the author's own session — 404k output tokens
and a 68-minute post-PR tail per kill, against 74k and 13 minutes for the same
contract run detached (5.4x).

## Test

`tests/test_hunt_triage_prestep.py`, 8 assertions, fast tier. Every assertion is
scoped to step 2b's own text by a bounded region extraction anchored on step 3
(`re.search(r"^2b\.\s.*?(?=^3\.\s)", ...)`), mirroring
`tests/test_hunt_worktree_ownership.py`. No file-wide substring checks: those
would pass on a 2b that landed in the wrong place or in the wrong shape, which is
the tautological-test antipattern. Includes a negative guard asserting 2b does
*not* restate step 3's ownership predicate, and an ordering test pinning
2 < 2b < 3.

Proven RED on the pre-fix state: checked out `origin/main`'s `SKILL.md` under the
new test file and got 8/8 failures, all on the region extraction rather than on
an incidental string.

## Gate results

- `make check` — **885 passed, 1 skipped**, exit 0
- `make lint` — **21 passed**, exit 0
- `erg check tickets/` — PASS, 359 tickets
- `/verify-adherence t0378-hunt-triage-step` — **PASS**, `mechanical_failures: []`,
  `semantic_findings: []`. `/gaze` may skip the mechanical phase.

Two `make check` failures seen on the first run were environmental, not from this
branch: `test_reviewers_audition.sh` is a French-locale decimal-comma artifact
(`latency=70,0s` vs the expected `70.0s`) and passes 31/31 under `LC_ALL=C`;
`test_guard_cd_primary_repo.sh` fails on stderr polluted by a missing local
`openrouter` credential file and passes with `BASH_ENV` unset. Both reproduce on
a clean `origin/main`. The counts above are from the clean-env run.

## Adherence finding, and why the design stands

The first adherence pass returned one blocking semantic finding: step 2b's three
skips are total, so a `Label: needs-human` ticket picked up by a raid Phase 5
executor executes with no triage at all — and nothing upstream screens the label
(raid never mentions it, and pick-ticket's only exclusion is the manual
`.git/beat-skip.json` opt-in).

The design was kept rather than split, because ticket 0378 authorizes exactly
this behaviour in as many words. Its Verification bullet 3 reads "a raid Phase 5
executor invoking `Skill(hunt)` proceeds straight to execution (no triage, no
recursive spawn)"; bullet 2's unqualified "returns a decision list" is the
general case bullet 3 excepts. Making an executor re-triage would also change
what a raid does mid-run on a ticket it had already picked — a contract change
with a blast radius into raid, which is the author's call rather than a pre-PR
fix.

What the finding did expose is that the limit was implicit — a fresh reader of 2b
concluded needs-human was screened everywhere, which is why it came back blocking
rather than as a note. So 2b now states that each skip is total, says why an
interactive hunt's executor needs no second triage, and points at where a raid's
targets should have been screened instead: **new ticket 0390**, blocked by 0378,
which excludes `Label: needs-human` in pick-ticket and states it in raid Phase 1.
Screening at the picker is cheaper than re-triaging in every executor and leaves
0378's recursion guard alone. 0390 is referenced, not closed, by this PR.

The re-run after that change returned PASS with no findings.

## Ticket Action 3 — sweep result, no-op

Action 3 asked for a sweep of the raid/beat launch prompts for contradictions
with the new step. Result recorded in the ticket log: **no contradictions.**
`beat.py` runs `/raid`, never `/hunt`; raid Phase 5 spawns
`Agent(isolation:"worktree")`, which produces a clean `agent-*` worktree that
passes step 3's ownership check, so the triage stays off for it as intended;
gaze and verify-adherence reference only hunt step 9's adherence label. One gap
was found and closed in passing: the ad hoc headless `claude -p "/hunt <id>"`
pattern is now an explicit third skip case in 2b.

## Exit criterion 3 is post-merge

Exit criterion 3 — "one real interactive hunt observed taking the detached path
(transcript or session log referenced in this ticket's log)" — is left
**unchecked deliberately**. It is observational and cannot be satisfied by this
execution, which ran as a detached raid executor, not an interactive session. It
becomes verifiable only once the step is on `main` and the author runs an
interactive `/hunt`. The ticket log records this.

## Notes for review

- This branch resumes a killed executor's salvaged WIP (commit `6d37e53`, already
  pushed under that message, so its description lands here rather than as a
  reword). The salvage was missing the author field on its two log notes, which
  `erg check` accepts but `tickets/AGENTS.md` specifies; restored in `059dfca`.
- Sibling ticket 0376 (PR #692) landed a test-run budget in the same file while
  this branch was in flight. Integrated by merge rather than rebase per
  `rules/git.md`'s same-file multi-PR-wave rule, and the union was grep-verified
  for both siblings' markers — a clean auto-merge exit is not by itself proof
  both survived. 0376's edits sit in steps 8–12, this one's in 2b; no overlap.


MinhHaDuong:
## /gaze actions

round: 2
tier: full
adherence: PASS — 0 blocking (guard tests 8/8, adherence suite 21/21, erg check 359 tickets)
review: 4 findings returned in-band (2 minor, 2 nit — B4 dismissed as factually wrong: `[[NNNN]]` has precedent in closed/0170)
review-pr: 3 tagged minors returned in-band (2 verifiable, 1 consider); external seats: copilot 3 findings, 2 OpenRouter seats failed fail-open
simplify: 1 commit `1e2892c` — 5 needless regexes → plain `in`, `model:` assertion pinned, spelling, 0390's false pick-ticket premise corrected (erg ready already skip-lists needs-human; real gap is raid Phase 1's direct read); 3 skips validated (raid-Phase-5 wording duplication is mandated by 0378 Action 2; position test needs uncollapsed text; 4-skill needs-human restatement is beyond this PR)
fix agent: 3 commits (`5c7e5e5` workflow.md /hunt row neutralized + carve-out, `f76ab67` inline leading-token parsing rule + RED-first guard test, `ae7ab8c` skip-count clause)

/gaze: slow run — subagent tokens ~956k > 500k warn (full tier, 2 rounds); continued per threshold policy.

## /verify-gate verdict

verdict: APPROVED

Exit criteria:
- Step 2b landed with triage tells, detach default, inline override, recursion guard: ADDRESSED — skills/hunt/SKILL.md 2b block (6d37e53, refined bdc0a42/1e2892c/f76ab67/ae7ab8c)
- Skill-text guard test in place and green: ADDRESSED — tests/test_hunt_triage_prestep.py, 9/9 on tip ae7ab8c
- One real interactive hunt observed taking the detached path: MISSING, validated rationale — post-merge observational (ticket log 09:21Z); ticket closes at merge per convention, item to be logged when observed

Unresolved review comments: none — round-1 blockers C2 (workflow.md /hunt row pre-empted 2b's triage via step 3's ownership misread) and B2 (unspecified inline syntax) verified fixed in the diff; C1/P1–P3 fixed in 1e2892c; B1 fixed in ae7ab8c; C3 consider-informational
Unresolved simplify: none
Adherence: PASS

Rationale:
Round 2 re-verified every fix-round claim against actual diffs: exactly 3 commits touching only the 3 mandated files, the workflow.md misclassification path closed (not annotated), the parsing-rule test genuinely RED-first and region-bounded, fast tier 789/789, no contradiction introduced. Scope overflow: none — all files trace to exit criteria or the round-1 fix list. Merge decision belongs to the caller.

telemetry: tier=full wall~=1420s agents=10 tokens~=956k cost~=na

PR #696, opened 2026-07-28
skills(0377): scope review rounds — re-run only objecting perspectives

MinhHaDuong:
**Ticket:** tickets/0377-review-rounds-re-run-the-full-perspectiv.erg

Builds on merged PR #692 (ticket 0376, test-run budget) — hunt step 12's budget wording is left untouched; this PR only edits steps 11 and 13.

## Why

The 2026-07-28 trace analysis of three climate-finance-het raids put review panels at 44–56% of all subagent output tokens, the largest cost bucket, ahead of gaze/gates (10–22%) and execution itself (20–30%). PR 1172 alone drew 23 verification agents against 1 execute agent.

The spend is not empty ritual — post-review fix edits appeared in 17 of 19 kills that reached review — but round pricing was flat: round N re-ran every perspective even when round N−1 gave most of them nothing to say.

## What

**`skills/review-pr/SKILL.md`** — new § Round scoping. Round 1 is unchanged (full proportional panel). Round N>1 re-runs only the perspectives whose previous verdict was comment or request-changes, plus **one** regression agent covering the whole cleared set — one agent, not one per cleared perspective, since it is a sanity pass rather than a re-review. The skill derives the round by counting its own posted reviews on the merge request; no caller passes a round number, because a caller that miscounts would silently under-review while the PR's own history cannot be miscounted.

**`skills/gaze/SKILL.md`** — the #562 clause sent any round ≥ 2 to the full battery unconditionally. That over-corrected. #562's real lesson was that a round-2 battery was needed on an unchanged diff, and that is satisfied by keeping Agents A and B, phase 5, and the phase 6 gate at full strength every round; only Agent C is scoped. The reset condition is a substantial rewrite (new files, or >~half the lines changed since the last review), **not** an unchanged diff — an unchanged diff is precisely the case where the earlier clearances still hold.

**`skills/review-pr-prose/SKILL.md`** — gaze's Agent C is `/review-pr` *or* `/review-pr-prose`, so the prose sibling gains the parallel rule. The AI-tells auditor is exempt from scoping: it scans the full text, so revised prose is new surface for it whatever the last round's verdicts were.

**`skills/hunt/SKILL.md`** — steps 11/13 name the loop iteration as a round and state that no round number is passed.

Three gaps surfaced by the pre-PR adherence gate, all the same failure family — the rule stated where the actor could not read it:

[...]

PR #698, opened 2026-07-28
ticket(0390): screen Label: needs-human before autonomous execution

MinhHaDuong:
**Ticket:** tickets/0390-screen-needs-human-before-autonomous-ex.erg

## What

Hunt step 2b (ticket 0378) triages a ticket before any code is written, but 0378's own Verification bullet 3 scopes that triage to skip for an already-detached executor. A raid picks its own tickets and spawns its own Phase 5 executors, so no interactive hunt ever runs the triage for them — leaving the autonomous path, where the author is not watching, as the one path unscreened.

Two selection surfaces, two different fixes:

- **`skills/pick-ticket/SKILL.md`** is ALREADY screened. Every label in `tickets/.ergrc` suppresses its ticket from `erg ready` output, so a `needs-human` ticket never reaches the candidate set. The gaze round on 0378 corrected the ticket's original premise here. This PR therefore **names the existing mechanism** in the candidate-selection prose rather than adding a second, hand-rolled label check. A duplicate screen is a definition to keep in sync with `.ergrc`, and the drifted copy is the one that runs.
- **`skills/raid/SKILL.md`** Phase 1 reads `tickets/` directly, never through `erg ready`, so it must state the exclusion itself: never a raid target, dropped before wave grouping, surfaced in the wrap-up briefing as a success outcome — the same standing 0378 gives a returned batched decision list. Wrap up step 2 itemizes the excluded tickets so that routing promise has a landing spot.
- **`skills/hunt/SKILL.md`** gets a two-line wording fix only: its clause said screening raid's targets "is upstream work — see [[0390]]", which this PR makes false. Step 2b's logic, its recursion guard, and 0378's bullet-3 scoping are untouched; `tests/test_hunt_triage_prestep.py`'s 9 guards stay green.

**Action 3 decided (author approved, 2026-07-28):** a raid whose queue holds only skip-labelled tickets returns the batched decision lists, not an empty-run report. The alternative considered was reporting an idle run; rejected because it discards the one artifact such a run actually produces — a single question round the author can answer in one pass.

## Review rounds folded in

Round 1 (4 verifiable findings, all resolved in `e5dc0e5`):

- **Explicit-ID carve-out** (red-team `request-changes`). Phase 1 also parses comma-separated IDs, so an unconditional "never a raid target" silently dropped a ticket named by hand — revoking this ticket's own invariant. The exclusion is now scoped to target *discovery*.
- **The exclusion is the label set, not one member.** `.ergrc` skip-lists `deferred` too, and Phase 1's rationale applies identically. It now reads the set from `.ergrc`.
- **"Run report" named nothing that exists** in raid's vocabulary. Renamed to the wrap-up briefing.
- **Stale hunt forward reference** unstaled (see above).

Round 2 (`2197938`):

- **"Explicit means the author asked" is false for the only caller that exists.** `scripts/beat.py` builds `/raid <id>` unattended and is the only programmatic `/raid` call site. The carve-out survives, but its rationale is now accurate: the autonomous path stays screened by `pick-ticket` upstream of beat, and that safety lives in beat's pick step, not here.
- **`.ergrc` fails asymmetrically** — an absent file falls back to defaults, an empty `[labels]` section fails open. Phase 1 now tells a hand-parse to fall back explicitly.
- **Both new guards were polarity-blind.** The reviewer demonstrated mutations that passed on prose stating the logical *negation* of the guarded invariant. The carve-out test now captures and asserts the verb, and the pick-ticket test asserts the delegation is stated positively before checking command shapes are absent. This raised the mutation cost; it did **not** make the guards inversion-proof — see the round-3 retraction below.

Round 3 (`cfd0b12`):

- **Retraction: the guards are not inversion-proof.** Round 2 claimed they were "fixed by pinning meaning rather than vocabulary." That was wrong. Review round 3 reproduced three mutations that keep all 8 guards green: appending a retraction sentence after the pinned verb, appending a genuine hand-rolled screen *after* the required delegation phrases, and prefixing the exclusion with "It is a common misreading that". What these guards actually do is catch **deletion and vocabulary drift** — a later editor removing the carve-out, or renaming the briefing back to "run report". What they do not catch is an editor who leaves the sentence standing and contradicts it two sentences later. Not fixed with a fourth regex round, because that is the finding: regex over prose cannot be made polarity-proof. Recorded as a correcting note in the ticket log.
- **`.ergrc` fail-open documented in `pick-ticket` too** — round 2 put it only in raid Phase 1, but pick-ticket's filter is exactly what round 2's safety rationale for the carve-out rests on.
- Two findings recorded in the ticket log rather than fixed: `erg`'s label match is case-sensitive on both key and value, and no re-screen runs between Phase 1 and Phase 5, so a ticket labelled mid-raid is caught nowhere. Both are real; neither is introduced here.

## Tests

`tests/test_needs_human_screen.py` — **8 guards** using **bounded-region extraction** (pick-ticket step 1 anchored on step 2; raid Phase 1 anchored on the Phase 2 heading; raid Wrap up anchored on Circuit breakers), per `tests/test_hunt_triage_prestep.py`. Not file-wide substring matches: `skills/pick-ticket/SKILL.md` already contained the string `needs-human` before this branch, in the *beat-skip* step, so a file-wide check would have passed on the unfixed tree.

**RED proof.** The whole 8-guard suite is RED against `origin/main` (8/8, reproduced by review round 3). The original 5 guards: 4 of 5 fail against `origin/main`, the fifth being a negative guard that is green on both sides by design — separately proven a ratchet by injecting a `grep -L "^Label: needs-human"` line, which turns it RED.

*Correction to an earlier version of this body*: the "5 of 6 assertions fail" figure was attributed to round 2. It belongs to the **review-round-1** commit (`e5dc0e5`) — those five assertions (`deferred`, `briefing`, `/raid <id>`, `discovery`, wrap-up `skip-labelled`) were introduced there, not in round 2. Measured against the round-1 commit, only 1 of 6 round-2 assertions flips, because `pick-ticket`'s region was written once in `ab96f97` and never re-edited.

## Gates (at `cfd0b12`)

- `make check-fast` — **797 passed, 104 deselected**
- `make lint` — **27 passed, 874 deselected**
- `/verify-adherence` — **PASS**, no mechanical failures

Per the author's 2026-07-28 gate directive the pre-PR gate is `make check-fast` + `make lint` only; the ticket's `make check` exit criterion is satisfied by the fast gate, noted in the ticket log.

Review round 3 ran the full `make check` anyway and reports **899 passed, 1 failed**: `test_guard_cd_primary_repo.sh`, which asserts empty stderr while the ambient `BASH_ENV` emits a missing-provider warning. Environmental, pre-existing, and on a surface this PR's diff does not touch — but it will bite any full-gate run on this machine.

## Deviation from the launch brief

The brief pinned `skills/hunt/SKILL.md` as unchanged, on two grounds: 0378's bullet-3 scoping must stand, and PR #696 was in flight against that file. The first still holds and is preserved (wording only, guards green). The second expired — #696 merged as `475333f` without touching those lines. Two independent review passes flagged the stale reference, and the ticket's own "Relevant files" section anticipates the edit, so it is folded in here rather than deferred to a ticket that would exist only to fix a sentence.

Related follow-up: tickets/0391-re-screen-skip-labels-at-raid-phase-5-la.erg (bundled — mid-raid re-screen window)
Cross-repo: MinhHaDuong/git-erg#325 (case-sensitive label match)


MinhHaDuong (commented):
## /review-pr — five-agent panel (correctness · consistency · scope · red team · doc propagation)

**Verdict: comment.** No blocker. CI 9/9 green. The core claim is verified true by execution, not by reading, and the RED proof reproduces exactly. Six findings below, four of them `verifiable:`.

### Verified positives (recorded so they are not re-litigated)

- **[correctness] The central premise is TRUE.** Confirmed three ways: `tickets/.ergrc` `[labels]` lists `needs-human`; `tickets/erg ready --help` documents the skip-list; and an empirical two-ticket fixture store returned only the unlabelled ticket. Action 1's "name the mechanism, do not duplicate it" therefore rests on a real filter.
- **[correctness] The gap is real.** `skills/raid/SKILL.md` Phase 1 reads *"read open tickets from git-erg `tickets/` or forge"* — no `erg ready`, no `pick-ticket`, on `origin/main` or the branch.
- **[correctness] RED proof reproduces.** Branch checkout: 5 passed. Same test file against an `origin/main` checkout: **4 failed, 1 passed**. Anchors match the actual branch text; no drift.
- **[correctness] Coverage sweep clean.** `beat` delegates to pick-ticket→raid; `molt`/`scry`/`lair` do not select-and-spawn; `nightbeat-supervisor` already excludes `needs-human`/`deferred` in its own Invariant 1. No second unscreened surface.
- **[consistency] Style clean.** `ruff check` passes; `make check-skills-drift` → in sync; no frontmatter touched. Em-dashes per added paragraph: 1 / 0 / 0 — `rules/prose/_all.md` holds. The idiom matches the declared model `tests/test_hunt_triage_prestep.py` exactly.
- **[scope] Every added line maps to an Action.** pick-ticket +6 → Action 1; raid +8 → Action 2; raid +5 → Action 3; tests → Test section; 2 ticket log lines are decision records, not artifacts. `skills/hunt/SKILL.md` untouched — Invariant 1 holds.

### Findings

**1. `verifiable:` raid's exclusion has no carve-out for an explicitly named ticket ID — and the panel split on whether that breaks an invariant.**
`skills/raid/SKILL.md:69` — *"Otherwise: parse comma-separated ticket IDs."* `:83-86` — *"**`Label: needs-human` is never a raid target.** … Drop it from the queue before wave grouping."* Unconditional; grep for `explicit|override|force` in Phase 1 returns nothing. So `/raid 0123` on a `needs-human` ticket the author named by hand is silently dropped.

Dissent preserved verbatim. **Red team (request-changes, high confidence):** *"the merged prose as written revokes an invariant the ticket explicitly promises, with no textual escape hatch."* **Synthesis reading:** ticket 0390's invariant *"The author can still force execution of a `needs-human` ticket explicitly"* is satisfied elsewhere — `skills/hunt/SKILL.md` step 2b skips the whole triage for *"a hunt the author invoked with an explicit `inline` argument"*. The force path exists; raid is simply silent about its own explicit-ID case. That downgrades this from a contradiction to a gap, which is why the panel verdict is `comment` and not `request-changes`. The author decides which reading governs. A one-clause fix closes it either way: *"…never an automatically picked raid target; an ID named directly in `$ARGUMENTS` is the author's explicit call."*

**2. `verifiable:` `deferred` is skip-listed in `.ergrc` but stays a raid target.**
`tickets/.ergrc` `[labels]`: *"All labels suppress `erg ready` output"* — then `needs-human` **and** `deferred`. `pick-ticket`'s new prose is correctly generic (defers to `.ergrc`). raid Phase 1 hardcodes the single literal `Label: needs-human`; `grep -c deferred skills/raid/SKILL.md` → 1 hit, and it is *"deferred post-merge steps"* at `:297`, unrelated. The PR's own rationale — Phase 1 bypasses the tool-level filter, so it must state the exclusion itself — generalizes to both labels, but the fix covers one. Either exclude both, or name in the prose why `deferred` is deliberately still raidable.

**3. `verifiable:` "run report" is not raid's vocabulary, and the promise has no landing spot.**
`grep -n "run report\|briefing" skills/raid/SKILL.md`: `run report` appears only at `:86` and `:93` — both added by this PR. Raid's own final artifact is the **briefing** (`:297`, `:329` — *"Write briefing (session log + merge request list + test delta)"*). A reader following Phase 1's pointer finds no "run report" section. Worse, Wrap up step 2 itemizes merged PRs and ESCALATED PRs but says nothing about excluded tickets — contrast the existing parked-PR case at `:296-298`, which explicitly instructs *"make the briefing name each parked PR and its deferred post-merge steps"* at the point the briefing is written. Suggest: rename to "briefing", and add the excluded tickets to Wrap up step 2's itemization.

**4. `verifiable:` the declined stale reference — both stated reasons have now expired, and no follow-up is filed.**
`skills/hunt/SKILL.md:44-46`: *"screening raid's own target selection is upstream work — see [[0390]] — not a second triage here."* The PR body declines the fix citing (a) the Invariant pinning hunt unchanged and (b) PR #696 in flight on that file. #696 merged as `475333f`, and its diff touched `skills/hunt/SKILL.md:121-127` only — never lines 41-46, so there was no real conflict. `ls tickets/*.erg` shows no follow-up (open: 0205, 0207, 0356, 0375, 0390). Ticket 0390's own Relevant-files section asks for it: *"its clause naming this upstream screen should stop pointing at an unfiled gap once this lands."* Updating a prose pointer is not reworking triage logic, so reading the Invariant as forbidding it is a stretch. Fold in the one-liner, or file the follow-up before merge so the promise is not carried only in a PR description.

**5. `consider:` the negative guard's forbidden-token list has a false-positive class.**
`test_pick_ticket_does_not_hand_roll_the_screen` bans `\b(grep|rg|awk|sed|jq)\b` and the phrases `Label:` / `label header` / `read the label` anywhere in step 1. Red team constructed a tripping edit that adds no second screen at all: *"do not add a second label check here — conceptually the same as a grep on the ticket's Label header, just owned by erg instead."* Pure documentation, red guard. Scope also flags `jq` specifically: step 1 already calls `erg ready --json` with no consumer, so `| jq '.id'` is a plausible future edit for reasons unrelated to labels. Not a defect today — the assertion messages are self-diagnosing — but the maintenance cost is real. Two corrections worth recording: the guard's own `\brg\b` comment is **right** (`"rg " in "erg ready"` is True; `re.search(r"\brg\b", "erg ready")` is None), and `\bsed\b` does not match `used`/`based`/`parsed`.

**6. `consider:` region-extraction fragility fails loudly but misleadingly.**
Renumbering pick-ticket's steps, or renaming `## Phase 1: Select`, makes 2-4 tests fail with a generic `could not locate…` assertion rather than a content error — a maintainer doing an unrelated rename reads it as a broken screen. Also a prefix match, not anchored: renaming to `## Phase 1: Selection` still matches. Inherited from the declared precedent, so not this PR's to fix.

### Not pursued

- `nofollow:` **Marker-tier split.** `pytest.ini:10` defines `adherence` as *"mechanical gate — grep/AST ratchets and hygiene checks"*, which is what this file is, yet it carries no marker. But roughly half the sibling SKILL.md-text guards are unmarked (`test_hunt_invokes_review_pr.py`, `test_raid_invokes_hunt.py`, the declared model itself) and half mark `adherence`. The PR follows its cited precedent faithfully; the precedent-vs-`pytest.ini` conflict is repo-wide and predates this branch.
- `nofollow:` **Prose-only verification.** Ticket 0390's two Verification bullets are behavioural ("does not spawn an executor", "names it in the run report"); the guards verify only that the prose says so. Inherent to a repo whose code is prose — noted, not chargeable here.
- `nofollow:` **Test density.** 13 phrase assertions for 19 prose lines vs the precedent's ~18 for 46 — denser per line, but consistent in style and function count. The 4-assertion negative guard is not busywork: the ticket log records that an earlier draft (gaze-694) proposed exactly the hand-rolled grep it forbids.
- `nofollow:` **Severity floor.** The 0334/0338 cost evidence is reused from ticket 0378 and applied to a second entry point by architectural inference rather than fresh raid-specific measurement. Sound as an extension of accepted precedent.

---
Panel verdicts: correctness **approve** · consistency **comment** · scope **comment** · red team **request-changes** (finding 1) · doc propagation **comment**.

MinhHaDuong (commented):
## /review-pr round 2 — five-agent panel (correctness · consistency · scope · red team · doc propagation)

**Verdict: request-changes — record corrections only. No logic defect found.**

All four round-1 findings are genuinely resolved in `e5dc0e5`, the guards are a real ratchet (7 of 8 RED on main), and both gates are green. What blocks is that round 2 changed the branch without updating the record: the PR body now asserts an invariant the diff violates and carries a "declined nit" section describing a defect this PR already fixed. The PR body is the merge-commit message, so a false claim in it is introduced-by-this-PR debt, not a nit. Four mechanical text edits clear it.

Panel verdicts: correctness **approve** · consistency **comment** · scope **comment** · red team **request-changes** · doc propagation **request-changes**.

### Round-1 ledger — all resolved

| # | Round-1 finding | Status | Evidence |
|---|---|---|---|
| 1 | No carve-out for an explicitly named ID | RESOLVED | `skills/raid/SKILL.md:96-99`; `test_raid_phase1_carves_out_an_explicit_author_id` RED on main, green on branch |
| 2 | `deferred` skip-listed but still raidable | RESOLVED | Phase 1 now says "any other header listed in `.ergrc`" and names `deferred`; test RED on main |
| 3 | "run report" not raid's vocabulary, no landing spot | RESOLVED, correct step | new sentence sits inside `## Wrap up` step 2; "briefing" is raid's pre-existing term on main (`:284`, `:316`) |
| 4 | Stale forward reference in `skills/hunt/SKILL.md` | RESOLVED | `:44-45` now reads "raid Phase 1 now screens its own targets ([[0390]])"; hunt's own 9 guards still pass |

### Verified positives (recorded so they are not re-litigated)

- **[correctness] The premise is true, by execution.** Throwaway two-ticket store, committed `tickets/erg`: with `[labels]` listing `needs-human`, `erg ready` returned only the unlabelled ticket (`Ready tickets (1): 0002`) while `erg list` showed both. Re-run independently by the synthesizer — same result.
- **[correctness] The gap is real.** `skills/raid/SKILL.md:72` — *"If $ARGUMENTS is 'all open': read open tickets from git-erg `tickets/` or forge"* — no `erg ready`, no `pick-ticket`, on main or branch.
- **[correctness] RED proof reproduces.** Branch: 8 passed. Same file against `origin/main` text: **7 failed, 1 passed.**
- **[correctness] Coverage sweep clean.** `molt`/`scry`/`lair`/`healthcheck`/`gaze` do not select-and-spawn from a raw `tickets/` scan; `nightbeat-supervisor` routes through `beat.py`, which routes through `pick-ticket`. No second unscreened surface.
- **[red team] The force invariant genuinely holds, traced end to end.** `/raid 0123` → Phase 1 carve-out → Phase 5 executor → `Skill(hunt)` → `skills/hunt/SKILL.md:29-30,41-42`: *"an executor already running detached… Each skip is total, the needs-human branch included."* The executor does **not** re-triage. The carve-out is not void.
- **[red team] No later phase re-admits an excluded ticket.** Ralph loop (`:327`), Phase 7 merge, Phase 8, circuit breakers — waves are built only from Phase 1's filtered set.
- **[consistency] Mechanics clean.** `make check-fast` 797 passed / 104 deselected; `make lint` 27 passed; `make check-skills-drift` in sync; `ruff check tests/test_needs_human_screen.py` clean (the 8 repo-wide ruff errors are byte-identical on `origin/main`). Test idiom matches the declared precedent `tests/test_hunt_triage_prestep.py`, down to the shared docstring line "Text-grep hygiene test — fast tier, no marker."
- **[scope] Round-2 additions are legitimate review responses, and both are logged.** The carve-out repairs a violation of the ticket's own Invariants line; the `.ergrc` generalization is a corollary of Phase 1's stated rationale. Ticket log `2026-07-28T14:13Z` records both. Hunt step 2b's *logic* is untouched — only a status pointer changed — so ticket Invariant 1 holds.

---

### Blocking — the record does not match the diff

**B1. `verifiable:` The PR body asserts a false invariant.**
Body, "Invariants held": *"`skills/hunt/SKILL.md` is untouched — 0378's recursion guard stands."* Actual: `git diff origin/main...HEAD -- skills/hunt/SKILL.md` → 2 insertions, 2 deletions, landed in `e5dc0e5` (whose own message says "unstaled hunt SKILL.md's forward reference"). The *ticket's* narrower claim ("Hunt step 2b is unchanged") is still accurate and verified; the PR body's blanket "untouched" is not. Confidence high.

**B2. `verifiable:` The "Known nit, deliberately declined" section describes a defect this PR fixed, quoting text that no longer exists.**
Body: *"`skills/hunt/SKILL.md:46` reads 'screening raid's own target selection is upstream work — see [[0390]]' … Not fixed here."* Current `:45-46`: *"raid Phase 1 now screens its own targets ([[0390]]); neither is a second triage here."* The quoted string is gone. Delete the section or replace with "fixed in `e5dc0e5`". Confidence high.

**B3. `verifiable:` The ticket body still says "run report" twice, contradicting its own log and the shipped prose.**
`tickets/0390-…erg:67` — *"…surfaced to the author in the run report, as a success outcome…"*; `:83` — *"…names it in the run report"*. The log at `2026-07-28T14:13Z` records *"renamed run report -> wrap-up briefing"*, and `skills/raid/SKILL.md:92` ships "wrap-up briefing". Round-1 finding 3 was fixed in the skill and not in the ticket that asked for it. Confidence high.

**B4. `verifiable:` PR body counts are stale.**
Body says "**five guards**"; `grep -c "^def test_"` → **8**. Body reports `make check-fast — 794 passed` / `make lint — 871 deselected`; actual on the branch head: **797 passed** / **874 deselected**. `git rev-list --count HEAD..origin/main` → **4** — the branch has not been rebased since the gate numbers were taken (`rules/git.md` § "Rebase at every gate"). Confidence high.

---

### Non-blocking findings

**F1. `verifiable:` The carve-out's stated rationale is false for the only programmatic caller — and it inverts which path the new screen actually protects.** *Dissent preserved; the author decides.*

`scripts/beat.py:1221` constructs `f"/raid {ticket_id}\n\nRunning on: {hostname}"` — the explicit-ID shape — fully unattended. `skills/raid/SKILL.md:96-99` says of that shape: *"That is the author forcing execution."* It is not. Grep confirms `beat.py:1221` is the **only** programmatic `/raid` call site in the repo, and **no caller anywhere passes `"all open"`**.

Two consequences worth the author's attention:

- The prose's reason is wrong even though today's outcome is right. `ticket_id` comes from `parse_pick()` on `pick-ticket`'s output, which *is* screened by `erg ready`. Nothing in Phase 1 states, relies on, or would survive the loss of that upstream fact — a future requeue/retry path constructing an ID from an unscreened source routes straight through the carve-out.
- Because the discovery path is only ever entered by a human typing `/raid all open`, the net new protection Phase 1's exclusion adds is on the **interactive** invocation, not the autonomous one. The autonomous caller was, and remains, protected by `pick-ticket` — the surface the ticket declared already covered.

**Red team (request-changes, high confidence), verbatim:** *"the fix that answered round 1 reopens the exact autonomous-execution gap ticket 0390 exists to close — currently masked only by an incidental, unenforced upstream fact."* Note also that `9c719e0`'s pre-carve-out text carried the rationale sentence *"a raid runs where the author is not watching"*, which `e5dc0e5` dropped.

**Correctness (approve, high confidence), verbatim:** *"the one programmatic caller of raid always lands on the carve-out (explicit-ID) branch with a pre-screened ID… the discovery-path exclusion this PR adds is exercised only by a human-typed `/raid all open`, exactly the gap the ticket describes."*

Smallest edit that satisfies both readings: scope the carve-out to the invocation's *origin*, not its shape — *"…never a ticket the author named by ID in their own session. An automated caller relaying an ID (e.g. `beat.py`) must have screened it upstream, as `pick-ticket` does."*

**F2. `verifiable:` `.ergrc` has an asymmetric failure mode, and raid's hand-parse specifies neither side.**
Phase 1 says *"Read the label set from `.ergrc` rather than hardcoding it here"* with no statement of what to do when the file is absent or its section empty. Measured on the committed `erg`, same fixture store:

```
[labels] with needs-human, deferred  →  Ready tickets (1): 0002
[labels] present but EMPTY           →  Ready tickets (2): 0001 (labels: needs-human), 0002
.ergrc ABSENT                        →  Ready tickets (1): 0002
```

An **absent** `.ergrc` is safe — `erg` falls back to its defaults. An **empty** `[labels]` section fails open. Raid's hand-parse inherits neither behaviour: an agent told to read the file and finding it missing parses nothing and screens nothing. One clause fixes it: *"if `.ergrc` is absent or its `[labels]` section empty, fall back to `needs-human` and `deferred`."* (This corrects doc propagation's finding 7, which assumed the absent-file case was the dangerous one — measured, it is the safe one.) Confidence high.

**F3. `verifiable:` `test_raid_phase1_carves_out_an_explicit_author_id` checks word presence, not polarity.**
Replaced the carve-out paragraph with its logical negation ("**No carve-out.** …naming a ticket by ID does not bypass the skip-label screen; both paths are excluded identically") in a scratch copy:

```
$ pytest tests/test_needs_human_screen.py -k carves_out -q
1 passed, 7 deselected
```

The exact revocation of ticket 0390's Invariant passes the guard written to protect it. Reproduced independently by the synthesizer. This also means F1's remediation, once written, will not be mechanically enforced. Confidence high.

**F4. `verifiable:` `test_pick_ticket_does_not_hand_roll_the_screen` is defeatable by the case it forbids.**
Appended to pick-ticket step 1: *"As a belt-and-suspenders check, also inspect each candidate ticket's needs-human marker yourself and drop it, independent of erg ready."* — a second hand-rolled screen, precisely Action 1's forbidden case.

```
$ pytest tests/test_needs_human_screen.py -k pick_ticket -q
2 passed, 6 deselected
```

Reproduced independently. The round-1 narrowing to `\b(grep|rg|awk|sed|jq)\b(?=\s*[-"'`])` fixed the noun false-positives but opened this: `jq .labels` and `grep needs-human` also no longer match. Confidence high.

**F5. `verifiable:` `test_raid_phase1_excludes_every_ergrc_label`'s `"deferred"` assertion can pass vacuously.**
Red team reverted the exclusion to `needs-human`-only while keeping the word `deferred` in unrelated prose ("a ticket may be deferred for unrelated scheduling reasons, which is a separate concern from this screen") — test green. The converse also holds: a correct paraphrase avoiding the literal token ("needs-human and its postponed sibling") turns it red. Confidence high.

**F6. `verifiable:` One of the eight guards is green on `origin/main`.**
`test_pick_ticket_does_not_hand_roll_the_screen` — all four of its sub-assertions pass on main, because main's step-1 text never discusses labels at all. It is a forward ratchet against a future regression, not a proof of this change. Legitimate, but it should not be counted among the RED-proof guards. Confidence high.

**F7. `consider:` Action 3 shipped broader than the author approved.**
Ticket log `:11`: *"a raid whose queue holds only **needs-human** tickets returns the batched decision lists"*. Shipped prose `skills/raid/SKILL.md:101`: *"If EVERY discovered ticket is **skip-labelled**…"* — i.e. `needs-human` ∪ `deferred`. A `deferred` ticket is postponed, not awaiting a decision, yet Phase 1 instructs naming it *"with the decisions it is waiting on"*. An all-`deferred` queue therefore manufactures a decision round for tickets awaiting none. Either the approved decision was meant to cover both (say so in the log) or the prose should scope the rule to `needs-human`. Confidence medium — the discrepancy is quoted and certain; whether it matters is the author's call.

**F8. `consider:` No test locks the hunt reword.**
`grep "0390\|screens its own" tests/` → no match outside the ticket. The 236-line suite scopes itself in its docstring to "two selection surfaces" and omits hunt as a third, so the wording round 1 asked for can silently regress. Confidence high on the gap, medium on whether a ninth guard is worth it.

**F9. `consider:` Four phrasings of one mechanism, plus a spelling split.**
`hunt:44` "skip-lists the label out" · `pick-ticket:23-27` "label screen … suppresses" · `raid:83` "skip-labelled … hold a ticket back" · `nightbeat-supervisor:42` "tickets labeled needs-human or deferred". Also `labelled` (×3, new here) beside `labeled` (pre-existing) describing the identical set. No file violates `lang/en.md` on its own; the corpus now reads as four rules. A glossary pass, not this PR's job. Confidence medium.

**F10. `consider:` Rebase before merging.** Branch is 4 commits behind `origin/main`; `rules/git.md` wants currency at each gate, and the gate numbers in B4 were taken against the stale base. Confidence high.

### Not pursued

- `nofollow:` **`[[0390]]` wikilink is unique in `skills/`** (`grep -rn '\[\[0[0-9]\{3\}\]\]' skills/` → 1 hit). Every other skill uses plain "ticket 0390". Inherited from ticket 0378's text, not introduced here; nothing resolves wikilinks mechanically, so nothing breaks when 0390 moves to `tickets/closed/`.
- `nofollow:` **Out-of-order ticket-log timestamps** (14:40Z before 13:34Z). `git log -p --follow` places both on `origin/main` before this branch existed. This PR's three added lines are correctly ordered.
- `nofollow:` **Behavioural-vs-textual verification.** The ticket's two Verification bullets are behavioural ("does not spawn an executor"); the guards assert only that the prose says so. Inherent to a repo whose code is prose, and identical in the declared precedent.
- `nofollow:` **`tickets/AGENTS.md` and `rules/workflow.md` are silent on the skip-label screen as a cross-skill rule.** Real gap — `AGENTS.md:36` still frames `Label:` as `erg ready` filtering only, when it now also gates raid's direct read — but outside this ticket's declared scope. Worth a follow-up.
- `nofollow:` **Severity-floor evidence borrowed from 0378.** The 0334/0338 citation is verbatim from ticket 0378 and establishes "needs-human executed autonomously is costly", not "raid Phase 1 specifically let one through". The gap is independently verifiable by reading Phase 1, so the thin citation does not weaken the case.
- `nofollow:` **Test density.** 236 lines / 8 tests / 19 asserts for ~33 prose lines, vs the precedent's 191 / 9 / 18 for ~55. Denser, but three regions each need their own extractor plus positive and negative guards. No two tests assert the same claim; no pre-existing guard elsewhere in `tests/` covers these two files.
- `nofollow:` **`README.md` catalog and `STATE.md`.** `make check-skills-drift` in sync; `grep "0390\|0378" STATE.md` → no match. Frontmatter `description:` fields correctly omit the screen, matching how `hunt`'s description omits 0378's triage.

---

**To clear the verdict:** B1-B4 (edit the PR body, edit two ticket-body lines, rebase and re-take the gate numbers). F1 and F2 are one clause each in `skills/raid/SKILL.md` and are the two findings most worth folding in while the file is open; F3-F6 are test-quality debt that would justify a follow-up ticket rather than another round here.

MinhHaDuong:
## `/review-pr` — five-agent panel

Panel: Correctness, Consistency, Scope, Red team, Doc propagation (all `sonnet`, decorrelated from the coder tier). Panel verdicts: **approve ×2** (Scope, Doc propagation), **comment ×2** (Correctness, Consistency), **request-changes ×1** (Red team). Dissent preserved below — the split is real, and it is about the tests, not the prose.

The behaviour change is sound. Three agents independently verified that the raid Phase 1 exclusion, the explicit-ID carve-out, and the `pick-ticket` restraint each describe the harness accurately, that `beat.py:1176→1221` is the only programmatic `/raid <id>` caller as the prose claims, and that ticket 0378's Verification bullet 3 is untouched and still load-bearing. What follows is what the panel found on top of that.

---

## Blocker — the new guards do not hold the invariant they were written for

**Red team `request-changes`; reproduced independently by the synthesizing agent before posting.**

`tests/test_needs_human_screen.py`'s 8 assertions are presence checks over prose. Every one of them passes on a mutation that states the **logical negation** of the invariant it guards. This is the same defect review round 2 says it fixed, and it is not fixed. Three mutations, applied to scratch copies under `/tmp`, each run against the unmodified test file:

**1. Carve-out inverted — `test_raid_phase1_carves_out_an_explicit_author_id` stays green.** Round 2's fix was to capture and pin the verb (`m.group(1) == "runs"`). The verb sits in a bolded lead sentence; the negation goes in the next one:

```
**Carve-out: an explicit `/raid <id>` runs.** Correction: on reflection it does
not. The exclusion governs target *discovery* (the "all open" path) AND a ticket
named by ID equally; a skip-labelled ticket named explicitly is dropped like any
other.
```
→ `8 passed in 0.09s`. The regex `Carve-out: an explicit \`/raid <id>\` (\w+)` inspects exactly one word. A same-paragraph reversal is invisible to it.

**2. A genuine second hand-rolled screen — `test_pick_ticket_does_not_hand_roll_the_screen` stays green.** Round 2's fix was to require the delegation be stated positively *before* the absence checks. Keeping the required phrases and appending the antipattern satisfies both halves:

```
One screen, applied by the tool that owns the vocabulary (ticket 0390). As a
belt-and-braces measure, also walk each candidate entry yourself and silently
drop any whose ticket file carries the needs-human string, in case a stale
binary on this host forgot to honour the config.
```
→ `8 passed in 0.08s`. This is precisely what Action 1 forbids — a duplicate definition to keep in sync with `.ergrc` — and the guard written to forbid it does not see it.

**3. The core exclusion inverted — `test_raid_phase1_excludes_needs_human` stays green.**

```
**It is a common misreading that a skip-labelled ticket is never a raid target.**
In fact Phase 1 picks it like any other: ...
```
→ `8 passed`. Both required substrings (`Label: needs-human`, `never a raid target`) survive a sentence that says the opposite.

Red team reports the same result for the remaining five assertions, with mutant text attached for each.

**What is blocking here is the claim, not the guard strength.** Regex over natural-language prose cannot be made polarity-proof in general, and this PR should not spend a fourth round chasing it — round 2 measurably raised the cost of a mutation from "paraphrase" to "one appended sentence", and that is about the ceiling for the technique. What must not ship is the assertion that the problem is solved:

- PR body, *Review rounds folded in* → Round 2, bullet 3: *"Fixed by pinning meaning rather than vocabulary"* — false, demonstrated above.
- Ticket log `2026-07-28T14:30Z`: *"gave both polarity-blind guards real polarity"* — false. The log is append-only, so this needs a correcting `note` entry rather than an edit.

`/verify-gate` and every later reader key on those sentences. **Bounded ask: retract both claims and restate what the guards actually do** — they catch deletion and drift of the rule's vocabulary, not its inversion. That is a real if modest ratchet, honestly described. If the honest description makes a guard look like it is not earning its place, that is useful information too.

---

## Minor findings

**`verifiable:` — the round-2 RED-proof number does not reproduce.** *(Correctness, high confidence)* The PR body claims *"Round-2 guards: 5 of 6 assertions fail against the round-1 commit."* Evaluating each of the 6 round-2-touched assertions against `e5dc0e5`'s own skill files, only **1 of 6** (`"beat.py" in region`) flips. The other 5 were already true at round 1 — `skills/pick-ticket/SKILL.md` was written once in `ab96f97` and never edited again in this PR. Running the full module against round-1 files gives 7 passed / 1 failed, on the `beat.py` line. Round 1's claim ("4 of 5 fail against `origin/main`") *does* reproduce exactly, as does the 8/8-RED-against-`origin/main` result for the final suite.

**`verifiable:` — an empty `.ergrc [labels]` section defeats the screen on the one path the carve-out's safety argument rests on.** *(Correctness, Red team, and Doc propagation — reproduced independently three times against the committed `tickets/erg` binary.)* With `.ergrc` **absent**, `erg ready` correctly falls back to defaults and suppresses a `Label: needs-human` ticket. With `[labels]` **present but empty or comments-only**, the same ticket is returned as ready. raid Phase 1 names this asymmetry and tells a hand-parse to fall back — good. `pick-ticket` does not mention it at all, and `pick-ticket` is exactly where round 2's rationale locates the autonomous path's safety (*"beat picks through `pick-ticket`, whose candidates came from `erg ready`"*). That sentence holds only while `[labels]` is non-empty or the file is absent. Not present in the committed default state, so not a live hole — but the carve-out leans on a guarantee weaker than stated. One sentence in `pick-ticket`, or a follow-up ticket, closes the gap between claim and mechanism.

**`verifiable:` — the label match is case-sensitive on both header key and value.** *(Red team, high confidence; pre-existing `erg` behaviour, not introduced here)* `Label: Needs-Human` is rejected by `erg validate` but still appears in `erg ready --json` output. `label: needs-human` is not recognized as a `Label:` header at all and yields `"labels": []`. Neither `erg ready` nor raid Phase 1's literal-string grep folds case, and `pick-ticket` runs no `erg check`/`validate` gate before `erg ready`. A single wrong-case letter routes a needs-human ticket into autonomous execution. Ruled out as bypasses (checked, handled correctly): trailing whitespace on an `.ergrc` label line, CRLF line endings, and repeated `Label:` headers on one ticket.

**`verifiable:` — no re-screen between Phase 1 and Phase 5.** *(Red team, medium confidence)* Phases 2–5 of `skills/raid/SKILL.md` contain no mention of `Label`/`needs-human` (grep confirmed), and `skills/hunt/SKILL.md` step 2b deliberately disables itself for a detached Phase 5 executor, by design per 0378 bullet 3. A ticket the author hand-labels mid-raid, after Phase 1's one-time read, is caught nowhere. Raid waves run tens of minutes; the window is not negligible for the exact failure mode this ticket exists to close.

**`verifiable:` — em-dash chain.** *(Consistency, medium confidence)* `skills/raid/SKILL.md:108-109` carries two em-dashes in one sentence, against `rules/prose/_all.md` ("at most one em-dash per paragraph"). The only prose-rule hit in the new text — no antithesis-reversal, transition drumbeat, weak copula, importance inflation, or hedging stack found.

**`verifiable:` — the ticket's *Relevant files* section still calls `skills/hunt/SKILL.md` unchanged.** *(Scope, medium confidence)* The round-1 log entry records the deviation; the body prose it contradicts was not updated alongside it.

**`consider:` — dissent on "empty-run report" at `skills/raid/SKILL.md:114`.** Consistency reads it as a sibling the run-report→briefing rename missed, newly coined by this PR and then locked in permanently by `tests/test_needs_human_screen.py:262`. The synthesizing agent reads it as deliberate: Phase 1 must name what Action 3 rules *out*, and the guard's own message says so. Both readings are on the page; the author's call. If deliberate, a half-line in the test docstring distinguishing "the artifact" from "the outcome being rejected" would stop a future reader re-litigating it.

**`consider:` — prose register heavier than the surrounding corpus.** *(Consistency, medium confidence)* The new Phase 1 block runs 347 words for one screening rule; `pick-ticket`'s equivalent-scope step covers screening in ~84, and raid's own densest pre-existing subsection covers a four-step recovery in 239. Sentences like *"a hand-parse that inherits that asymmetry screens nothing while looking correct"* read as defensive glosses against a misreading rather than instructions to an executing agent. Precedent cuts the other way, though: hunt's own 0378-era step 2b established this register in the same skill family.

**`consider:` — child tickets from the monster-ticket decomposition are not explicitly re-screened.** *(Red team, low confidence)* The decomposition instruction sits textually right after the exclusion in Phase 1; nothing demands the resulting children pass back through it. Documentation gap, not a demonstrated bypass.

**`consider:` — commit `9c719e0` bundles two independent fixes** (widening a negative guard's regex; restoring a dropped log actor field). Both are explained in the message. *(Scope, low confidence)*

**`nofollow:` — `beat.py` hardcoded by name** in the carve-out prose. `rules/workflow.md` asks for capability names, but this is the established corpus idiom (`pick-ticket`, `molt`, `hunt`, `nightbeat-supervisor` all do the same), and the factual claim checks out against `scripts/beat.py:1221`. Not a regression this PR introduces.

**`nofollow:` — "PR" rather than "merge request"** throughout `skills/raid/SKILL.md`. Whole-file pre-existing drift; the new text adds none.

**`nofollow:` — `tickets/spec-erg-v1.md:67,87` still says `Tag`/`[tags]`** where the binary, `AGENTS.md`, and `.ergrc` all say `Label`/`[labels]`. Pre-existing, untouched, out of scope.

---

## Checked clean

- `tickets/erg check tickets/` — **PASS**, 359 tickets. No `Status:` header, log verbs `created`/`note` only, ticket correctly **not** pre-closed. The one out-of-order log timestamp predates this PR on `origin/main`; the four appended lines are correctly ordered.
- **Cross-PR ticket-ID collision for 0390** — enumerated open PRs (#698, #678) and queried each with `gh pr view <n> --json files` (not the `gh pr list --json files` trap). No collision.
- `make check-skills-drift` — **OK: Skills catalog is in sync.** No frontmatter `description:` changed in the three edited skills.
- **Cross-references** — `[[0390]]` resolves to one site (`skills/hunt/SKILL.md:46`) and is now accurate. Ticket 0378 is closed (`tickets/closed/0378-…`, autoclosed PR #694); its Verification bullet 3 is intact and still says what the carve-out relies on. `skills/nightbeat-supervisor/SKILL.md:42` already states the exclusion at its own layer — no contradiction. `skills/release/SKILL.md:27` scans all open tickets for `needs-human` deliberately, bypassing suppression because it needs to *see* the suppressed ones — different mechanism, no conflict.
- **"#696 merged as `475333f` without touching those lines"** — confirmed, but only via the merge's net effect (`git diff e920b7a 475333f -- skills/hunt/SKILL.md`); naive branch-tip diffing misleads here because #696's branch predates 0378's merge.
- **The `skills/hunt/SKILL.md` deviation stays wording-only** — the full-file diff against `origin/main` is exactly the two quoted lines; step 2b's logic, the recursion guard, and bullet-3 scoping are untouched, and `tests/test_hunt_triage_prestep.py` is 9/9 green.
- **Per-hunk scope audit** — every hunk maps to a stated ticket Action, Exit criterion, or the log/body wording sync. No orphan hunks.

## Test run

| Gate | Result |
|---|---|
| `make check-fast` | **797 passed, 104 deselected** — reproduces the PR body exactly (3 agents independently) |
| `make lint` | **27 passed, 874 deselected** — reproduces exactly |
| `tests/test_needs_human_screen.py` | 8 passed |
| `tests/test_hunt_triage_prestep.py` | 9 passed |
| `make check` (full) | **899 passed, 1 skipped, 1 failed** — see below |

The single `make check` failure is **environmental, not a PR defect**: `tests/test_bash_suites.py::test_shell_suite[test_guard_cd_primary_repo.sh]` fails two allow-cases whose assertion requires empty stderr, and stderr carries `bash-env: KEYS provider not found: openrouter` from a missing local provider file. `git diff origin/main...HEAD -- tests/test_bash_suites.py tests/shell scripts` is empty — this PR touches none of that surface. Worth noting that the ticket's `make check` exit criterion was satisfied by the author-directed fast gate (logged 2026-07-28T13:51Z), so this failure had not been surfaced before now; it is a standing local-environment condition that will bite any full-gate run on this machine.

---

## Recommendation

The prose change is correct, well-evidenced, and in scope — I would merge it. The one thing that should not land as written is the PR body's and ticket log's claim that the polarity-blindness is fixed: three mutations reproduce it in under a minute each, and a later gate will trust that sentence. Correct the two claims (plus the 5-of-6 RED number), and everything else above is author's-discretion follow-up.

MinhHaDuong:
/verify-gate round=1 verdict=REROLL
Exit criteria: 3/3  Review: 4 unresolved (all verifiable-tier minors; blocker resolved)  Simplify: 0
Minors: verifiable:6 (2 resolved, 4 unresolved) consider:4 nofollow:3 malformed:0

Top reasons:
1. `verifiable:` em-dash chain still present — `skills/raid/SKILL.md:108` ("stays screened upstream — beat picks through `pick-ticket`, whose candidates came from `erg ready` — which is why…"), two em-dashes in one sentence against `rules/prose/_all.md`. No commit after the 14:48:59Z review touches it.
2. `verifiable:` ticket body Relevant-files still says `skills/hunt/SKILL.md` is "unchanged by this ticket" while the diff changes it (wording-only, but the body was never synced with the round-1 log entry recording the deviation).
3. `verifiable:` case-sensitive label match — recorded only as a log note (14:50Z) in the very ticket this PR closes; once merged, the finding archives with it. Needs a durable disposition: a ticket at the destination repo (git-erg — this is erg-binary behaviour), or an explicit author waiver on the PR.
4. `verifiable:` no re-screen between raid Phase 1 and Phase 5 — same disposition gap: log note only, no follow-up ticket and no explicit waiver.

Rationale: The round-3 blocker is genuinely resolved — cfd0b12 (14:50:59Z, after the comment) retracts the inversion-proof claim in the PR body, appends the correcting log note (14:49Z, append-only discipline respected), and adds the pick-ticket fail-open paragraph; the RED-proof number correction is in the body. All three exit criteria carry concrete evidence: raid Phase 1 exclusion at skills/raid/SKILL.md:83-92 with wrap-up landing at :350 and the Action-3 decision at :113 (ab96f97..cfd0b12); pick-ticket names the erg-ready skip-label screen without hand-rolling a second one; the 8-guard suite plus the 9 hunt-triage guards run 17/17 green in the gate's isolated worktree at cfd0b12, with the 8/8-RED-against-origin/main proof independently reproduced by review round 3. make lint 27 passed; the single full-gate failure (test_guard_cd_primary_repo.sh) is environmental and on a surface this diff provably does not touch (author gate directive logged 13:51Z). What bounces the PR is mechanical, not substantive: the gate does not triage verifiable-tagged minors by the reviewer's closing recommendation — unresolved verifiable minors are blocker-adjacent by rule. Items 1-2 are one-line fixes; items 3-4 need only an explicit disposition the gate can see. Root cause: Agent Error (in-scope prose-rule violation and body/log desync left standing after three review rounds).

MinhHaDuong:
Exit criteria: 3/3  Review: 0 unresolved  Simplify: 0
Minors: verifiable:6 (6 resolved) consider:4 nofollow:3 malformed:0

All four round-1 unresolved items carry post-verdict evidence in `fc8133b` (14:59:34Z, after the 14:57:04Z round-1 comment):
1. Em-dash chain — `skills/raid/SKILL.md:108-109` rewritten to colon + comma; one em-dash remains in the paragraph, within rule.
2. Stale Relevant-files — ticket body now reads "gets a two-line wording fix", synced with the round-1 deviation log entry.
3. Case-sensitive label match — durable disposition verified: git-erg#325 exists, OPEN, at the destination repo per the cross-repo convention, and names the reliance created by this PR.
4. Phase 1→5 re-screen window — promoted to `tickets/0391-re-screen-skip-labels-at-raid-phase-5-la.erg`, bundled on this branch as a complete handoff document; referenced in the PR body without a close claim, so it survives the merge. Cross-PR collision scan for 0390/0391 via per-PR `gh pr view --json files` (positive control: PR #678's file list returned content): no collision. `erg check` PASS, 360 tickets.

Re-verified at head `fc8133b` in the gate's isolated worktree: 8-guard screen suite + 9 hunt-triage guards 17/17 green (the em-dash fix edited guarded prose, so this was not optional); `make lint` 27 passed; `make check-fast` 797 passed, 104 deselected. Scope vs merge-base `475333f` is exactly 6 files, each traceable to a ticket Action, the Test criterion, or a gate-mandated disposition. mergeStateStatus CLEAN.

Rationale: The strongest remaining attack is the one the PR itself now states honestly — the prose guards catch deletion and vocabulary drift, not semantic inversion — and after the cfd0b12 retraction no artifact overclaims otherwise, which is what round 1's blocker actually demanded. The two erg-behaviour gaps (case folding, mid-raid labelling) are real but pre-existing, and both now have trackable homes outside a ticket that closes on merge. Nit, non-blocking: the 0391 reference uses "Related follow-up:" rather than the `Ticket-ref:` line convention — functionally equivalent (no close claim), but the machine-parsed form is the documented one.

